PR #6, opened 2019-03-04
JWT based authentication support

t-chaik:
This series of patch introduce JWT based authentication support for the min gRPC endpoint:
- [x]  Server-side TLS support for the main gRPC endpoint:
    * [x] CLI configuration interface.
- [x] Client-side TLS support for the `"grpc"` `blobstore` backend:
- [x] JWT authentication for the main gRPC endpoint:
    * [x] JWT token extraction and validation.
    * [x] JWT token validation caching.
    * [x] CLI configuration interface.

Suggestion on better naming for API, CLI and configuration are very welcome.


pkg/auth/jwt.go
```diff
@@ -0,0 +1,47 @@
...
+	"io/ioutil"
+
+	"github.com/dgrijalva/jwt-go"
+	"github.com/grpc-ecosystem/go-grpc-middleware/auth"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+var (
```
EdSchouten: For stuff like bb-storage I try to structure everything as a bag of classes as much as possible. Would there be a way for us to eliminate these global variables?

t-chaik:
> Don't be discouraged by any of my remarks. Those changes look pretty neat already. Thanks for working on this! \o/

No problem, I submitted that WIP precisely to get these remarks! Has you may know, I'm very new to Go so any comments are more than welcome.

pkg/proto/blobstore/blobstore.proto
```diff
@@ -73,8 +73,13 @@ message CircularBlobAccessConfiguration {
 }
 
 message GRPCBlobAccessConfiguration {
-    // Endpoint address of the GRPC server (e.g., "localhost:8982").
+    // Endpoint address of the GRPC server (e.g., "http://localhost:8982").
```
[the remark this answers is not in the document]
t-chaik: Let's move TLS client details to a specific message then (holding `server_ca_file`, `client_cert_file` and `client_key_file`). We'll need an explicit boolean somewhere to activate TLS if we are not deducing it from the URL in order to support client-unauthenticated TLS encrypted connection to trusted server though.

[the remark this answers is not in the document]
EdSchouten: > We'll need an explicit boolean somewhere to activate TLS if we are not deducing it from the URL in order to support client-unauthenticated TLS encrypted connection to trusted server though.

In that case we could use the empty message for that, right? Protobuf makes a difference between an empty message and one that is absent. For Golang, you see that nested messages use pointers. When no nested message is provided, the pointer is nil. From a configuration file point of view, it will look like this:

```
// No TLS.
grpc {
  endpoint: "localhost:8982"
}

// TLS with default options.
grpc {
  endpoint: "localhost:8982"
  tls {}
}

// TLS with custom CA, etc.
grpc {
  endpoint: "localhost:8982"
  tls {
    server_ca: "..."
    ...
  }
}
```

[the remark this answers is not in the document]
t-chaik: Perfect!

cmd/bb_storage/main.go
```diff
@@ -85,10 +94,32 @@ func main() {
 	})
 
 	// RPC server.
-	s := grpc.NewServer(
-		grpc.StreamInterceptor(grpc_prometheus.StreamServerInterceptor),
-		grpc.UnaryInterceptor(grpc_prometheus.UnaryServerInterceptor),
-	)
+	var opts []grpc.ServerOption
+	if *tlsEncryption {
```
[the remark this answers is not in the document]
t-chaik: `flag` seems to enforce a default value and `string` cannot be `nil` as it seems, to defaulting to empty and string...

pkg/auth/jwt.go
t-chaik: Should be better now: I've moved this to a `struct` implementing an interface that will also serve as token validation cache (WIP).

cmd/bb_storage/main.go
```diff
@@ -85,10 +93,35 @@ func main() {
 	})
 
 	// RPC server.
-	s := grpc.NewServer(
-		grpc.StreamInterceptor(grpc_prometheus.StreamServerInterceptor),
-		grpc.UnaryInterceptor(grpc_prometheus.UnaryServerInterceptor),
-	)
+	var options []grpc.ServerOption
```
EdSchouten: Is there any chance we can place this chunk of code under `pkg/`? That way we can also use it in the other components (e.g., bb-scheduler).

cmd/bb_storage/main.go
```diff
@@ -85,10 +93,35 @@ func main() {
...
+	if len(*tlsCertFile) > 0 && len(*tlsKeyFile) > 0 {
```
EdSchouten: Maybe this is just personal preference, but I think it often makes more sense to use `*tlsCertFile != ""` for string comparison in Go. It's not only shorter, it also makes it more explicit this is about strings; not lists of objects.

cmd/bb_storage/main.go
```diff
@@ -10,23 +10,31 @@ import (
...
 	"google.golang.org/grpc/status"
 )
 
 func main() {
 	var (
 		blobstoreConfig  = flag.String("blobstore-config", "/config/blobstore.conf", "Configuration for blob storage")
 		webListenAddress = flag.String("web.listen-address", ":80", "Port on which to expose metrics")
+		tlsCertFile      = flag.String("tls-cert-file", "", "Certificate for TLS server authentication")
+		tlsKeyFile       = flag.String("tls-key-file", "", "Private key for TLS server authentication")
+		authMethod       = flag.String("auth-method", "none", "Authentication method")
```
EdSchouten: Mention which values are permitted here?

pkg/auth/jwt_auth_cache.go
```diff
@@ -0,0 +1,78 @@
...
+	return &jwtAuthCache{
+		headerScheme: "bearer",
+		jwtSecret:    bytes.Trim(data, "\n"),
+		tokenCache:   make(map[string]time.Time),
+		cacheOrder:   make([]string, 0),
+		cacheSize:    100,
+	}, nil
+}
+
+func BuildJWTAuthFunction(authCache AuthCache) func(ctx context.Context) (context.Context, error) {
```
EdSchouten: I think we could do without this helper function. The only difference between `authCache.ValidateCredentials` and the type of the function returned by this function is that this one returns an additional context.

What if `ValidateCredentials()` had the same prototype? In that case you could use `authCache.ValidateCredentials` directly, instead of `BuildJWTAuthFunction(authCache)`. This would also allow you to eliminate the `AuthCache` interface.

pkg/auth/jwt_auth_cache.go
```diff
@@ -0,0 +1,78 @@
...
+	data, err := ioutil.ReadFile(secretFile)
+	if err != nil {
+		return nil, err
+	}
+	return &jwtAuthCache{
+		headerScheme: "bearer",
+		jwtSecret:    bytes.Trim(data, "\n"),
+		tokenCache:   make(map[string]time.Time),
+		cacheOrder:   make([]string, 0),
+		cacheSize:    100,
```
EdSchouten: Would we want to expose this as a parameter of `NewJWTAuthCache()`?

pkg/auth/jwt_auth_cache.go
```diff
@@ -0,0 +1,78 @@
...
+	headerScheme string
+	jwtSecret    []byte
+	tokenCache   map[string]time.Time
+	cacheOrder   []string
+	cacheSize    int
+}
+
+// NewJwtAuthCache creates an AuthCache that implements bearer style
+// JWT based authorization validation.
+func NewJwtAuthCache(secretFile string) (AuthCache, error) {
```
EdSchouten: `NewJWTAuthCache`, per Go style guide.

pkg/auth/jwt_auth_cache.go
```diff
@@ -0,0 +1,78 @@
...
+	data, err := ioutil.ReadFile(secretFile)
```
EdSchouten: It may be best to let this function take a `[]byte` as an argument with the secret directly, as that would make unit testing significantly easier.

pkg/auth/jwt_auth_cache.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/grpc-ecosystem/go-grpc-middleware/auth"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type jwtAuthCache struct {
+	headerScheme string
+	jwtSecret    []byte
+	tokenCache   map[string]time.Time
```
EdSchouten: There is no guarantee that this cache is used synchronously, right? I guess you'll need to place a mutex around `tokenCache` and `cacheOrder` accesses.

pkg/blobstore/configuration/create_blob_access.go
```diff
@@ -56,6 +59,36 @@ func CreateBlobAccessObjectsFromConfig(configurationFile string) (blobstore.Blob
 	return contentAddressableStorage, actionCache, nil
 }
 
+func newClientTLSFromFiles(serverCaFile string, clientCertFile string, clientKeyFile string) (grpc_credentials.TransportCredentials, error) {
```
EdSchouten: Probably more elegant to pass in a pointer to the proto message here. That way the switch below doesn't need to be refactored in case more TLS options are added.

pkg/blobstore/grpc_metadata_forwarding_blob_access.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+type grpcMetadataForwardingBlobAccess struct {
+	remoteBlobAccess BlobAccess
+}
+
+// NewGrpcMetadataForwardingBlobAccess creates a BlobAccess that converts
+// incoming client request metadata to outcoming metadata before delegating
+// the actual request handling to an underlying GRPC BlobAccess.
+func NewGrpcMetadataForwardingBlobAccess(remoteBlobAccess BlobAccess) BlobAccess {
```
EdSchouten: `NewGRPCMetadataForwardingBlobAccess`

pkg/proto/blobstore/blobstore.proto
```diff
@@ -42,6 +42,10 @@ message BlobAccessConfiguration {
         // Fan out requests across multiple storage backends to spread
         // out load.
         ShardingBlobAccessConfiguration sharding = 9;
+
+        // Delegate requests to an underlying backend but forward
+        // incoming client metadata (including credentials).
+        GRPCMetadataForwardingBlobAccessConfiguration grpc_metadata_forwarding = 10;
```
EdSchouten: Do we expect any configuration options to appear for this adapter? If not, we could consider not having a `GRPCMetadataForwardingBlobAccessConfiguration` at all. Instead, we could embed a `BlobAccessConfiguration`. This has the advantage of preventing the need for an unnecessary `remote {}` block in the configuration.

pkg/auth/jwt_auth_cache.go
t-chaik: Yes, that was one of my concern and given that I don't know much about the Go threading model, I wasn't too sure about what the best approach would be. So let's protect the cache with a mutex then.

cmd/bb_storage/main.go
t-chaik: In some kind of `CreateServerOptions(...)` `func`? Sure, where do you think that should live? I've run out of ideas for naming...

EdSchouten: Exactly. If we don't feel like creating a separate package for that, I'm fine with throwing that in pkg/util.

t-chaik:
@EdSchouten: I've added basic unit tests for the JWT auth. validator. How does that all look to you now?

pkg/blobstore/configuration/create_blob_access.go
```diff
@@ -56,6 +59,36 @@ func CreateBlobAccessObjectsFromConfig(configurationFile string) (blobstore.Blob
 	return contentAddressableStorage, actionCache, nil
 }
 
+func newClientTLSFromFiles(config *pb.ClientTLSConfiguration) (grpc_credentials.TransportCredentials, error) {
+	var certificate tls.Certificate
```
EdSchouten: I think this looks a bit suspicious. In case no client certificate/key is provided, we end up creating a slice with a default-initialized `tls.Certificate`. What if we were to restructure this to:

```go
var certificates []tls.Certificates
if config.ClientCertFile != "" || config.ClientKeyFile != "" {
    certificate, err := tls.LoadX509KeyPair(....)
    if err != nil {
        return nil, err
    }
    certificates = append(certificates, certificate)
}
```

Added advantage of that approach is that you can then remove the `var err error` and `var pemCerts []byte` later on.

pkg/auth/jwt_auth_cache_test.go
```diff
@@ -0,0 +1,65 @@
...
+	exp := time.Date(2970, 1, 1, 1, 1, 0, 0, time.UTC)
+	secret := []byte("the-256-bit-secret")
+	ctx := generateContext(secret, exp)
+
+	authCache := auth.NewJWTAuthCache(secret, 1)
+	_, err := authCache.ValidateCredentials(ctx)
+	require.NoError(t, err)
+}
+
+func TestJWTAuthCacheInvalidCredentials(t *testing.T) {
```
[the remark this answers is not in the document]
t-chaik: Well, first thing do to should probably to submit that path upstream and see how the maintainer welcomes it. If if doesn't get accepted and considering that once cached we check for expiry ourself, we could fallback to a solution where we simply wait for expiry. Not as nice as time shift simulation but simple enough to work.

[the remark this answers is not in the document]
t-chaik: I've added a `TODO` for now and opened a side issue for this: https://github.com/buildbarn/bb-storage/issues/7.

mickael-carl:
Hi there! Mind rebasing your PR off master? Then we can merge this!

prestonvanloon:
Any update on this? Authentication seems like a hard requirement for many deployments of buildbarn.

EdSchouten:
@t-chaik is unfortunately no longer contributing to this project, which is why this PR never made any further progress. Fortunately, some progress has been made on this in the meantime. gRPC servers can nowadays be configured to use an authenticator:

https://github.com/buildbarn/bb-storage/blob/7e30b994ba912b94203545104137364884b1ae9c/pkg/proto/configuration/grpc/grpc.proto#L29-L65
https://github.com/buildbarn/bb-storage/blob/7e30b994ba912b94203545104137364884b1ae9c/pkg/grpc/authenticator.go#L15-L54

There are a couple of basic ones: always allow, always deny, TLS client certificate validation and an `any()` function that you can use to chain multiple authenticators. All that now needs to be done is that someone steps up and adds one that validates JWTs, or uses OAuth2, etc..

It now makes sense to work on this, because Bazel finally gained support for properly passing in credentials: https://github.com/bazelbuild/bazel/pull/10015 https://github.com/bazelbuild/bazel/pull/10634

prestonvanloon:
Awesome! Just two days ago :)

My project just got setup with RBE and are looking to use buildbarn for our CI pipeline. I'd love to be able to expose a public endpoint for my team to make authenticated build requests. How can I help? Does bb-storage need support for jwt/oauth2/etc or is it already good to go with the latest bazel and some flag configuration?

EdSchouten:
> Awesome! Just two days ago :)
> 
> My project just got setup with RBE and are looking to use buildbarn for our CI pipeline. I'd love to be able to expose a public endpoint for my team to make authenticated build requests. How can I help? Does bb-storage need support for jwt/oauth2/etc or is it already good to go with the latest bazel and some flag configuration?

You should be able to get it all working by doing this:
1. Ensure you're using Bazel master.
2. Call Bazel with `--remote_executor=grpcs://hostname:port --remote_header=authorization="Bearer some-oauth2-access-key-or-jwt-goes-here"
3. You will need to set up bb-storage, but you'll have to patch it up to add a new implementation of Authenticator that validates the token. You'll be able to extract the token from within the Authenticator by calling [metadata.FromIncomingContext](https://godoc.org/google.golang.org/grpc/metadata#FromIncomingContext)(ctx).[Get](https://github.com/grpc/grpc-go/tree/master/metadata/metadata.go#L100)("authorization").

EdSchouten:
Let’s close this specific PR, as the author will likely not update this.

PR #8, opened 2019-04-27
Port to windows

moroten:
This change ports pkg/filesystem/local_directory.go to Windows
    
To handle long paths, the paths stored in UNC format by using the prefix \\?\.

When removing files and directories, a retry sequence has been implemented to avoid problems with anti virus programs locking files.

This is part of buildbarn/bb-remote-execution#7

### DIFF
--- a/pkg/filesystem/local_directory_test.go
+++ b/pkg/filesystem/local_directory_test.go
@@ -13,12 +13,12 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-func openTmpDir(t *testing.T) filesystem.Directory {
+func openTmpDir(t *testing.T) (filesystem.Directory, string) {
 	p := filepath.Join(os.Getenv("TEST_TMPDIR"), t.Name())
 	require.NoError(t, os.Mkdir(p, 0777))
 	d, err := filesystem.NewLocalDirectory(p)
 	require.NoError(t, err)
-	return d
+	return d, p
 }
 
 func TestLocalDirectoryCreationFailure(t *testing.T) {
@@ -27,12 +27,14 @@ func TestLocalDirectoryCreationFailure(t *testing.T) {
 }
 
 func TestLocalDirectoryCreationSuccess(t *testing.T) {
-	d := openTmpDir(t)
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 	require.NoError(t, d.Close())
 }
 
 func TestLocalDirectoryEnterBadName(t *testing.T) {
-	d := openTmpDir(t)
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 
 	// Empty filename.
 	_, err := d.Enter("")
@@ -50,14 +52,16 @@ func TestLocalDirectoryEnterBadName(t *testing.T) {
 }
 
 func TestLocalDirectoryEnterNonExistent(t *testing.T) {
-	d := openTmpDir(t)
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 	_, err := d.Enter("nonexistent")
 	require.True(t, os.IsNotExist(err))
 	require.NoError(t, d.Close())
 }
 
 func TestLocalDirectoryEnterFile(t *testing.T) {
-	d := openTmpDir(t)
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 	f, err := d.OpenFile("file", os.O_CREATE|os.O_WRONLY, 0666)
 	require.NoError(t, err)
 	require.NoError(t, f.Close())
@@ -67,7 +71,8 @@ func TestLocalDirectoryEnterFile(t *testing.T) {
 }
 
 func TestLocalDirectoryEnterSymlink(t *testing.T) {
-	d := openTmpDir(t)
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 	require.NoError(t, d.Symlink("/", "symlink"))
 	_, err := d.Enter("symlink")
 	require.Equal(t, syscall.ENOTDIR, err)
@@ -75,7 +80,8 @@ func TestLocalDirectoryEnterSymlink(t *testing.T) {
 }
 
 func TestLocalDirectoryEnterSuccess(t *testing.T) {
-	d := openTmpDir(t)
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 	require.NoError(t, d.Mkdir("subdir", 0777))
 	sub, err := d.Enter("subdir")
 	require.NoError(t, err)
@@ -84,7 +90,8 @@ func TestLocalDirectoryEnterSuccess(t *testing.T) {
 }
 
 func TestLocalDirectoryLinkBadName(t *testing.T) {
-	d := openTmpDir(t)
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 
 	// Invalid source name.
 	require.Equal(t, status.Error(codes.InvalidArgument, "Invalid filename: \"\""), d.Link("", d, "file"))
@@ -102,20 +109,23 @@ func TestLocalDirectoryLinkBadName(t *testing.T) {
 }
 
 func TestLocalDirectoryLinkNotFound(t *testing.T) {
-	d := openTmpDir(t)
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 	require.Equal(t, syscall.ENOENT, d.Link("source", d, "target"))
 	require.NoError(t, d.Close())
 }
 
 func TestLocalDirectoryLinkDirectory(t *testing.T) {
-	d := openTmpDir(t)
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 	require.NoError(t, d.Mkdir("source", 0777))
 	require.True(t, os.IsPermission(d.Link("source", d, "target")))
 	require.NoError(t, d.Close())
 }
 
 func TestLocalDirectoryLinkTargetExists(t *testing.T) {
-	d := openTmpDir(t)
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 	f, err := d.OpenFile("source", os.O_CREATE|os.O_WRONLY, 0666)
 	require.NoError(t, err)
 	require.NoError(t, f.Close())
@@ -127,7 +137,8 @@ func TestLocalDirectoryLinkTargetExists(t *testing.T) {
 }
 
 func TestLocalDirectoryLinkSuccess(t *testing.T) {
-	d := openTmpDir(t)
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 	f, err := d.OpenFile("source", os.O_CREATE|os.O_WRONLY, 0666)
 	require.NoError(t, err)
 	require.NoError(t, f.Close())
@@ -136,7 +147,8 @@ func TestLocalDirectoryLinkSuccess(t *testing.T) {
 }
 
 func TestLocalDirectoryLstatBadName(t *testing.T) {
-	d := openTmpDir(t)
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 	_, err := d.Lstat("")
 	require.Equal(t, status.Error(codes.InvalidArgument, "Invalid filename: \"\""), err)
 	_, err = d.Lstat(".")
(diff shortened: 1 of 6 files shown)


moroten:
Comment from @EdSchouten in an email discussion:
> So to get this straight, I suspect you've implemented a local_directory_windows.go that is more pathname oriented, as opposed to the existing one that is file descriptor oriented. Though I agree this is the easiest to achieve and wouldn't necessarily be opposed to it, it means the implementation doesn't 'fix' the issues that the interface was trying to solve. Namely, the POSIX implementation can traverse the file system (mostly?) race-free and can deal with arbitrary pathname lengths.

My goal was just to get something to work and deal with the problems when or if they arise. I think the long UNC paths should be enough for most users.

> It turns out that on Windows, this can also be achieved. Just like *at() on POSIX systems, Windows NtCreateFile() can also open files using directory file descriptors. It turns out there is a Go package that provides bindings for these system calls:
> https://github.com/hillu/go-ntdll

This is totally new for me. I would prefer someone with more knowledge into this API to help out if it is wanted to go this way.

Regarding [`batched_store_blob_access.go`](https://github.com/buildbarn/bb-remote-execution/blob/74053bb8046e175f5b31617fe7defdef187af1c5/pkg/blobstore/batched_store_blob_access.go) closing files after [`action_digest_subdirectory_manager.go`](https://github.com/buildbarn/bb-remote-execution/blob/74053bb8046e175f5b31617fe7defdef187af1c5/pkg/environment/action_digest_subdirectory_manager.go) is removing the input directory structure:
> If we're going into the direction of using these functions, we could at the same time tackle the issue you experienced. By opening files using `FILE_SHARE_DELETE`, we can hold on to references after deletion. Some GSoC project, Boost AFIO, has some interesting notes on that:
> https://boostgsoc13.github.io/boost.afio/doc/html/afio/quickstart/mini_programs/filesystem_races.html
> https://boostgsoc13.github.io/boost.afio/doc/html/afio/FAQ/deleting_open_files.html

I don't think `FILE_SHARE_DELETE` will solve the problem as I assume the directories will still be locked. After thinking a while again, I would suggest to move the input directory population out from [`local_build_executor.go`](https://github.com/buildbarn/bb-remote-execution/blob/74053bb8046e175f5b31617fe7defdef187af1c5/pkg/builder/local_build_executor.go#L252) into a new `ManagedEnvironment` layer. Thereby, the flushing to CAS can be done before the input directory provider is removing the files. More on that in slack discussion and a separate pull request.

pkg/filesystem/local_directory_test.go
```diff
@@ -13,12 +14,12 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-func openTmpDir(t *testing.T) filesystem.Directory {
+func openTmpDir(t *testing.T) (filesystem.Directory, string) {
```
EdSchouten: Out of curiosity: did you run into issues with these tests leaving files behind on disk? The number of files left behind is not significant and Bazel will clean it up for us, right?

pkg/filesystem/local_directory_test.go
```diff
@@ -236,31 +258,38 @@ func TestLocalDirectoryOpenFileExistent(t *testing.T) {
...
+	d, p := openTmpDir(t)
+	defer os.RemoveAll(p)
 	_, err := d.OpenFile("file", os.O_RDONLY, 0666)
 	require.True(t, os.IsNotExist(err))
 	require.NoError(t, d.Close())
 }
 
 func TestLocalDirectoryOpenFileSymlink(t *testing.T) {
-	d := openTmpDir(t)
+	if runtime.GOOS == "windows" {
```
EdSchouten: What's the reason this test won't work on Windows? Is there no way to prevent it from following junction points? If not, can we at least document this here?

pkg/filesystem/local_directory_unix.go
```diff
@@ -13,6 +14,16 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const (
+	// Permissions do not have one-to-one mapping on Windows. Trivial mapping under Unix.
+	FileMode644 = 0644
```
EdSchouten: Instead of going into this direction, I'd rather see that `pkg/filesystem` is implemented in such a way that it only provides the actual subset of stuff that we need.

What if we were to change `FileInfo` from this:

```
type FileInfo interface {
	Name() string
	Mode() os.FileMode
}
```

To this:

```
type FileInfo interface {
	Name() string
	Type() os.FileMode // Just the file type; not the permission bits. No need to mask.
	IsExecutable() bool
}
```

Doing this would allow us to implement this in a more platform agnostic way and will also clean up some of the callers.

I suspect that this will also allow us to remove the `Umask()` calls from the unit tests, which were only needed to make exact comparisons against permission bits work.

pkg/filesystem/local_directory_unix.go
```diff
@@ -24,6 +35,10 @@ func validateFilename(name string) error {
 	return nil
 }
 
+func PortableSyscallUmask(mask int) (oldmask int) {
```
EdSchouten: In this case we only call this function with zero and never use the return value. What if we were to call it this?

```go
func ClearUmask() {
	syscall.Umask(0)
}
```

pkg/filesystem/local_directory_windows.go
```diff
@@ -0,0 +1,261 @@
...
+)
+
+const uncPrefix = `\\?\`
+
+type localDirectory struct {
+	longAbsPath string
+}
+
+func validateFilename(name string) error {
+	if name == "" || name == "." || name == ".." || strings.ContainsRune(name, '/') {
```
EdSchouten: `|| strings.ContainsRune(name, '\\')` ?

pkg/filesystem/local_directory_windows.go
```diff
@@ -0,0 +1,261 @@
...
+	return nil
+}
+
+func PortableSyscallUmask(mask int) () {
+	// No implementation on Windows
+}
+
+// NewLocalDirectory creates a directory handle that corresponds to a
+// local path on the system.
+func NewLocalDirectory(path string) (Directory, error) {
```
EdSchouten: So the idea behind `NewLocalDirectory()` is that it's this one-time 'hatch' for converting a local directory to something managed by the Filesystem layer. What you pass in here is effectively machine-dependent. This is why I'm not fully convinced you need all of the logic below.

Why can't we simply say: on Windows, NewLocalDirectory() **MUST** be called with a UNC path?

pkg/filesystem/local_directory_windows.go
```diff
@@ -0,0 +1,261 @@
...
+
+	f, err := os.OpenFile(fullPath, flag, perm)
+	if err != nil {
+		return nil, err
+	}
+	return f, nil
+}
+
+func (d *localDirectory) ReadDir() ([]FileInfo, error) {
+	// Obtain filenames in current directory.
```
EdSchouten: On UNIX, I had to hand-roll this function, because I needed `*at()` calls. Because you're dealing with pathname strings, I suspect you can just call into https://golang.org/pkg/io/ioutil/#ReadDir ?

pkg/filesystem/local_directory_windows.go
```diff
@@ -0,0 +1,261 @@
...
+		longAbsPath: longAbsPath,
+	}
+	return d, nil
+}
+
+func (d *localDirectory) validateFilenameAndJoin(name string) (string, error) {
+	if err := validateFilename(name); err != nil {
+		return "", err
+	}
+	// path.Join uses the wrong path separator, which is critical for UNC paths. Join manually.
```
EdSchouten: What about `filepath.Join()`?

pkg/filesystem/local_directory_windows.go
EdSchouten: Though I suspect that this no longer applies in case we alter `FileInfo` to diverge from `os.FileInfo`.

pkg/filesystem/local_directory_unix.go
EdSchouten: Alternatively: considering that `IsExecutable` only applies to regular file, we could maybe even go so far as to come up with our own enum:

```go
type FileType int

const (
	FileTypeRegularFile FileType = iota
	FileTypeExecutableFile
	FileTypeDirectory
	FileTypeSymlink
	FileTypeUnknown
)

type FileInfo interface {
	Name() string
	Type() FileType
}

pkg/filesystem/local_directory_test.go
moroten: Bazel will clean up, but I was running the tests manually. (I was actually cross compiling on Linux and then executing the tests manually on Windows.)

pkg/filesystem/local_directory_windows.go
moroten: There is a need for a helper function somewhere, so I thought it was more convenient to keep all UNC logic inside this file than spreading it out.

(By the way, I should rename some of the `longAbsPath` to `UNC`.)

pkg/filesystem/local_directory_windows.go
moroten: Thank you for pointing it out! (This is my first Go code ever, so I need help to find the right APIs.)

pkg/filesystem/local_directory_test.go
moroten: What is the test actually about on Linux? I lost the plot with `/etc/passwd` and `syscall.ELOOP`.

EdSchouten: That calling OpenFile() on a symlink doesn't actually follow the symlink. It should fail, as we shouldn't allow bb-worker to be tricked into follow them. Symlinks should be treated as ordinary objects in the file system.

`/etc/passwd` was chosen here as an example; any pathname string could be used here. I just thought that `/etc/passwd` was a good example of a file that should obviously not be accessible.

moroten:
> > I don't think FILE_SHARE_DELETE will solve the problem as I assume the directories will still be locked.
> 
> Could this be solved by renaming files to some temporary directory prior to deleting them?

Is there any problem with deleting the files and directories after all outputs have been uploaded?

EdSchouten:
> Is there any problem with deleting the files and directories after all outputs have been uploaded?

There is not really any problem in this specific case. You managed to write a patch that solves this specific issue, but at the cost of introducing some more cross-layer logic. My fear is that because Windows is a platform that is not used that often by most Buildbarn users, there is a fair chance that we'll introduce more bugs like these in the future if the current approach is chosen. It's basically an instance of [Hyrum's law](http://www.hyrumslaw.com/): 

"With a sufficient number of users of an API, it does not matter what you promise in the contract: all observable behaviors of your system will be depended on by somebody."

This is why I argue that we should try to make the local filesystem implementation on Windows as consistent with the UNIX implementation in terms of semantics as much as possible. Because there is likely no way for us to design the API in such a way that we can provide compile-time guarantees that files are always closed prior to deletion, and because Windows does provide facilities to automatically delete files upon closure, I think we should go the extra mile and use that.

There are other places in the Buildbarn source code where support for deleting files that are in use could come in handy in the future. For example, a [Log-structured Merge Tree (LSM)](https://en.wikipedia.org/wiki/Log-structured_merge-tree) backend for BlobAccess could be a lot simpler in case you could rely on the garbage collector to close file descriptors, as opposed to closing them explicitly. When doing that, you need support for deleting files that are in use anyway, as you can't rely on the garbage collector having run.

mickael-carl:
@moroten did you make any progress on this/are still interested in it?

moroten:
Yes, it is still of interest but going in low speed forward. Conclusions so far is that I have too many action input files to make hardlinking work fast enough on Windows, so FUSE filesystem is needed. I'm working on implementing it using https://github.com/billziss-gh/cgofuse/.

I have still not verified if `FILE_SHARE_DELETE` solves the problem also for deleting directories. Therefore, I don't think this PR is ready to merge.

moroten:
I thought I had answered before my holiday, but apparently not.

I am still interested in porting to Windows but have other tasks ongoing in parallel. Due to slow hard-linking performance on Windows, I've put some effort in creating a FUSE file system using https://github.com/billziss-gh/cgofuse. The files actually needed in each action will be downloaded on demand.

Expect updates in a few weeks time.

moroten:
> Just like *at() on POSIX systems, Windows NtCreateFile() can also open files using directory file descriptors.

https://docs.microsoft.com/en-us/windows/win32/devnotes/calling-internal-apis states

> The functions and structures in Winternl.h are internal to the operating system and subject to change from one release of Windows to the next, and possibly even between service packs for each release.

Shall I make an effort into porting to using the `NtCreateFile` via https://github.com/hillu/go-ntdll?

Shall I also handle file deletions before closing the files to avoid changing the order files are created, closed and deleted in the current implementation?

EdSchouten:
Closing this PR due to inactivity. Note that I have added a stub for Windows and enabled CI builds. That way we can be certain that we don't accidentally add even more UNIX-isms to the codebase. Feel free to reopen/file new PRs when needed.

PR #9, opened 2019-04-29
Add a Google Cloud Storage backend for Buildbarn.

mhines99:
This backend requires auth to be provided via config file (JWT token) or will try to use google default credentials.

Example config:

content_addressable_storage {
  gcs {
    bucket: "<mybucket>"
    key_prefix: "prefix/path"
    auth: "./jwt.json"
  }
}
action_cache {
  gcs {
    bucket: "<mybucket>"
    key_prefix: "prefix/other_path"
    auth: "./jwt.json"
  }
}

### DIFF
--- a/pkg/blobstore/configuration/create_blob_access.go
+++ b/pkg/blobstore/configuration/create_blob_access.go
@@ -1,6 +1,7 @@
 package configuration
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -15,15 +16,15 @@ import (
 	"github.com/buildbarn/bb-storage/pkg/blobstore/circular"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/sharding"
 	"github.com/buildbarn/bb-storage/pkg/filesystem"
-	pb "github.com/buildbarn/bb-storage/pkg/proto/blobstore"
 	"github.com/buildbarn/bb-storage/pkg/util"
 	"github.com/go-redis/redis"
 	"github.com/golang/protobuf/proto"
-	"github.com/grpc-ecosystem/go-grpc-prometheus"
-
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+
+	pb "github.com/buildbarn/bb-storage/pkg/proto/blobstore"
+	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 )
 
 // CreateBlobAccessObjectsFromConfig creates a pair of BlobAccess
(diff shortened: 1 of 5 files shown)


pkg/proto/blobstore/blobstore.proto
```diff
@@ -159,3 +162,14 @@ message SizeDistinguishingeBlobAccessConfiguration {
...
+
+message GCSBlobAccessConfiguration {
+       // Name of the GCS bucket.
+       string bucket = 1;
+   
+       // Prefix for keys, e.g. 'bazel_cas/'.
+       string key_prefix = 2;
+
+       // Auth JWT location.
+       string auth = 3;
```
EdSchouten: Pie in the sky: at some point I'd like to see if we could add support for dynamic load balancing/sharding/storage reconfiguration through a separate health checker service. To make this easier to use, more flexible, etc. I don't think it would make sense to source credentials and such from local files. Instead, we should simply place them in the configuration directly (e.g., using a `bytes` field).

If we still want the ability to split up the configuration across multiple files, we could look into something like [Jsonnet](https://jsonnet.org/).

pkg/blobstore/configuration/create_blob_access.go
```diff
@@ -15,15 +16,15 @@ import (
...
 	"github.com/go-redis/redis"
 	"github.com/golang/protobuf/proto"
-	"github.com/grpc-ecosystem/go-grpc-prometheus"
-
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+
+	pb "github.com/buildbarn/bb-storage/pkg/proto/blobstore"
+	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
```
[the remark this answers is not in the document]
mhines99: style thing.

seperate builtins from "repo" from renamed or "_" imports

pkg/proto/blobstore/blobstore.proto
mhines99: I actually started with that but as this is config file it was going to annoying to "stick" bytes into the config file

In general i go for the smallest change to achieve the goal.  So I am open to doing abit more here but would want a concrete answer for on disk how does this config file get loaded with the right bytes

mhines99:
> Hi Marcus,
> 
> First of all, thanks for working on this! \o/
> 
> The existing S3 bucket implementation that is in Buildbarn was written a pretty long time ago; mid last year I guess? Since that time, the Go/Google folks announced the [Go Cloud Development Kit](https://gocloud.dev/); basically a set of wrappers around S3, GCS, Azure buckets, etc. implementing a common interface.
> 
> Instead of going into the direction of adding a GCS blob access, what are your thoughts on simply tossing out the existing S3 backend entirely, replacing it with a single `CloudBucketBlobAccess` that can be configured to use {S3,GCS,Azure}?

So that actually is exactly what my next cl's were going to do is refactor your blob access into that. I didn't know that was released. I am happy to do that cl next if you want to switch to using that for blob access

Shouldn't take me long.  I also really like that they have a good testing story for that library

mhines99:
ping?

EdSchouten:
Hi Marcus,

Feel free to make such a change as part of this pull request. I don't think it makes a lot of sense to have a stop in between. Thanks!

mhines99:
I don't agree with that - i have working cluster using this as a solution

though i didn't have to change your containers from being built with a raw image to using the distroless/base image which includes CA cert

having a working intermediate solution is better than trying to investigate a "potential" larger change that may not even work out

EdSchouten:
> Having a working intermediate solution is better than trying to investigate a "potential" larger change that may not even work out

That may well be the case, but at the same time it also doesn't make sense to upstream a change like this without investigating alternatives.

PR #10, opened 2019-05-02
add opencensus Instrumentation

apesternikov:
Also allow to build a container image on mac and specify registry/repo/tag for push


WORKSPACE
```diff
@@ -26,32 +26,34 @@ http_archive(
     urls = ["https://github.com/bazelbuild/rules_docker/archive/v0.7.0.tar.gz"],
 )
 
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
EdSchouten: All `WORKSPACE`, `BUILD.bazel` and `*.bzl` files are formatted using [Buildifier](https://github.com/bazelbuild/buildtools/tree/master/buildifier). Could you please rerun this? Thanks!

EdSchouten: For `BUILD.bazel` files it's good to know that a `bazel run //:gazelle` already reformats those.

cmd/bb_storage/main.go
```diff
@@ -40,6 +52,32 @@ func main() {
 		log.Fatal(http.ListenAndServe(*webListenAddress, nil))
 	}()
 
+	pe, err := prometheus_exporter.NewExporter(prometheus_exporter.Options{
```
EdSchouten: We likely want to add functionality like this to other bb-* processes as well, right? If so, could we please add this to some `pkg/*`?

cmd/bb_storage/main.go
```diff
@@ -25,8 +33,12 @@ import (
 
 func main() {
 	var (
-		blobstoreConfig  = flag.String("blobstore-config", "/config/blobstore.conf", "Configuration for blob storage")
-		webListenAddress = flag.String("web.listen-address", ":80", "Port on which to expose metrics")
+		blobstoreConfig      = flag.String("blobstore-config", "/config/blobstore.conf", "Configuration for blob storage")
+		webListenAddress     = flag.String("web.listen-address", ":80", "Port on which to expose metrics")
+		agentEndpointURI     = flag.String("jaeger.agent-endpoint", "127.0.0.1:6831", "jaeger agent address")
+		collectorEndpointURI = flag.String("jaeger.collector-endpoint", "http://127.0.0.1:14268/api/traces", "jaeger collector endpoint")
+		serviceName          = flag.String("service-name", "bb_storage", "service name for tracing")
```
EdSchouten: These flags are also specific to Jaeger, right? Could we prefix those with `jaeger.` as well?

While there, please start sentences with uppercase.

cmd/bb_storage/main.go
```diff
@@ -15,8 +15,16 @@ import (
 	"github.com/buildbarn/bb-storage/pkg/cas"
 	"github.com/buildbarn/bb-storage/pkg/util"
 	"github.com/grpc-ecosystem/go-grpc-prometheus"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
+	prometheus_exporter "contrib.go.opencensus.io/exporter/prometheus"
```
EdSchouten: The other of imports has so far been to sort alphabetically by second-level domain. First comes `github.com`, followed by `golang.org`. Could we put `opencensus.io` after that?

third_party/com_github_census_instrumentation_opencensus_proto.patch
```diff
@@ -0,0 +1,45 @@
+--- gen-go/agent/metrics/v1/BUILD.bazel.orig	2019-04-30 11:03:46.000000000 -0700
```
EdSchouten: Could you clarify why this patch is needed? Is this something that should also be upstreamed?

tools/workspace-status.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/sh -e
```
EdSchouten: I'd rather not see this change and the one to `container.bzl` land. The idea is that `container_push_official()` is just meant for... well, pushing official container images. This script is also explicitly made to only work on a clean copy of `master`. We never want to push anything else.

If people want to push bb-* containers into their own registries, they should come up with their own `container_push()` directives. People can always make their own internal repository that has bb-* as externals, containing nothing more than a sequence of `container_push()` directives for `@com_github_buildbarn_bb_storage//cmd/bb_storage:bb_storage_container`.

cmd/bb_storage/BUILD.bazel
```diff
@@ -30,10 +37,12 @@ go_binary(
     visibility = ["//visibility:public"],
 )
 
-container_image(
+go_image(
     name = "bb_storage_container",
-    entrypoint = ["/bb_storage"],
-    files = [":bb_storage"],
+    embed = [":go_default_library"],
+    goarch = "amd64",
```
mickael-carl: Is this really required? According to [documentation](https://github.com/bazelbuild/rules_docker#go_image) both `goarch` and `goos` can be omitted it seems.

cmd/bb_storage/main.go
```diff
@@ -25,8 +33,12 @@ import (
...
 func main() {
 	var (
-		blobstoreConfig  = flag.String("blobstore-config", "/config/blobstore.conf", "Configuration for blob storage")
-		webListenAddress = flag.String("web.listen-address", ":80", "Port on which to expose metrics")
+		blobstoreConfig      = flag.String("blobstore-config", "/config/blobstore.conf", "Configuration for blob storage")
+		webListenAddress     = flag.String("web.listen-address", ":80", "Port on which to expose metrics")
+		agentEndpointURI     = flag.String("jaeger.agent-endpoint", "127.0.0.1:6831", "jaeger agent address")
+		collectorEndpointURI = flag.String("jaeger.collector-endpoint", "http://127.0.0.1:14268/api/traces", "jaeger collector endpoint")
+		serviceName          = flag.String("service-name", "bb_storage", "service name for tracing")
+		alwaysSample         = flag.Bool("always-sample", false, "record all traces. warning! do not use this in production!")
```
mickael-carl: Not sure if the warning is required, the [doc](https://godoc.org/go.opencensus.io/trace#AlwaysSample) already mentions that constantly sampling is likely to generate a lot of data, and in context, it is fairly easy to deduce that it might not be a good idea in a heavily used production service.

cmd/bb_storage/main.go
```diff
@@ -40,6 +52,32 @@ func main() {
 		log.Fatal(http.ListenAndServe(*webListenAddress, nil))
 	}()
 
+	pe, err := prometheus_exporter.NewExporter(prometheus_exporter.Options{
+		Namespace: "bb_storage",
+		Registry: prometheus.DefaultRegisterer.(*prometheus.Registry),
+	})
+	if err != nil {
+		log.Fatalf("Failed to create the Prometheus stats exporter: %v", err)
```
mickael-carl: Nitpicking  but the rest of the code uses log.Fatal exclusively, maybe sticking with it for consistency sake would be better?

cmd/bb_storage/main.go
```diff
@@ -40,6 +52,32 @@ func main() {
...
+	}
+	view.RegisterExporter(pe)
+	if err := view.Register(ocgrpc.DefaultServerViews...); err != nil {
+		log.Fatalf("Failed to register ocgrpc server views: %v", err)
+	}
+	zpages.Handle(nil, "/debug")
```
mickael-carl: Maybe allow this endpoint to be configurable?

cmd/bb_storage/main.go
```diff
@@ -40,6 +52,32 @@ func main() {
...
+	je, err := jaeger.NewExporter(jaeger.Options{
+		AgentEndpoint:          *agentEndpointURI,
+		CollectorEndpoint:      *collectorEndpointURI,
+		ServiceName:            *serviceName,
```
mickael-carl: See [documentation](https://godoc.org/contrib.go.opencensus.io/exporter/jaeger#Options): ServiceName is deprecated and [Process](https://godoc.org/contrib.go.opencensus.io/exporter/jaeger#Process) should be used instead.
It might be worth noting that basic auth is also then not supported. My guess is that it isn't used that much, but maybe this should be either documented or made configurable as well?

pkg/blobstore/circular/circular_blob_access.go
```diff
@@ -59,37 +61,49 @@ func NewCircularBlobAccess(offsetStore OffsetStore, dataStore DataStore, stateSt
...
+	span.Annotatef(nil, "Store allocated, offset %d", offset)
 
 	// Write the data to storage.
 	if err := ba.dataStore.Put(r, offset); err != nil {
 		return err
 	}
 
+	span.Annotate(nil, "Obtaining lock")
 	ba.lock.Lock()
+	span.Annotate(nil, "Lock obtained, calling GetCursors")
```
mickael-carl: Nit: stay consistent with [previous annotation](https://github.com/buildbarn/bb-storage/pull/10/files#diff-b7987a20374b1951b942f7bacdd9f27aR67).

cmd/bb_storage/BUILD.bazel
apesternikov: it is required if building machine is not linux/amd64. This is required to allow to build docker images on mac machine.

cmd/bb_storage/main.go
apesternikov: thank you for the  great catch

pkg/blobstore/circular/circular_blob_access.go
apesternikov: Could you please be more specific?

cmd/bb_storage/main.go
apesternikov: What is the use case? It is not really different from pprof endpoints, which is configured by default.

pkg/blobstore/circular/circular_blob_access.go
mickael-carl: Taking a look at [this annotation](https://github.com/buildbarn/bb-storage/pull/10/files#diff-b7987a20374b1951b942f7bacdd9f27aR67), and [this one](https://github.com/buildbarn/bb-storage/pull/10/files#diff-b7987a20374b1951b942f7bacdd9f27aR103), they are actually on the same logic (i.e. getting a lock and calling getCursors). It would be nice if the annotations were the same, given that the logic stays identical across those traces.

cmd/bb_storage/BUILD.bazel
mickael-carl: You could still achieve this using `bazel build --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //...`, although I'm not sure which is best. If bb-storage were to be ported on more platforms, setting the default to linux_amd64 might not be the best choice anymore. @EdSchouten, thoughts?

EdSchouten: I agree with @mickael-carl here. Even though the container images are most likely to be used on linux/amd64, it's not unrealistic for people to also build them for linux/arm64.

cmd/bb_storage/main.go
EdSchouten: Hardcoding it to `/debug` is fine.

cmd/bb_storage/BUILD.bazel
apesternikov: You still can build containers for the full list of supported architectures using this way:
```
[ go_image(
  name = 'image_' + OS + ' '+ ARCH,
  embed = [":go_default_library"],
  goarch = ARCH,
  goos = OS,
  pure = 'on',
  visibility = ["//visibility:public"],
)  for (ARCH, OS) in [
    ('amd64', 'linux'),
    ('amd64', 'freebsd'),
    ('armv7', 'linux'),
    # add your favorite architecture/os here
  ]
```
Of course you can do this using platforms=.... flag. The big difference, with goarch in BUILD file you DON'T need to specify any flags and still build a any configured architecture.
so instead of 
`bazel build --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //...`
for linux, then 10 other platforms for other architectures, you just do
`bazel build //...`
and build all your configured images for all architectures in one single pass. Moreover, you can push all your containers into a registry using single runnable command.
This is exactly what bazel is moving to as explained in its roadmap here https://bazel.build/roadmaps/configuration.html
the roadmap goal is to make `$ bazel build :all` **just works, for whatever platform(s) you build for** and we, happy users of go and bazel, are fortunate to enjoy it working right now.

apesternikov: let me know if you guys still think you want to stick to command line flags.

tools/workspace-status.sh
apesternikov: ping?

cmd/bb_storage/BUILD.bazel
apesternikov: ping?

tools/workspace-status.sh
mickael-carl: I think it would be best to keep only `container_push_official` within bb-*. The reason for that is twofold:
- Considering adding a parametrized `container_push`, how far should we make it configurable? Likely all the way to stamping. But then you might also want to use [rules_k8s](https://github.com/bazelbuild/rules_k8s) and have some configurable `k8s_deploy` with the same amount of flexibility. In the end this would add quite some work for something that can be achieved by...
- Using your own `WORKSPACE` to have your own `container_push()` as well as `k8s_deploy()` as @EdSchouten pointed out. This will be much easier when the [nested `WORKSPACE`](https://github.com/bazelbuild/bazel/issues/1943) feature lands, as you will be able to have such `WORKSPACE` containing all the pieces of Buildbarn nested within.

While I understand the hassle of creating a separate repository specifically for deploying Buildbarn to your own cluster, I think that eventually it will spare us quite some effort, and it will become easier in the (somewhat) near future.

cmd/bb_storage/BUILD.bazel
mickael-carl: I do see a problem with this approach, namely that this will force us to maintain a list of `(ARCH, OS)` somewhere, if not in multiple places, across several repositories. This will likely end up being quite tedious and error-prone.
Currently docker supports [multi-arch containers](https://docs.docker.com/engine/reference/commandline/manifest/), but this hasn't really been picked up by `rules_docker` at this time. It is however [considered](https://github.com/bazelbuild/rules_docker/issues/300).
Furthermore, you are right that `bazel build :all` should just work, and it might just already, *but*, it is not as efficient currently as what [the design document](https://docs.google.com/document/d/1vc8v-kXjvgZOdQdnxPTaV0rrLxtP2XwnD2tAZlYJOqw/) proposes. Eventually when this lands, you will simply be able to configure a list of targets through a rule, that you will be able to pass to Bazel as build configuration, which is already much cleaner imho.
In the mean time, I think it is best if we stick to command-line flags though, just to avoid adding hard-coded values in build files. You could also specify those platforms [as config in your bazelrc](https://github.com/kubernetes/kubernetes/blob/master/build/root/.bazelrc#L29) as well, for convenience.

mickael-carl:
I think at the moment @EdSchouten is quite busy, so it might take some time until this is merged unfortunately.

tools/workspace-status.sh
apesternikov: I rolled back this change.

cmd/bb_storage/BUILD.bazel
apesternikov: rolled back

apesternikov:
I believe all questions are resolved at this point, can we proceed?

mickael-carl:
I’ve pinged @EdSchouten but he is still unavailable for the time being unfortunately 😕

cmd/bb_storage/BUILD.bazel
mickael-carl: I think this wasn't reverted in the end? Could you do so as well? Then I can merge it finally!

apesternikov: oops sorry for that. it was only partially reverted. fixed now.

PR #12, opened 2019-05-10
TLS support in bb-storage grpc server

ailiev:


pkg/proto/configuration/bb_storage/bb_storage.proto
```diff
@@ -38,4 +38,15 @@ message StorageConfiguration {
 
   // List of instances which can upload to the Action Cache.
   repeated string allow_ac_updates_for_instances = 6;
+
+  // TLS configuration, if enabled.
+  TLSConfiguration tls = 7;
+}
+
+// Used for both GRPC and HTTP listeners.
+message TLSConfiguration {
```
ailiev: Looks like this should be in a separate proto, so it can be used from the other components.
Will extract it.

pkg/proto/configuration/server/server.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package buildbarn.configuration.server;
```
ailiev: I would suggest having the grpc and http/metrics listen ports in here too.

mickael-carl: Other components importing this then would import the ports message as well which could be a bit confusing right? Also right now configuration is consistent across components, so either this would break consistency or you would need a server.proto for each, but in some cases, this would be a bit overkill (e.g. [bb-scheduler configuration](https://github.com/buildbarn/bb-remote-execution/blob/master/pkg/proto/configuration/bb_scheduler/bb_scheduler.proto))

mickael-carl:
Would you mind including the cert and key actually in the config file? That way we can keep configuration files to a minimum, and not add additional logic to read those files

ailiev:
> Would you mind including the cert and key actually in the config file? That way we can keep configuration files to a minimum, and not add additional logic to read those files

Two reasons not to:
- at least the private key is a good candidate to reside in a separate Secret file.
- the TLS APIs actually all take file names, not contents. so we don't do any file reading. Passing in contents is an option in the TLS lib `X509KeyPair()`, and probably for the http flow as well though with more code.

I'm more attached to the first reason, to at least keep the separate files option.

WDYT?

palvarez89:
> > Would you mind including the cert and key actually in the config file? That way we can keep configuration files to a minimum, and not add additional logic to read those files
> 
> Two reasons not to:
> 
> * at least the private key is a good candidate to reside in a separate Secret file.

Totally agree with this one. In this case I think extra "configuration" files are justified.

pkg/util/tls.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Can't load X509 certificate and keypair with config %v: %s",
+			tlsParams, err)
+	}
+	cfg := &tls.Config{Certificates: []tls.Certificate{cert}}
+	creds := credentials.NewTLS(cfg)
+	return creds, nil
+}
+
+// several helpers to help with testing
```
mickael-carl: Not only testing right? Since this is used by HttpListenAndServe which is called in `main.go`

ailiev: Oh yes, both code paths use them, just that without the testing usage, the code here could have been a bit simpler. I can expand the comment.

pkg/util/tls.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+	"net"
+	"net/http"
+
+	pb "github.com/buildbarn/bb-storage/pkg/proto/configuration/server"
+
+	"google.golang.org/grpc/credentials"
+)
+
+// ValidateTls checks params are in order, return non-nil if not.
```
mickael-carl: As a general (nit) note: could you make sure all comments are full phrases without abbreviations? That will make it clearer imho

pkg/util/tls_test.go
```diff
@@ -0,0 +1,108 @@
...
+		t.Logf("serve failed or completed with: %v", serveErr)
+
+	}()
+	defer srv.Close()
+
+	time.Sleep(1 * time.Second)
+
+	// check the server is actually in good shape
+	withTls := cfg != nil
+	var prot string
```
mickael-carl: Nit: I'd prefer to keep the full variable name `protocol`

Qinusty:
What is the state of this PR? It would be nice to get this merged if working.

EdSchouten:
Hi @ailiev! Thanks for your contribution! Sorry for not being able to integrate this change for you. That said, PR #34 contains a commit that adds this feature as well. Let me close this PR, so that we can continue to track this in #34.

PR #13, opened 2019-06-10
Add option to get S3 access key and secret from environment

ailiev:
if configured as env:$name

### DIFF
--- a/pkg/blobstore/configuration/BUILD.bazel
+++ b/pkg/blobstore/configuration/BUILD.bazel
@@ -1,4 +1,4 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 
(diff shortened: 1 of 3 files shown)


mickael-carl:
Out of curiosity, why not use the [aws-sdk](https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/#NewEnvCredentials) to load credentials from environment?

ailiev:
> load credentials automagically from environment, among other default sources.

There is apparently some such capability already in the code, any idea if that would use the standard env vars?
> 			// If AccessKeyId isn't specified, allow AWS to search for credentials.
>			// In AWS EC2, this search will include the instance IAM Role

ailiev:
Confirmed it picks up standard AWS env vars if none are supplied in the barn config.

Thus closing this issue. Possibly refinement may be needed if multiple AWS services need to be used with different credentials, but that's not the case now.

PR #18, opened 2019-06-25
badger backend implementation

apesternikov:
this backend uses https://github.com/dgraph-io/badger


mickael-carl:
Hi there! Thanks for contributing! I'm curious as to what the rationale behind using badger is. We already provide quite a few backends and I think it would be best to keep number of backends low, as this adds extra maintenance burden. Could you explain what would be the advantages compared to existing ones?

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -190,3 +194,12 @@ message SizeDistinguishingBlobAccessConfiguration {
...
   int64 cutoff_size_bytes = 3;
 }
+
+message BadgerBlobAccessConfiguration {
+    // Directory where the files created by the badger storage
+    // backend are located.
+    string directory = 1;
+
+    // TTL for stored objects
+    google.protobuf.Duration ttl = 2;
```
mickael-carl: Could you document here the default value?

pkg/cas/byte_stream_server.go
```diff
@@ -59,6 +60,7 @@ func NewByteStreamServer(blobAccess blobstore.BlobAccess, readChunkSize int) byt
 }
 
 func (s *byteStreamServer) Read(in *bytestream.ReadRequest, out bytestream.ByteStream_ReadServer) error {
+	log.Println("byteStreamServer.Read", in)
```
mickael-carl: Shouldn't this make use of tracing?

pkg/blobstore/configuration/create_blob_access.go
```diff
@@ -66,6 +67,17 @@ func createBlobAccess(configuration *pb.BlobAccessConfiguration, storageType str
 		return nil, errors.New("Storage configuration not specified")
 	}
 	switch backend := configuration.Backend.(type) {
+	case *pb.BlobAccessConfiguration_Badger:
+		backendType = "badger"
+		opts := badger.DefaultOptions
+		opts.Dir = backend.Badger.Directory
```
mickael-carl: Why not use `badger.DefaultOptions(backend.Badger.Directory)` ? According to [documentation](https://godoc.org/github.com/dgraph-io/badger#Options.WithDir) this sets both `Dir` and `ValueDir` to the provided path.

pkg/blobstore/configuration/create_blob_access.go
```diff
@@ -66,6 +67,17 @@ func createBlobAccess(configuration *pb.BlobAccessConfiguration, storageType str
...
+		opts.ValueDir = backend.Badger.Directory
+		opts.SyncWrites = false
```
mickael-carl: Why use async writes? Isn't this a problem for cache consistency?

pkg/blobstore/badger_blob_access.go
```diff
@@ -0,0 +1,160 @@
...
+	if ttl != nil {
+		ba.ttl = time.Duration(ttl.Seconds) * time.Second
+	}
+	go ba.gc()
+	return ba
+}
+
+// Badger storage garbage collection
+func (ba *badgerBlobAccess) gc() {
+	ticker := time.NewTicker(5 * time.Minute)
```
mickael-carl: Out of curiosity, what is the reasoning behind choosing those values for garbage collection? Should they be configurable as well for end user?

pkg/blobstore/badger_blob_access.go
```diff
@@ -0,0 +1,160 @@
...
+	if err := ctx.Err(); err != nil {
+		r.Close()
+		return err
+	}
+	value, err := ioutil.ReadAll(r)
+	r.Close()
+	if err != nil {
+		return util.StatusWrapWithCode(err, codes.Unavailable, "Failed to put blob")
+	}
+	span.Annotate(nil, "creating Transaction")
```
mickael-carl: Nit: `Creating` rather than `creating`

pkg/blobstore/badger_blob_access.go
```diff
@@ -0,0 +1,160 @@
...
+	txn := ba.db.NewTransaction(true)
+	defer txn.Discard()
+	span.Annotate(nil, "Transaction created")
+	err = txn.SetWithTTL([]byte(digest.GetKey(ba.blobKeyFormat)), value, ba.ttl)
+	if err != nil {
+		return util.StatusWrapWithCode(err, codes.Unavailable, "Failed to set blob")
+	}
+	span.Annotate(nil, "Set finished")
+	// Provide a callback so write is asynchronous.
+	err = txn.Commit(func(err error) {
```
mickael-carl: As pointed out, async writes and deletion might break consistency, which is require, afaik.

pkg/blobstore/badger_blob_access.go
```diff
@@ -0,0 +1,160 @@
...
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+	if len(digests) == 0 {
+		return nil, nil
+	}
+	txn := ba.db.NewTransaction(false)
+	defer txn.Discard()
+	var missing []*util.Digest
+	it := txn.NewIterator(badger.IteratorOptions{PrefetchValues: false})
```
mickael-carl: From the [documentation](https://godoc.org/github.com/dgraph-io/badger#Txn.NewIterator): `Using prefetch is recommended if you're doing a long running iteration, for performance.`. What is the reason then to disable it?

EdSchouten:
Closing this PR due to inactivity.

Note that I'm in principle not against adding support for more storage backends, do note that systems like Badger may not be a right fit. Log-structured merge-trees (LSMs) offer semantics that are far stronger than what is required by Buildbarn.

First of all, they are databases, not caches. This means that additional logic/processes needs to be added to constrain size. I think we can all agree that a cache should be simpler in terms of logic than a database. Using an LSM in combination with a garbage collector unfortunately is not.

Second, LSMs are designed to keep track of the key space in sorted order, so that it may be iterated. There is no need to apply any sorting in the case of the CAS and AC, as there simply is no iteration. Instead of an LSM, a hash table may be sufficient.

Since this PR was filed, Buildbarn has gained a new storage backend called LocalBlobAccess. I think time would better be spent either optimizing that one or adding any features that are missing.

PR #20, opened 2019-07-22
Split ContentAddressableStorage interface into reader and writer

moroten:
It is possible to split the `ContentAddressableStorage` interface into a reader and a writer. `ReadWriteDecouplingContentAddressableStorage` already exists and is using one reader and one writer, so this split is a natural refactoring.

### DIFF
--- a/pkg/cas/content_addressable_storage.go
+++ b/pkg/cas/content_addressable_storage.go
@@ -12,13 +12,20 @@ import (
 // ContentAddressableStorage provides typed access to a Bazel Content
 // Addressable Storage (CAS).
 type ContentAddressableStorage interface {
+	ContentAddressableStorageReader
+	ContentAddressableStorageWriter
+}
+
+type ContentAddressableStorageReader interface {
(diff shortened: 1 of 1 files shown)


mickael-carl:
Hi! Thanks for contributing!
I'm just curious what this brings, it seems to be mostly a cosmetic change right now?

moroten:
It's part of my (slow moving) implementation of FUSE for Linux and Windows. It feels annoying to write a `ContentAddressableStorageReader` implementation which at the moment has to implement writer parts using a simple panic.

I've just uploaded https://github.com/buildbarn/bb-remote-execution/pull/17, which is part of https://github.com/moroten/bb-remote-execution/commits/master.

EdSchouten:
Closing this, as I think the right strategy is to simply remove ContentAddressableStorage altogether. More details can be found in this comment @ bb-remote-execution: https://github.com/buildbarn/bb-remote-execution/pull/17#issuecomment-642560987

PR #24, opened 2019-09-09
Support partial read requests from CAS

clintharrison:
During large builds, we run into this unimplemented code path. This should handle this case now, and I've added some tests as well.


EdSchouten:
Hi @clintharrison! Instead of merging this change, my plan is to push #34 forward. This change contains a new abstraction, pkg/blobstore/buffer, that allows us to do offset reads on blobs natively. ByteStreamServer will be able to pass down the offset directly, so there's no need to add any code that does skipping there.

Thanks for your contribution regardless and my apologies for not fixing this issue sooner.

PR #29, opened 2019-10-09
Detect destructive circular cache config change

moroten:
Avoid the user getting surprised that the cache gets totally trashed when enlarging the cache. Error out instead so the user understands the consequence.


EdSchouten:
Thanks for the patch! That said, the plan remains to simply kick CircularBlobAccess out of the tree. Let's not make any efforts to change any of the code that we have right now.

Issue #28, opened 2019-10-05
Build without the bytes with circular cache


moroten:
Using https://blog.bazel.build/2019/05/07/builds-without-bytes.html requires all action outputs to persist during a build. When the circular disk storage gets full, Bazel experiences sporadic failures when blobs are being evicted from the cache.

To solve this, one could rewrite the CAS blobs requested through `Get()` or `FindMissing()`, whenever they get close to being evicted. The threshold can be configured as a percentage threshold.

Action cache requests can also call `FindMissing()` to the CAS for all outputs. As this introduces latency, it should be a configurable option as some users might choose to avoid automatically evicting blobs from their storage.

Any comments or shall I start implementing?

EdSchouten:
CircularBlobAccess implements a circular buffer, meaning it is FIFO-like. Builds without the Bytes requires that storage is LRU-like. This means that CircularBlobAccess is simply unsuitable.

PR #34 adds a new storage backend, LocalBlobAccess, that implements an eviction scheme that should be compatible with Builds without the Bytes. The NewLocalBlobAccess() function has good documentation on how it works under the hood.

The code that is part of #34 will only allow you to use LocalBlobAccess for in-memory storage. The goal is that we work on extending it later on to support on-disk storage, meaning it should become a full replacement for CircularBlobAccess. The goal is that we'll be kicking CircularBlobAccess out of the tree when that happens.

Considering that this issue is about CircularBlobAccess specifically, I think we should close it. We shouldn't attempt to make any improvement in that area. Thanks for reporting it regardless!

PR #33, opened 2019-12-16
Ensure the AC correctly tracks the output blobs associated with an action

adamj-codethink:
This PR mitigates the problem of cache clean-up and cached actions, which particularly effect builds without the bytes.

This is documented here: https://github.com/bazelbuild/bazel/issues/8250

We pass the blobstore to the Action Cache, so that it can query the remote blob stores, to check the associated output digests are still cached for a given action. If not a blob not found response is returned, which will trigger bazel to rebuild that given action.

### DIFF
--- a/cmd/bb_storage/main.go
+++ b/cmd/bb_storage/main.go
@@ -91,7 +91,7 @@ func main() {
 		grpc.UnaryInterceptor(grpc_prometheus.UnaryServerInterceptor),
 		grpc.StatsHandler(&ocgrpc.ServerHandler{}),
 	)
-	remoteexecution.RegisterActionCacheServer(s, ac.NewActionCacheServer(actionCache, allowActionCacheUpdatesForInstances))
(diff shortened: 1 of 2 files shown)


moroten:
This tries to solve #28.

I tried a similar solution a while ago but realized that you need to touch all the blobs in such way that they are not evicted during the build, just checking for existence is not enough. I got into troubles when just doing a simple existence check on a circular storage that was filled.

My solution was to use a larger storage space until implementing a proper solution, which hasn't happened yet. Let us see what code Ed Schouten will release early next year, if that solves our problems.

EdSchouten:
Hey! Yes, the latest bb-storage code has a special CompletenessCheckingBlobAccess that implements this verification step. It can be enabled optionally, because there are cases in which you don’t want to enable it (in sharded setups). Let’s close this PR and continue to track this issue in #28.

EdSchouten:
CompletenessCheckingBlobAccess is part of #34.

PR #35, opened 2020-01-03
Add error in case of missing grpc listener

Qinusty:
Adds a simple error case for a lack of configuration on grpc listeners

### DIFF
--- a/pkg/grpc/grpc.go
+++ b/pkg/grpc/grpc.go
@@ -50,6 +50,11 @@ func NewGRPCClientFromConfiguration(configuration *configuration.GRPCClientConfi
 // network addresses of UNIX socket paths provided.
 func NewGRPCServersFromConfigurationAndServe(configurations []*configuration.GRPCServerConfiguration, registrationFunc func(*grpc.Server)) error {
 	serveErrors := make(chan error)
+
+	if len(configurations) == 0 {
+		return status.Error(codes.InvalidArgument, "Expected GRPC server configuration is missing")
+	}
+
 	for _, configuration := range configurations {
 		// Create an authenticator for requests.
 		authenticator, err := NewAuthenticatorFromConfiguration(configuration.AuthenticationPolicy)
@@ -86,6 +91,10 @@ func NewGRPCServersFromConfigurationAndServe(configurations []*configuration.GRP
 		// Add Prometheus timing metrics.
 		grpc_prometheus.EnableHandlingTimeHistogram(grpc_prometheus.WithHistogramBuckets(prometheus.DefBuckets))
 		grpc_prometheus.Register(s)
+
+		if len(configuration.ListenAddresses)+len(configuration.ListenPaths) == 0 {
+			return status.Error(codes.InvalidArgument,"GRPC server configured without any listen addresses or paths")
+		}
 
 		// TCP sockets.
 		for _, listenAddress := range configuration.ListenAddresses {


pkg/grpc/grpc.go
```diff
@@ -87,6 +88,10 @@ func NewGRPCServersFromConfigurationAndServe(configurations []*configuration.GRP
 		grpc_prometheus.EnableHandlingTimeHistogram(grpc_prometheus.WithHistogramBuckets(prometheus.DefBuckets))
 		grpc_prometheus.Register(s)
 
+		if len(configuration.ListenAddresses)+len(configuration.ListenPaths) == 0 {
+			return errors.New("GRPC server configured without any listen addresses or paths")
```
EdSchouten: Inside the Buildbarn code base we use gRPC error codes, even for non-gRPC service code. The reason for this is that over time, the boundary of what is service code and what is not may shift around.

Instead of using `errors.New()`, could you please use `status.Error(codes.InvalidArgument, ...)`? Thanks!

pkg/grpc/grpc.go
```diff
@@ -87,6 +88,10 @@ func NewGRPCServersFromConfigurationAndServe(configurations []*configuration.GRP
 		grpc_prometheus.EnableHandlingTimeHistogram(grpc_prometheus.WithHistogramBuckets(prometheus.DefBuckets))
 		grpc_prometheus.Register(s)
 
+		if len(configuration.ListenAddresses)+len(configuration.ListenPaths) == 0 {
```
EdSchouten: Out of interest, did you run `gofmt`? I would have expected it to add spaces around the + operator in this case.

Qinusty: This is no longer relevant as I changed this conditional, but I did check and gofmt appears to prefer the spaces removed.

Qinusty:
> This will only cause an error in case a server is created without any bind addresses (which is good).
> 
> I think you were interested in covering another case, namely when no servers are declared. Could you also add a check for that?

Both cases are now covered and tested.

pkg/grpc/grpc.go
```diff
@@ -50,6 +51,11 @@ func NewGRPCClientFromConfiguration(configuration *configuration.GRPCClientConfi
 // network addresses of UNIX socket paths provided.
 func NewGRPCServersFromConfigurationAndServe(configurations []*configuration.GRPCServerConfiguration, registrationFunc func(*grpc.Server)) error {
 	serveErrors := make(chan error)
+
+	if configurations == nil {
```
EdSchouten: `len(configurations) == 0`

It is possible for a slice to have length zero, but not point to a nil memory address.

Qinusty: Fair point, the change now works as expected with `nil` or `[]`

pkg/grpc/grpc.go
```diff
@@ -87,6 +93,10 @@ func NewGRPCServersFromConfigurationAndServe(configurations []*configuration.GRP
 		grpc_prometheus.EnableHandlingTimeHistogram(grpc_prometheus.WithHistogramBuckets(prometheus.DefBuckets))
 		grpc_prometheus.Register(s)
 
+		if len(configuration.ListenAddresses)+len(configuration.ListenPaths) == 0 {
+			return errors.New("GRPC server configured without any listen addresses or paths")
```
EdSchouten: I think this still needs to be changed to `status.Error()` and we're good to go!

PR #38, opened 2020-02-13
Extend opencensus support

arlyon:
In an effort to improve observability (in conjunction with https://github.com/buildbarn/bb-remote-execution/pull/32) I've annotated a number of additional spans. This obviously couples the trancing tightly to the rest of the system, but the functions only actually do anything if opentracing is added in the config.

(the image below is using both this patch and the one referenced on bb-remote-execution)

image


EdSchouten:
Because I started reviewing your be-remote-execution PR first, let's continue discussing the overall approach there. Once that's nailed down, we'll look at this PR to sort out the details.

Awesome work! The screenshots look impressive!

EdSchouten:
Closing this due to inactivity. Be sure to reach out if you are continuing to work on this!

PR #43, opened 2020-03-06
Additional options added to Redis config

Daveo24:
Added additional Redis configs to be used for using Redis as backend:

- Password
- DialTimeout for Redis Client
- ReadTimeout for Redis Client
- WriteTimeout for Redis Client.

**Config changes**: We need to include a similar kind of config as below in storage.yaml (using K8). See the changed blobstore.proto for the configs. (Overridden with Default values, if above configs not present.)


pkg/blobstore/configuration/create_blob_access.go
```diff
@@ -261,12 +261,41 @@ func createBlobAccess(configuration *pb.BlobAccessConfiguration, options *blobAc
 				backend.Redis.ReplicationCount,
 				replicationTimeout)
 		case *pb.RedisBlobAccessConfiguration_Single:
+
+			dialTimeout := 5 * time.Second
```
EdSchouten: Is there any need to provide explicit default values? What about just declaring them like this?
```
var dialTimeout time.Time
```
That way we can simply defer to go-redis for the default values.

pkg/blobstore/configuration/create_blob_access.go
```diff
@@ -261,12 +261,41 @@ func createBlobAccess(configuration *pb.BlobAccessConfiguration, options *blobAc
...
+			if mode.Single.DialTimeout != nil {
+				dialTimeout, err = ptypes.Duration(mode.Single.DialTimeout)
+				if err != nil {
+					return nil, err
```
EdSchouten: Though I know that the rest of this source file is particularly bad at this, could you change the error handling to be a bit like this:
```
return nil, util.StatusWrap(err, "Failed to obtain dial timeout")
```
That way it's more obvious to the user where the error comes from.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -187,6 +187,24 @@ message SingleRedisBlobAccessConfiguration {
 
   // Numerical ID of the database.
   int32 db = 2;
+
+  // Redis Auth Password
```
EdSchouten: Please end sentences with full stops (periods). Also wrap them at 72 characters. Also make sure there is a space after the `//`.

pkg/blobstore/configuration/create_blob_access.go
```diff
@@ -261,12 +261,41 @@ func createBlobAccess(configuration *pb.BlobAccessConfiguration, options *blobAc
 				backend.Redis.ReplicationCount,
```
EdSchouten: It looks like we hardcode these values for the clustered Redis solution. Could you please also use those options there? Be sure to move all of the duration conversion/parsing above the `switch` statement, so that it can be reused in both branches.

pkg/blobstore/configuration/create_blob_access.go
Daveo24: Absolutely, I will change it later on today,.

pkg/blobstore/configuration/create_blob_access.go
Daveo24: There was no reason in the end, I had it exactly as you have pointed out at the start and I was going to set my own default values which I then didn't do so I will change it back.

pkg/proto/configuration/blobstore/blobstore.proto
Daveo24: Thanks, I will follow that standard from now on.

pkg/blobstore/configuration/create_blob_access.go
```diff
@@ -200,22 +200,46 @@ func createBlobAccess(configuration *pb.BlobAccessConfiguration, options *blobAc
 
 		tlsConfig, err := util.NewTLSConfigFromClientConfiguration(backend.Redis.Tls)
 		if err != nil {
-			return nil, err
+			return nil, util.StatusWrap(err, "failed to obtain new tls")
```
EdSchouten: The current convention is to capitalize all error messages, and to make abbreviations uppercase where applicable:
```go
return nil, util.StatusWrap(err, "Failed to obtain TLS configuration")
```

pkg/blobstore/configuration/create_blob_access.go
```diff
@@ -252,9 +276,9 @@ func createBlobAccess(configuration *pb.BlobAccessConfiguration, options *blobAc
 						MaxRetries:      maxRetries,
 						MinRetryBackoff: minRetryDur,
 						MaxRetryBackoff: maxRetryDur,
-						DialTimeout:     10 * time.Second,
-						ReadTimeout:     100 * time.Second,
-						WriteTimeout:    100 * time.Second,
+						DialTimeout:     dialTimeout * time.Second,
```
EdSchouten: I think you have to drop the `* time.Second` here.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -187,6 +187,23 @@ message SingleRedisBlobAccessConfiguration {
 
   // Numerical ID of the database.
   int32 db = 2;
+
+  // Redis Auth Password, "" for NO Password
+  string password = 3;
+
+  // Dial timeout for establishing new connections.
+  // Default is 5 (seconds) or you could select your own seconds '30s'
```
EdSchouten: Maybe it's best to document it in such a way that the actual default value remains undescribed. It would become invalid if go-redis were to change its default value, right?

PR #47, opened 2020-03-09
Implement `BatchReadBlobs`

sdclarke:
In attempting to find the incompatibilities between Buildbarn and [BuildStream](https://gitlab.com/BuildStream/buildstream) I found that one of the issues was the fact that [`BatchReadBlobs`](https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L285-L307) was not implemented in Buildbarn, so this pull requests adds an implementation of it.


pkg/cas/content_addressable_storage_server.go
```diff
@@ -47,7 +47,36 @@ func (s *contentAddressableStorageServer) FindMissingBlobs(ctx context.Context,
 }
 
 func (s *contentAddressableStorageServer) BatchReadBlobs(ctx context.Context, in *remoteexecution.BatchReadBlobsRequest) (*remoteexecution.BatchReadBlobsResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "This service does not support batched reading of blobs")
+	// Asynchronously call Get() for every blob.
+	responsesChan := make(chan *remoteexecution.BatchReadBlobsResponse_Response, len(in.Digests))
+	for _, reqDigest := range in.Digests {
+		go func(reqDigest *remoteexecution.Digest) {
```
EdSchouten: This places no upper bound on the number of concurrent RPCs this would generate. If a caller passes in a list of thousands of digests, this would spawn thousands of goroutines, which could cause thousands of network connections/channels/.... to be opened to storage backends, etc.

For an initial implementation, I would suggest that no concurrency is added. If it turns out that this kills performance, we can always add it later on either by adding a constant limit on concurrency or by pushing support for batch reading/writing into BlobAccess itself. The latter would prevent batched calls from ending up being decomposed in case bb-storage instances are chained/stacked.

I would also suggest that BatchUpdateBlobs() is adjusted accordingly; it seemingly also has no such limit.

pkg/cas/content_addressable_storage_server.go
```diff
@@ -47,7 +47,36 @@ func (s *contentAddressableStorageServer) FindMissingBlobs(ctx context.Context,
...
+			digest, err := digest.NewDigestFromPartialDigest(in.InstanceName, reqDigest)
+			var buf buffer.Buffer
+			if err == nil {
+				buf = s.contentAddressableStorage.Get(
+					ctx,
+					digest)
+			}
+			maxSizeBytes, err := buf.GetSizeBytes()
```
EdSchouten: This calls `GetSizeBytes()` on a nil buffer in case the branch above is not taken.

Also keep in mind that `GetSizeBytes()` is only of use in code that is CAS/AC agnostic. For blobs coming from the CAS, you can simply use `digest.GetSizeBytes()`.

pkg/cas/content_addressable_storage_server.go
```diff
@@ -47,7 +47,36 @@ func (s *contentAddressableStorageServer) FindMissingBlobs(ctx context.Context,
...
+			var data []byte
+			if err == nil {
+				data, err = buf.ToByteSlice(int(maxSizeBytes))
```
EdSchouten: This effectively puts no limit on the amount of memory a call to `BatchReadBlobs()` may consume. If I called `BatchReadBlobs()` with a thousand digests of 1 GB objects, this function would consume 1 TB of memory.

Considering that this is about limiting Protobuf message sizes, you could extend this type to also respect the `maximumMessageSizeBytes` configuration option. Some loop at the beginning of this function should be added to ensure the combined size of all blobs is not larger than `maximumMessageSizeBytes`.

sdclarke: Would you be happy with `"Attempted to read %d bytes, while a maximum of %d bytes is permitted"` as the error message when exceeding maximumMessageSizeBytes?

EdSchouten: "Attempted to read a total of %d bytes, ..." maybe, to make it more explicit that it's the aggregate that counts?

pkg/cas/content_addressable_storage_server.go
```diff
@@ -47,62 +49,58 @@ func (s *contentAddressableStorageServer) FindMissingBlobs(ctx context.Context,
...
+			totalSize, s.maximumMessageSizeBytes)
 	}
 
-	// Recombine results.
 	var response remoteexecution.BatchReadBlobsResponse
-	for i := 0; i < len(in.Digests); i++ {
-		response.Responses = append(response.Responses, <-responsesChan)
+	for _, reqDigest := range in.Digests {
+		digest, err := digest.NewDigestFromPartialDigest(in.InstanceName, reqDigest)
+		var buf buffer.Buffer
```
EdSchouten: This variable may be removed if the line below used `buf :=`. Alternatively: just append the `.ToByteSlice()` call to the `.Get()` call. There are already many other places where that's done.

pkg/cas/content_addressable_storage_server_test.go
```diff
@@ -0,0 +1,88 @@
...
+	buf := buffer.NewValidatedBufferFromByteSlice(a)
+	contentAddressableStorage.EXPECT().Get(ctx, digest1).Return(buf)
+	b := make([]byte, 234)
+	buf2 := buffer.NewValidatedBufferFromByteSlice(b)
+	contentAddressableStorage.EXPECT().Get(ctx, digest2).Return(buf2)
+
+	contentAddressableStorageServer := cas.NewContentAddressableStorageServer(contentAddressableStorage, 1<<16)
+
+	response, err := contentAddressableStorageServer.BatchReadBlobs(ctx, request)
+	if err != nil {
```
EdSchouten: `require.NoError()`

pkg/cas/content_addressable_storage_server_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func TestContentAddressableStorageServerBatchReadBlobsSuccess(t *testing.T) {
+	ctrl, ctx := gomock.WithContext(context.Background(), t)
+	defer ctrl.Finish()
+
+	digest1 := digest.MustNewDigest("ubuntu1804", "0000000000000000000000000000000000000000000000000000000000000001", 123)
```
EdSchouten: With the unit tests I always try to make them representative of what may happen in real life. Because many of the tests use `NewValidatedBufferFromByteSlice()`, we can get away with using digests that don't match up with the data, but it's not what may be possible in practice.

Could you please change these digests to the following?
```go
digest1 := digest.MustNewDigest("ubuntu1804", "409a7f83ac6b31dc8c77e3ec18038f209bd2f545e0f4177c2e2381aa4e067b49", 123)
digest2 := digest.MustNewDigest("ubuntu1804", "0479688f99e8cbc70291ce272876ff8e0db71a0889daf2752884b0996056b4a0", 234)
```

pkg/cas/content_addressable_storage_server_test.go
```diff
@@ -0,0 +1,88 @@
...
+	digest2 := digest.MustNewDigest("ubuntu1804", "0000000000000000000000000000000000000000000000000000000000000002", 234)
+
+	request := &remoteexecution.BatchReadBlobsRequest{
+		Digests: []*remoteexecution.Digest{
+			digest1.GetPartialDigest(),
```
EdSchouten: Nit:

I'm always a big fan of this: https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

Calls to functions like `.GetPartialDigest()` often make it harder to understand what's actually being passed around under the hood. It would have been perfectly fine to put `&remoteexecution.Digest{}` literals here instead of calling `.GetPartialDigest()`.

pkg/cas/content_addressable_storage_server_test.go
```diff
@@ -0,0 +1,88 @@
...
+	response, err := contentAddressableStorageServer.BatchReadBlobs(ctx, request)
+	if err != nil {
+		t.Errorf("%v", err)
+	}
+	require.Equal(t, &remoteexecution.BatchReadBlobsResponse{
+		Responses: []*remoteexecution.BatchReadBlobsResponse_Response{
+			{
+				Digest: digest1.GetPartialDigest(),
+				Data:   a,
+				Status: status.Convert(nil).Proto(),
```
EdSchouten: Nit: I think these `Status:` lines can be removed entirely, right? This evaluates to nil.

pkg/cas/content_addressable_storage_server_test.go
```diff
@@ -0,0 +1,88 @@
...
+			},
+			{
+				Digest: digest2.GetPartialDigest(),
+				Data:   b,
+				Status: status.Convert(nil).Proto(),
+			},
+		},
```
EdSchouten: Would it make sense to extend this test to include a third digest for which `ToByteSlice()` returns `NotFound`?

pkg/cas/content_addressable_storage_server_test.go
sdclarke: When I tried without the `Status:` lines the test failed and showed a diff where the expected output had no `status` field but the actual output had `status: <>`.

pkg/cas/content_addressable_storage_server_test.go
sdclarke: Just to clear up my confusion, did you mean where the data buffer is larger than the digest's `SizeBytes` field for example? This would cause `InvalidArgument` from `ToByteSlice`, given that I can't find any way for a `NotFound` to be returned I guess that's what you meant?

pkg/cas/content_addressable_storage_server_test.go
EdSchouten: Ah, interesting. I incorrectly thought that `status.Convert(nil).Proto()` would evaluate to `nil`. I think that's even more of a reason that we shouldn't use this construct, as it makes it harder to understand what's going on.

Now the question becomes: what do we want to return in this case? I suspect that clients will generally support both. Returning `nil` results in a smaller serialized representation. But an empty Status object (the existing practice) is also acceptable.

If we do want to return an empty Status object, I guess we could improve readability by using this:

```go
import (
    status_pb "google.golang.org/genproto/googleapis/rpc/status"
)

    {
        Status: &status_pb.Status{},
    }
```

pkg/cas/content_addressable_storage_server_test.go
EdSchouten: I'm talking about the case where someone calls BatchReadBlobs(), but passes in one or more digests that are not present in the backing store. Or alternatively: Calling BatchReadBlobs(), even though the backing store is unavailable. That should let the BatchReadBlobs() call succeed, but should cause a non-OK Status to be embedded into the results. You can simulate this by creating a buffer like this:
```go
buf := buffer.NewBufferFromError(codes.NotFound, "The object you requested could not be found")
```

pkg/cas/content_addressable_storage_server_test.go
```diff
@@ -0,0 +1,117 @@
...
+func TestContentAddressableStorageServerBatchReadBlobsFailure(t *testing.T) {
+	ctrl, ctx := gomock.WithContext(context.Background(), t)
+	defer ctrl.Finish()
+
+	digest1 := digest.MustNewDigest("ubuntu1804", "0000000000000000000000000000000000000000000000000000000000000001", 123)
+	digest2 := digest.MustNewDigest("ubuntu1804", "0000000000000000000000000000000000000000000000000000000000000002", 234)
+
+	request := &remoteexecution.BatchReadBlobsRequest{
+		Digests: []*remoteexecution.Digest{
+			digest1.GetPartialDigest(),
```
EdSchouten: Could you eliminate the use of GetPartialDigest() here as well?

sdclarke: Ah I knew there would be something I'd missed, I forgot to change the hashes of the digests in the second test too.

EdSchouten: Yeah, the hashes could stay like that if we wanted to. It's not unlikely that some 123 byte blob exists that has that checksum. We don't instantiate any buffers as part of this test, so we're fine.

PR #51, opened 2020-03-18
CloudBlobAccess: Use Buffer API for Put

### LINKED ISSUE #50: CloudBlobAccess can cause OOM with big artifacts
When writing to a CloudBlobAccess, the entire blob gets loaded into memory thanks to a `buffer.ToReader` call. In systems without a large amount of memory this can cause an OOM, especially in workloads which push a lot of large blobs (e.g. docker images).

Othko97:
Fixes a bug where CloudBlobAccess would cause OOM when pushing large artifacts, thanks to a `buffer.ToReader` call pulling the entire blob into memory.

Uses a ChunkReader instead, and reads/writes in chunks. I wanted to avoid setting a flag to check if EOF was reached, but I couldn't find a way to (probably due to this being my first go patch).

Closes #50

### DIFF
--- a/pkg/blobstore/cloud_blob_access.go
+++ b/pkg/blobstore/cloud_blob_access.go
@@ -51,6 +51,7 @@ func (ba *cloudBlobAccess) Put(ctx context.Context, digest digest.Digest, b buff
 	w, err := ba.bucket.NewWriter(ctx, ba.getKey(digest), nil)
 	if err != nil {
 		cancel()
+		b.Discard()
 		return err
 	}
 	// In case of an error (e.g. network failure), we cancel before closing to
--- a/pkg/blobstore/cloud_blob_access.go
+++ b/pkg/blobstore/cloud_blob_access.go
@@ -2,7 +2,6 @@ package blobstore
 
 import (
 	"context"
-	"io"
 
 	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
 	"github.com/buildbarn/bb-storage/pkg/digest"
@@ -48,9 +47,6 @@ func (ba *cloudBlobAccess) Get(ctx context.Context, digest digest.Digest) buffer
 }
 
 func (ba *cloudBlobAccess) Put(ctx context.Context, digest digest.Digest, b buffer.Buffer) error {
-	r := b.ToReader()
-	defer r.Close()
-
 	ctx, cancel := context.WithCancel(ctx)
 	w, err := ba.bucket.NewWriter(ctx, ba.getKey(digest), nil)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


EdSchouten:
> Fixes a bug where CloudBlobAccess would cause OOM when pushing large artifacts, thanks to a `buffer.ToReader` call pulling the entire blob into memory.

`Buffer.ToByteSlice()` will attempt to load everything into memory. `Buffer.ToReader()` shouldn't. If it does, that's a bug on its own.

pkg/blobstore/cloud_blob_access.go
```diff
@@ -59,14 +61,25 @@ func (ba *cloudBlobAccess) Put(ctx context.Context, digest digest.Digest, b buff
 	}
 	// In case of an error (e.g. network failure), we cancel before closing to
 	// request the write to be aborted.
```
EdSchouten: Please take some time to read this comment. There is a reason why the code is the way it is: the order in which `cancel()` and `w.Close()` is called matters here. This instructs go-cloud to cancel the transmission of the current file.

pkg/blobstore/cloud_blob_access.go
```diff
@@ -59,14 +61,25 @@ func (ba *cloudBlobAccess) Put(ctx context.Context, digest digest.Digest, b buff
...
-	if _, err = io.Copy(w, r); err != nil {
```
EdSchouten: It looks like this code predates the addition of `Buffer.IntoWriter()`. Maybe it would be best to replace this line by:

```go
if _, err := b.IntoWriter(w); err != nil {
```

That way this backend becomes agnostic of the way data is passed in. If it turns out this still causes OOM issues, then it must either be something wrong in pkg/blobstore/buffer, go-cloud, the AWS SDK, etc.

pkg/blobstore/cloud_blob_access.go
Othko97: Ah, I hadn't noticed that the order flips depending on whether the close was unexpected or not, but I really should have.

pkg/blobstore/cloud_blob_access.go
Othko97: I think this is the actual fix to this, I've made this change and am just testing that it does fix my issue. I guess this should teach me to read APIs more carefully!

pkg/blobstore/cloud_blob_access.go
```diff
@@ -59,7 +55,7 @@ func (ba *cloudBlobAccess) Put(ctx context.Context, digest digest.Digest, b buff
 	}
```
EdSchouten: Could you also call `b.Discard()` inside the error path above? Sorry; should have mentioned that earlier.

pkg/blobstore/cloud_blob_access.go
Othko97: This didn't fix the issue, but the fact it didn't leads me to believe that the issue is not in the bb-storage source code. In any case, I think this is still a beneficial change.

EdSchouten:
Thanks for making this cleanup!

PR #53, opened 2020-04-07
Correct lies in error message

asuffield:
The text of the error is inverted from the actual test.

Also change the error code to reflect the intent.

### DIFF
--- a/pkg/ac/action_cache_server.go
+++ b/pkg/ac/action_cache_server.go
@@ -42,7 +42,7 @@ func (s *actionCacheServer) UpdateActionResult(ctx context.Context, in *remoteex
 		return nil, err
 	}
 	if instance := digest.GetInstance(); !s.allowUpdatesForInstances[instance] {
-		return nil, status.Errorf(codes.Unimplemented, "This service can only be used to get action results for instance %#v", instance)
+		return nil, status.Errorf(codes.PermissionDenied, "This service does not accept action results for instance %#v", instance)
 	}
 	return in.ActionResult, s.blobAccess.Put(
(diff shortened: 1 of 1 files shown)


pkg/ac/action_cache_server.go
```diff
@@ -42,7 +42,7 @@ func (s *actionCacheServer) UpdateActionResult(ctx context.Context, in *remoteex
 		return nil, err
 	}
 	if instance := digest.GetInstance(); !s.allowUpdatesForInstances[instance] {
-		return nil, status.Errorf(codes.Unimplemented, "This service can only be used to get action results for instance %#v", instance)
+		return nil, status.Errorf(codes.PermissionDenied, "This service does not accept action results from instance %#v", instance)
```
EdSchouten, line 45: I don't see how the existing error message is a lie. The service can only be used to **get** action results for instance foo. It cannot be used to **put** action results for instance foo.

Still, using `PermissionDenied` here seems more sensible. I'm not a native English speaker. That said, shouldn't 'from' in the new error message be 'for'?

"This shop does not sell collars from dogs"
->
"This shop does not sell collars for dogs"

asuffield: This is turning into the subtleties of English, but anyway... I see what you were trying to say, it just doesn't naturally parse that way. You were aiming for:

"This service can only be used to (get action results) for (instance default)"

What it reads as right now is:

"This service can only be used (to get action results) (for instance default)"

Which has the opposite interpretation: this service can only be used for instance default, the one we are rejecting.

from -> for is reasonable here, changed

PR #54, opened 2020-04-07
Add support for grpc health check protocol

asuffield:
This is used with a probe like
https://github.com/grpc-ecosystem/grpc-health-probe or
https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/health_check.proto#envoy-api-msg-core-healthcheck-grpchealthcheck
to determine whether a service is both healthy and the correct
service. It is typically deployed as a load balancer health check, to
handle the scenario where:

  1. Our service is running on 1.2.3.4:8981
  2. Our service is rescheduled to run on 2.3.4.5:8981
  3. A different grpc server is started on 1.2.3.4:8981

Named service health checks allow the load balancer to notice that it
has the wrong backend, and not send it any rpcs.

In a future iteration, we should also redesign the API in this package
so that the various services can annnounce when they are ready to
serve traffic, for zero-downtime rollouts behind a load balancer. For
now, just serving a named health check at all is a step forwards.

### DIFF
--- a/pkg/grpc/grpc.go
+++ b/pkg/grpc/grpc.go
@@ -12,6 +12,8 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/health"
+	"google.golang.org/grpc/health/grpc_health_v1"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/reflection"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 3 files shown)


pkg/grpc/grpc.go
```diff
@@ -150,6 +152,13 @@ func NewGRPCServersFromConfigurationAndServe(configurations []*configuration.Ser
 		// Enable grpc reflection.
 		reflection.Register(s)
 
+		// Enable grpc health service
+		h := health.NewServer()
+		grpc_health_v1.RegisterHealthServer(s, h)
+		// TODO: construct an API for the caller to indicate when it is healthy and set this. Until then, we'll just track this function's extry and exit.
+		h.SetServingStatus(configuration.HealthCheckService, grpc_health_v1.HealthCheckResponse_SERVING)
+		defer h.SetServingStatus(configuration.HealthCheckService, grpc_health_v1.HealthCheckResponse_NOT_SERVING)
```
EdSchouten, line 160: What’s the purpose of this deferred statement? Don’t we already `log.Fatal()` in case gRPC has an error

asuffield: Just defensive, in case the other code ever changes.

EdSchouten: Please remove it for the time being then. This function isn't designed at all to allow continuation in case of failure. Adding `defer` statements like these may only cause us to incorrectly assume it is.

asuffield: Done

PR #58, opened 2020-04-24
Adds Mknod() to filesystem.Directory

sdclarke:
Allows Mknod to be called on linux, on darwin this will give a fixed error as this operation will not work.

This is needed for a patch to bb-remote-execution to allow installation of device nodes in the input root of the action being run.

### DIFF
--- a/pkg/filesystem/directory.go
+++ b/pkg/filesystem/directory.go
@@ -79,6 +79,8 @@ type Directory interface {
 	Lstat(name string) (FileInfo, error)
 	// Mkdir is the equivalent of os.Mkdir().
 	Mkdir(name string, perm os.FileMode) error
+	// Mknod is the equivalent of unix.Mknod().
+	Mknod(name string, perm os.FileMode, dev int) error
 	// ReadDir is the equivalent of ioutil.ReadDir().
 	ReadDir() ([]FileInfo, error)
 	// Readlink is the equivalent of os.Readlink().
--- a/pkg/filesystem/local_directory_linux.go
+++ b/pkg/filesystem/local_directory_linux.go
@@ -2,5 +2,31 @@
 
 package filesystem
 
+import (
+	"os"
+	"runtime"
+
+	"golang.org/x/sys/unix"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
 // deviceNumber is the equivalent of POSIX dev_t.
 type deviceNumber = uint64
+
+func (d *localDirectory) Mknod(name string, perm os.FileMode, dev int) error {
+	if err := validateFilename(name); err != nil {
+		return err
+	}
+	defer runtime.KeepAlive(d)
+
+	var unixPerm uint32
+	switch perm & os.ModeType {
+	case os.ModeDevice | os.ModeCharDevice:
+		unixPerm = uint32(unix.S_IFCHR | (perm & os.ModePerm))
+	default:
+		return status.Error(codes.InvalidArgument, "The provided file mode is not for a character device")
+	}
+
+	return unix.Mknodat(d.fd, name, unixPerm, dev)
+}
--- a/internal/mock/BUILD.bazel
+++ b/internal/mock/BUILD.bazel
@@ -169,5 +169,6 @@ go_library(
         "@go_googleapis//google/longrunning:longrunning_go_proto",
         "@org_golang_google_grpc//:go_default_library",
         "@org_golang_google_grpc//metadata:go_default_library",
+        "@org_golang_x_sys//unix:go_default_library",
     ],
 )
--- a/pkg/filesystem/local_directory_darwin.go
+++ b/pkg/filesystem/local_directory_darwin.go
@@ -2,5 +2,16 @@
 
 package filesystem
 
+import (
+	"os"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
 // deviceNumber is the equivalent of POSIX dev_t.
 type deviceNumber = int32
+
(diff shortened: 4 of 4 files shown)


pkg/filesystem/directory.go
```diff
@@ -79,6 +79,8 @@ type Directory interface {
 	Lstat(name string) (FileInfo, error)
 	// Mkdir is the equivalent of os.Mkdir().
 	Mkdir(name string, perm os.FileMode) error
+	// Mknod is the equivalent of unix.Mknod().
+	Mknod(name string, perm os.FileMode, dev int) error
```
EdSchouten, line 83: I think that Linux uses uint64 for the device number. This interface tries to be OS independent (because it can be implemented arbitrarily). Should we simply use uint64 here as well?

pkg/filesystem/device.go
```diff
@@ -0,0 +1,40 @@
...
+// when creating device nodes in the input root dev directory
+type Device struct {
+	Name string
+	Rdev deviceNumber
+	Mode os.FileMode
+}
+
+// CreateDev will take a directory and a list of devices and make nodes
+// of those devices in the /dev subdirectory of the given directory
+func CreateDev(inputRootDirectory Directory, devices []Device) error {
```
EdSchouten, line 21: I don’t think we should put code like this in this specific package. This is too strongly related to a certain use case. This should likely be somewhere in bb-remote-execution?

pkg/filesystem/directory.go
sdclarke: When doing this I tried to match the underlying Mknod call which takes the device number as an int. would it be better pass a uint64 and cast it to an int within the directory.Mknod function, or to just leave it as it is?

EdSchouten: Hmmm... I just did a `git grep` on the go-sys package:

```
$ git grep 'func Mknodat' unix
unix/syscall_freebsd.go:func Mknodat(fd int, path string, mode uint32, dev uint64) (err error) {
unix/zsyscall_aix_ppc.go:func Mknodat(dirfd int, path string, mode uint32, dev int) (err error) {
unix/zsyscall_aix_ppc64.go:func Mknodat(dirfd int, path string, mode uint32, dev int) (err error) {
unix/zsyscall_dragonfly_amd64.go:func Mknodat(fd int, path string, mode uint32, dev int) (err error) {
unix/zsyscall_linux.go:func Mknodat(dirfd int, path string, mode uint32, dev int) (err error) {
unix/zsyscall_netbsd_386.go:func Mknodat(dirfd int, path string, mode uint32, dev int) (err error) {
unix/zsyscall_netbsd_amd64.go:func Mknodat(dirfd int, path string, mode uint32, dev int) (err error) {
unix/zsyscall_netbsd_arm.go:func Mknodat(dirfd int, path string, mode uint32, dev int) (err error) {
unix/zsyscall_netbsd_arm64.go:func Mknodat(dirfd int, path string, mode uint32, dev int) (err error) {
unix/zsyscall_openbsd_386.go:func Mknodat(dirfd int, path string, mode uint32, dev int) (err error) {
unix/zsyscall_openbsd_amd64.go:func Mknodat(dirfd int, path string, mode uint32, dev int) (err error) {
unix/zsyscall_openbsd_arm.go:func Mknodat(dirfd int, path string, mode uint32, dev int) (err error) {
unix/zsyscall_openbsd_arm64.go:func Mknodat(dirfd int, path string, mode uint32, dev int) (err error) {
unix/zsyscall_solaris_amd64.go:func Mknodat(dirfd int, path string, mode uint32, dev int) (err error) {
```
It looks like all OSes use `int`, except for FreeBSD. And FreeBSD we don't need to worry about, because `Mknod()` with a device number doesn't work in the first place. Devices can only be accessed through plain devfs. Let's use `int` then!

pkg/filesystem/file_info.go
```diff
@@ -15,6 +15,8 @@ const (
 	FileTypeDirectory
 	// FileTypeSymlink means the file is a symbolic link.
 	FileTypeSymlink
+	// FileTypeCharacterDevice means the file is a character device
```
EdSchouten, line 18: I suspect that this part of the change isn't really needed, right? I guess none of the code in Buildbarn ever needs to distinguish between FileTypeCharacterDevice and FileTypeOther.

sdclarke: In the patch which uses this functionality (https://github.com/buildbarn/bb-remote-execution/pull/52) I make use of this distinction to skip uploading character devices without failing, whilst still allowing the upload to fail if `FileTypeOther` appears. https://github.com/buildbarn/bb-remote-execution/pull/52/files#diff-48b3c6a7d58dce08bdf7f742b515e206

pkg/filesystem/local_directory_linux.go
```diff
@@ -2,5 +2,21 @@
...
 // deviceNumber is the equivalent of POSIX dev_t.
 type deviceNumber = uint64
+
+func (d *localDirectory) Mknod(name string, perm os.FileMode, dev int) error {
+	if err := validateFilename(name); err != nil {
+		return err
+	}
+	defer runtime.KeepAlive(d)
+
+	return unix.Mknodat(d.fd, name, uint32(perm), dev)
```
EdSchouten, line 21: So the idea here is to use `os.FileMode`, which I think is perfect! The only thing that concerns me is that it may not necessarily be valid to convert `perm` to a uint32 directly. According to [the documentation](https://golang.org/pkg/os/#FileMode), bits like `ModeCharDevice` have values that are identical across platforms. Its value isn't necessarily the same as `unix.S_IFCHR`.

Should we add logic to switch on `perm&os.ModeType` and validate that that this is equal to `os.ModeDevice|os.ModeCharDevice`?

PR #59, opened 2020-04-27
WIP: Add persistent LocalBlobAccess

tomcoldrick-ct:
Opening this to get preliminary feedback on my approach here, at least in terms of implementing a persistent LocationRecordArray.

Adds a persistent digest-location map, works pretty much the same way as
the in-memory one. As the "cuckoo" hashing approach of the keys mean
that we end up with an integral key, we may simply use this to define an
offset in a file, that is

    offset = key * serialised object width

At the moment, the serialisation/deserialisation is full of magic numbers, which I'd rather avoid, but I'm relatively new to Go and could not find some way to SizeOf() the underlying objects.

I've made this state directory configurable, and only be created if this
is set. I'm not particularly happy with this, and I'd like advice on a
better way to specify this. In particular, we shouldn't allow a persistent backend with non-persistent storage, as this will cause failures on restarts.

I still need to store the block mapping state and add logic to reload this, but first I wanted to make sure I'm not barking up the wrong tree with the digest-location map implementation.

Testing this (I've used bazel and BuildStream clients) has so far gone without a hitch.

Edit: I've now made a (probably rather ugly) storage for the block mapping state, and added logic to reload this. This has made the BlockDevice backend be stateful always, but I can't really think of a nice way to get this configurable - maybe explicitly specifying an empty string for the `stateDirectory` could work?

As I said, I'm not very experienced with Go, so I'd love feedback on how I can make this better!


tomcoldrick-ct:
I'm going to close this PR in favour of two which split the work into smaller chunks - one to implement the LocationRecordArray, and another to add state storage for the LocalBlobAccess itself.

The first such PR (for the LocationRecordArray) is #62.

PR #60, opened 2020-05-18
Add Utimes to directory interface

sdclarke:
Add UtimesNano to directory interface to allow setting the mtime of files in the directory.

### DIFF
--- a/pkg/cas/blob_access_content_addressable_storage.go
+++ b/pkg/cas/blob_access_content_addressable_storage.go
@@ -87,6 +87,11 @@ func (cas *blobAccessContentAddressableStorage) GetFile(ctx context.Context, dig
 		directory.Remove(name)
 		return err
 	}
+	time := filesystem.DeterministicFileModificationTimestamp
+	if err := directory.Chtimes(name, time, time); err != nil {
+		directory.Remove(name)
+		return err
+	}
 	return nil
 }
 
--- a/pkg/filesystem/directory.go
+++ b/pkg/filesystem/directory.go
@@ -3,6 +3,23 @@ package filesystem
 import (
 	"io"
 	"os"
+	"time"
+)
+
+var (
+	// DeterministicFileModificationTimestamp is a fixed timestamp that
+	// can be provided to Directory.Chtimes() to give files deterministic
+	// modification times. It is used by bb_worker to ensure that all
+	// files in the input root of a build action have the same
+	// modification time. This is needed to make certain kinds of build
+	// actions (most notably Autoconf scripts) succeed.
+	// 2000-01-01T00:00:00Z was chosen, because it's easy to distinguish
+	// from genuine timestamps. 1970-01-01T00:00:00Z would be impractical
+	// to use, because it tends to cause regressions in practice.
+	// Examples:
+	// https://bugs.python.org/issue34097
+	// https://gerrit.wikimedia.org/r/#/c/mediawiki/core/+/437977/
+	DeterministicFileModificationTimestamp = time.Date(2000, time.January, 1, 0, 0, 0, 0, time.UTC)
 )
 
 // CreationMode specifies whether and how Directory.Open*() should
@@ -92,6 +109,8 @@ type Directory interface {
 	RemoveAllChildren() error
 	// Symlink is the equivalent of os.Symlink().
 	Symlink(oldName string, newName string) error
+	// Chtimes sets the atime and mtime of the named file.
+	Chtimes(name string, atime, mtime time.Time) error
 
 	// Function that base types may use to implement calls that
 	// require double dispatching, such as hardlinking and renaming.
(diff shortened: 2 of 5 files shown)


pkg/filesystem/directory.go
```diff
@@ -93,6 +95,8 @@ type Directory interface {
 	// Symlink is the equivalent of os.Symlink().
 	Symlink(oldName string, newName string) error
 
+	UtimesNano(name string, ts []unix.Timespec) error
```
EdSchouten, line 98: Using the `unix` package as part of this interface seems like an unwise idea. This API is supposed to be OS agnostic. It's also fine to have virtual implementations of it that are independent of any OS.

Maybe it makes more sense to provide a `Chtimes()` function, similar to [`os.Chtimes()`](https://golang.org/pkg/os/#Chtimes)?

pkg/cas/blob_access_content_addressable_storage.go
```diff
@@ -87,6 +88,10 @@ func (cas *blobAccessContentAddressableStorage) GetFile(ctx context.Context, dig
 		directory.Remove(name)
 		return err
 	}
+	time := time.Date(2011, time.November, 11, 11, 11, 11, 0, time.UTC)
+	if err := directory.Chtimes(name, time, time); err != nil {
+		return err
```
EdSchouten, line 93: Maybe we should call `directory.Remove()` here as well?

pkg/filesystem/local_directory.go
```diff
@@ -354,6 +355,24 @@ func (d *localDirectory) Symlink(oldName string, newName string) error {
...
+	var ts [2]unix.Timespec
+	var err error
+	if ts[0], err = unix.TimeToTimespec(atime); err != nil {
+		return err
+	}
+	if ts[1], err = unix.TimeToTimespec(mtime); err != nil {
+		return err
+	}
+
+	return unix.UtimesNanoAt(d.fd, name, ts[0:], unix.AT_SYMLINK_NOFOLLOW)
```
EdSchouten, line 373: Nit: `ts[:]` should also work.

pkg/filesystem/local_directory.go
```diff
@@ -354,6 +355,24 @@ func (d *localDirectory) Symlink(oldName string, newName string) error {
...
+func (d *localDirectory) Chtimes(name string, atime, mtime time.Time) error {
+	if err := validateFilename(name); err != nil {
+		return err
+	}
+	defer runtime.KeepAlive(d)
+
+	var ts [2]unix.Timespec
+	var err error
+	if ts[0], err = unix.TimeToTimespec(atime); err != nil {
+		return err
```
EdSchouten, line 367: Could you add `util.StatusWrapWithCode(err, codes.InvalidArgument, "Cannot convert access time")` here?

pkg/cas/blob_access_content_addressable_storage.go
```diff
@@ -87,6 +88,10 @@ func (cas *blobAccessContentAddressableStorage) GetFile(ctx context.Context, dig
 		directory.Remove(name)
 		return err
 	}
+	time := time.Date(2011, time.November, 11, 11, 11, 11, 0, time.UTC)
```
EdSchouten, line 91: I have some other code that I want to push out at some point in the future that also has to set file timestamps to a deterministic value. What are your thoughts on adding a constant to pkg/filesystem that we can use throughout the codebase?

```go
var (
    DeterministicFileModificationTimestamp = time.Date(2000, time.January, 1, 0, 0, 0, 0, time.UTC)
)
```

I can adjust my patchset to make use of that constant as well. Silly question: we already started using 2000-01-01 on our end. Would you mind using that as well?

sdclarke:
> Could you maybe add some unit testing coverage for this?
I've added a test to `LocalDirectory`, I'm not sure there's anything to test in `BlobAccessContentAddressableStorage` but maybe I was just going about it incorrectly.

pkg/filesystem/file.go
```diff
@@ -2,6 +2,13 @@ package filesystem
 
 import (
 	"io"
+	"time"
+)
+
+var (
+	// DeterministicFileModificationTimestamp is a fixed timestamp set to
```
EdSchouten, line 9: This comment is somewhat tautological. It describes what it is, as opposed to describing the intent. What are your thoughts on something like this?

> DeterministicFileModificationTimestamp is a fixed timestamp that can be provided to Directory.Chtimes() to give files deterministic modification times. It is used by bb_worker to ensure that all files in the input root of a build action have the same modification time. This is needed to make certain kinds of build actions (most notably Autoconf scripts) succeed.
>
> 2000-01-01T00:00:00Z was chosen, because it's easy to distinguish from genuine timestamps. 1970-01-01T00:00:00Z would be impractical to use, because it tends to cause regressions in practice. Examples:
>
> https://bugs.python.org/issue34097                                            
> https://gerrit.wikimedia.org/r/#/c/mediawiki/core/+/437977/ 

Could you also move this constant into directory.go? I guess it makes more sense to place it there, because Chtimes() is also declared there.

EdSchouten:
Thanks!

PR #62, opened 2020-06-04
Add a FileBasedLocationRecordArray 

tomcoldrick-ct:
Implements a LocationRecordArray that is stored in a file on disk, rather than in memory. This allows  some of the state for a LocalBlobAccess to persist between restarts, paving the way to a fully persistent LocalBlobAccess. It also likely allows for larger DigestLocationMaps, especially in memory-constrained systems.

This is a cleanup of the first half of #59.

### DIFF
--- a/pkg/blobstore/local/file_backed_location_record_array.go
+++ b/pkg/blobstore/local/file_backed_location_record_array.go
@@ -1,0 +1,63 @@
+package local
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+
+	"github.com/buildbarn/bb-storage/pkg/blockdevice"
+)
+
+// FileBackedLocationRecordSize is the size of a single LocationRecord in bytes.
+const (
+	FileBackedLocationRecordSize = sha256.Size + 4 + 8 + 8 + 8
+)
+
+type fileBackedLocationRecordArray struct {
+	recordFile blockdevice.ReadWriterAt
+}
+
+// NewFileBackedLocationRecordArray creates a persistent LocationRecordArray.
+// Works by using a file as an array-like structure, writing serialised
+// LocationRecords next to each other and calculating where to Read/Write using
+// the length of a serialised record multiplied by the index.
+func NewFileBackedLocationRecordArray(recordFile blockdevice.ReadWriterAt) LocationRecordArray {
+	return &fileBackedLocationRecordArray{
+		recordFile: recordFile,
+	}
+}
+
+func (lra *fileBackedLocationRecordArray) Get(index int) (LocationRecord, error) {
+	var record [FileBackedLocationRecordSize]byte
+	if _, err := lra.recordFile.ReadAt(record[:], int64(index)*FileBackedLocationRecordSize); err != nil {
+		return LocationRecord{}, err
+	}
+
+	// Deserialize the read record into a LocationRecord
+	l := LocationRecord{
+		Key: LocationRecordKey{
+			Attempt: binary.LittleEndian.Uint32(record[32:]),
+		},
+		Location: Location{
+			BlockID:     int(binary.LittleEndian.Uint64(record[36:])),
+			OffsetBytes: int64(binary.LittleEndian.Uint64(record[44:])),
+			SizeBytes:   int64(binary.LittleEndian.Uint64(record[52:])),
+		},
+	}
+	copy(l.Key.Digest[:], record[:])
+	return l, nil
+}
+
+func (lra *fileBackedLocationRecordArray) Put(index int, locationRecord LocationRecord) error {
+	// Serialise the LocationRecord ready to be written to disk
+	var record [FileBackedLocationRecordSize]byte
+	copy(record[:], locationRecord.Key.Digest[:])
+	binary.LittleEndian.PutUint32(record[32:], locationRecord.Key.Attempt)
+	binary.LittleEndian.PutUint64(record[36:], uint64(locationRecord.Location.BlockID))
+	binary.LittleEndian.PutUint64(record[44:], uint64(locationRecord.Location.OffsetBytes))
+	binary.LittleEndian.PutUint64(record[52:], uint64(locationRecord.Location.SizeBytes))
+
+	if _, err := lra.recordFile.WriteAt(record[:60], int64(index)*FileBackedLocationRecordSize); err != nil {
+		return err
+	}
+	return nil
+}
--- a/pkg/blobstore/local/hashing_digest_location_map.go
+++ b/pkg/blobstore/local/hashing_digest_location_map.go
@@ -132,7 +132,10 @@ func (dlm *hashingDigestLocationMap) Get(digest CompactDigest, validator *Locati
 	key := LocationRecordKey{Digest: digest}
 	for {
 		slot := dlm.getSlot(&key)
-		record := dlm.recordArray.Get(slot)
+		record, err := dlm.recordArray.Get(slot)
+		if err != nil {
+			return Location{}, err
+		}
 		if !validator.IsValid(record.Location) {
 			// Record points to a block that no longer
 			// exists. There is no need to continue
@@ -164,10 +167,15 @@ func (dlm *hashingDigestLocationMap) Put(digest CompactDigest, validator *Locati
 	}
 	for iteration := 1; iteration <= dlm.maximumPutAttempts; iteration++ {
 		slot := dlm.getSlot(&record.Key)
-		oldRecord := dlm.recordArray.Get(slot)
+		oldRecord, err := dlm.recordArray.Get(slot)
+		if err != nil {
+			return err
+		}
 		if !validator.IsValid(oldRecord.Location) {
 			// The existing record may be overwritten directly.
-			dlm.recordArray.Put(slot, record)
+			if err := dlm.recordArray.Put(slot, record); err != nil {
+				return err
+			}
 			dlm.putInserted.Observe(float64(iteration))
 			return nil
 		}
@@ -175,7 +183,9 @@ func (dlm *hashingDigestLocationMap) Put(digest CompactDigest, validator *Locati
 			// Only allow overwriting an entry if it points
 			// to a newer version of the same blob.
 			if oldRecord.Location.IsOlder(record.Location) {
-				dlm.recordArray.Put(slot, record)
+				if err := dlm.recordArray.Put(slot, record); err != nil {
+					return err
+				}
 				dlm.putUpdated.Observe(float64(iteration))
 				return nil
 			}
@@ -187,7 +197,9 @@ func (dlm *hashingDigestLocationMap) Put(digest CompactDigest, validator *Locati
 			// it does point to older data than the record
 			// we're trying to insert. Displace the old
 			// record.
-			dlm.recordArray.Put(slot, record)
+			if err := dlm.recordArray.Put(slot, record); err != nil {
+				return err
+			}
 			record = oldRecord
 		}
 		record.Key.Attempt++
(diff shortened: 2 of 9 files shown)


pkg/blobstore/local/file_backed_location_record_array.go
```diff
@@ -0,0 +1,88 @@
...
+
+	return LocationRecord{
+		Key:      key,
+		Location: location,
+	}
+}
+
+type fileBackedLocationRecordArray struct {
+	recordFile filesystem.FileReadWriter
+	lock       sync.Mutex
```
EdSchouten, line 50: I don't think you need any locking on this, right? Don't the levels above already provide a sufficient amount of locking?

pkg/blobstore/local/file_backed_location_record_array.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (lra *fileBackedLocationRecordArray) Get(index int) (LocationRecord, error) {
+	lra.lock.Lock()
+	defer lra.lock.Unlock()
+
+	offset := int64(index * 60)
+	record := make([]byte, 60)
+	_, err := lra.recordFile.ReadAt(record, offset)
+	if err != nil {
```
EdSchouten, line 71: I would write this as:
```go
var record [60]byte
if _, err := lra.recordFile.ReadAt(record[:], int64(index) * 60); err != nil {
        return LocationRecord{}, err
}
```
- You can use arrays here instead of slices.
- `int64(index * 60)` is more likely to suffer from integer overflows than `int64(index) * 60`. A plain `int` may be 32 bits in size.
- Use scoped variables where possible.

The same holds for the code in `Put()`.

pkg/blobstore/local/file_backed_location_record_array.go
```diff
@@ -0,0 +1,88 @@
+package local
+
+import (
+	"encoding/binary"
+	"sync"
+
+	"github.com/buildbarn/bb-storage/pkg/filesystem"
+)
+
```
EdSchouten, line 9: Maybe it's good to add a `const` for the number 60?
```go
const (
   FileBackedLocationRecordSize = len(LocationRecord.Key.Digest) + 4 + 8 + 8 + 8
)
```
It's good to make this public, because pkg/blobstore/configuration also depends on it.

pkg/blobstore/local/file_backed_location_record_array.go
```diff
@@ -0,0 +1,88 @@
...
+func serializeLocationRecord(locationRecord LocationRecord) []byte {
+	serialised := make([]byte, 60)
+	for i := 0; i < 32; i++ {
```
EdSchouten, line 12: This can also use `copy()`.

pkg/blobstore/local/file_backed_location_record_array.go
```diff
@@ -0,0 +1,88 @@
...
+	binary.LittleEndian.PutUint64(serialised[36:44], uint64(locationRecord.Location.BlockID))
+	binary.LittleEndian.PutUint64(serialised[44:52], uint64(locationRecord.Location.OffsetBytes))
+	binary.LittleEndian.PutUint64(serialised[52:60], uint64(locationRecord.Location.SizeBytes))
+
+	return serialised
+}
+
+func deserializeLocationRecord(record []byte) LocationRecord {
+	var keyDigest [32]byte
+	copy(keyDigest[0:32], record[0:32])
```
EdSchouten, line 25: Keep in mind that `copy()` copies `min(len(a), len(b))` bytes. It's therefore perfectly fine to just call:
```go
copy(keyDigest, record)
```
That said, I guess it would make a lot of sense to write all of this in a way that's a bit more dense.
```go
l := LocationRecord{
        Key: LocationRecordKey{
                Attempt: binary.LittleEndian.Uint32(record[32:]),
        },
        BlockID: int(binary.LittleEndian.Uint64(record[36:]),
        OffsetBytes: int64(binary.LittleEndian.Uint64(record[44:]))
        SizeBytes: int64(binary.LittleEndian.Uint64(record[52:]))
}
copy(l.Key.Digest, record)
```

You're only using these serializing/deserializing functions in one place. I guess readability wouldn't suffer if you simply inlined this into `Get()` and `Put()`.

pkg/blobstore/local/hashing_digest_location_map.go
```diff
@@ -165,18 +168,25 @@ func (dlm *hashingDigestLocationMap) Put(digest digest.Digest, validator *Locati
 	}
 	for iteration := 1; iteration <= dlm.maximumPutAttempts; iteration++ {
 		slot := dlm.getSlot(&record.Key)
-		oldRecord := dlm.recordArray.Get(slot)
+		oldRecord, err := dlm.recordArray.Get(slot)
+		if err != nil {
+			return err
```
EdSchouten, line 173: So far all of these code paths were written to increment Prometheus histograms. We don't do this for the error cases. I guess that's fine, because we already provide BlobAccess error metrics? No need to change anything; simply braindumping!

pkg/blobstore/configuration/create_blob_access.go
```diff
@@ -496,6 +516,30 @@ func createDigestLocationMap(config *pb.LocalBlobAccessConfiguration, options *b
 		options.storageTypeName)
 }
 
+func createFileBackedDigestLocationMap(config *pb.LocalBlobAccessConfiguration, digestLocationMapConfig *pb.LocalBlobAccessConfiguration_FileBackedDigestLocationMap_, options *blobAccessCreationOptions, filename string) (local.DigestLocationMap, error) {
+	digestMapDirectory, err := filesystem.NewLocalDirectory(digestLocationMapConfig.FileBackedDigestLocationMap.Directory)
+	if err != nil {
+		return nil, err
+	}
+	defer digestMapDirectory.Close()
+	digestMapFile, err := digestMapDirectory.OpenReadWrite(filename, filesystem.CreateReuse(0644))
```
EdSchouten, line 525: I don't think `CreateReuse()` is safe just yet. If data is stored in memory, but the digest-location map is stored on disk (crazy I agree), it would cause us to reuse old digest-location map entries. For now I would suggest to use `CreateExcl()` and unlink the file up front.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -459,6 +459,27 @@ message LocalBlobAccessConfiguration {
     // intended to grow capacity beyond the size of memory.
     BlockDevice block_device = 10;
   }
+
+  message InMemoryDigestLocationMap {}
+
+  message FileBackedDigestLocationMap {
+    // Directory to house the digest-location map's state files.
+    string directory = 1;
```
EdSchouten, line 467: I am going to ask something from you that is technically speaking wrong, but it will make both our lives a bit easier.

I have a change on my side that I will likely push within the next couple of weeks that will change LocalBlobAccess to no longer require multiple digest-location maps for multiple instance names. This means that if we had a file-backed digest-location map, we wouldn't need to use a full directory. A single filename would suffice.

Could you already go ahead and write your code to work like this? Just use the same logic for the CAS and the AC. It's wrong in case you use multiple instances right now, but that will get fixed by me.

pkg/blobstore/local/file_backed_location_record_array.go
```diff
@@ -0,0 +1,88 @@
...
+type fileBackedLocationRecordArray struct {
+	recordFile filesystem.FileReadWriter
+	lock       sync.Mutex
+}
+
+// NewFileBackedLocationRecordArray creates a persistent LocationRecordArray.
+// Works by using a file as an array-like structure, writing serialised
+// LocationRecords next to each other and calculating where to Read/Write using
+// the length of a serialised record multiplied by the index.
+func NewFileBackedLocationRecordArray(recordFile filesystem.FileReadWriter) LocationRecordArray {
```
EdSchouten, line 57: What's a bit annoying is that `blockdevice.BlockDevice` does not implement a `filesystem.FileReadWriter`. This is a shame, because FileBackedLocationRecordArray would work perfectly fine with a block device, while the current implementation currently doesn't.

What are your thoughts on renaming this to BlockDeviceBackedLocationRecordArray and letting it use `blockdevice.BlockDevice`? It's all right to refer to it as being 'file backed' in the configuration file. It's just that it could also be block device backed if we wanted to.

pkg/blobstore/configuration/create_blob_access.go
EdSchouten: Alternative: do use `CreateReuse()`, but call `Truncate(0)` before truncating it to the desired size.

tomcoldrick-ct: Would this not completely empty the map? I realise this is unsafe with the InMemory backend, but a big benefit of this work is that it's necessary for persistent storage with the BlockDevice backend. I guess when that's added we'll have to add another case statement to check which backend is used and only empty the map for InMemory?

I'm realising just now that within the scope of this PR, this is also unsafe with BlockDevice backed storage as the Blocks don't get loaded on restart, so I'll definitely fix this up here in any case.

pkg/blobstore/local/file_backed_location_record_array.go
tomcoldrick-ct: I was reasonably sure that there was already enough locking to be safe without adding one here, but I erred on the side of safety. I'll remove the locks here, if they're not needed.

pkg/blobstore/local/file_backed_location_record_array.go
tomcoldrick-ct: I don't particularly mind mandating the use of a block device for the backend, but doing this is a little inflexible. Is there any reason not to implement a `filesystem.FileReadWriter` for a block device (I guess it's a bit of an oxymoron)?

Alternatively, could we not use a less strict interface (I think a ReadWriterAt would be sufficient) and allow either?

pkg/blobstore/configuration/create_blob_access.go
EdSchouten: "Yes" to the above! \o/

pkg/blobstore/local/file_backed_location_record_array.go
EdSchouten: You can't Truncate() a block device. Yes, sorry. The interface is already called blockdevice.ReadWriterAt; not blockdevice.BlockDevice.

pkg/blobstore/local/file_backed_location_record_array.go
tomcoldrick-ct: For some reason using `LocationRecord.Key.Digest` gave errors, so I used the Sha256.Size directly.

pkg/blobstore/configuration/create_blob_access.go
tomcoldrick-ct: So, I've added this now, however I've also added support for using a block device. As we can't truncate a block device, we will hit this problem again. I'm not well versed enough to know of some way to empty the array if it's backed by a block device. Any ideas?

pkg/blobstore/local/file_backed_location_record_array.go
tomcoldrick-ct: I've updated this to use a `blockdevice.ReadWriterAt`, with it being configurable whether to use a file or a block device for the storage. If preferred, I can make this block-device only and automatically calculate the size of the map (or do this in the block device case). I also wasn't sure if it might be better to just detect whether the path points to a block device or not, which is another option.

pkg/blobstore/local/file_backed_location_record_array.go
tomcoldrick-ct: I've inlined the serializing/deserializing functions, which I think has probably improved readability actually. For the `copy` I still had to slice the arrays, but I removed the limits.

EdSchouten:
Commit https://github.com/buildbarn/bb-storage/commit/8c04dc77bd6ef9a70a48c18d027b0a3c55c56b9d added a slightly modified version of this code, called BlockDeviceBackedLocationRecordArray.

Thanks for your contribution!

PR #63, opened 2020-06-05
Store state for BlockDevice-backed LocalBlobAccess

tomcoldrick-ct:
Allows a LocalBlobAccess configured with a BlockDevice backend to store
the current state, allowing it to be reloaded after a (clean) restart.
The state is stored in a file with an initial integer, expressing the
oldest block ID at time of write, followed by records expressing a
block's ID, whether it's old/current/new and the offset it's placed at.

For example:
  5 ← Oldest block Id
  0,O,123
  0,C,456
  0,N,789 ← Offset in device
  ↑ ↑
  | Whether the block is Old/Current/New
  |
  ID with respect to other blocks of the same type, restarts at 0 for
  each of old/current/new

I'm not really sure on how better to configure/store this, so feedback on that would be very welcome :)

The second half of #59, the first of which is at #62.

### DIFF
--- a/pkg/blobstore/configuration/create_blob_access.go
+++ b/pkg/blobstore/configuration/create_blob_access.go
@@ -404,9 +404,11 @@ func createBlobAccess(configuration *pb.BlobAccessConfiguration, options *blobAc
 		var sectorSizeBytes int
 		var blockSectorCount int64
 		var blockAllocator local.BlockAllocator
+		var statePath string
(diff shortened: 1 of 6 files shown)


EdSchouten:
Commit https://github.com/buildbarn/bb-storage/commit/8c04dc77bd6ef9a70a48c18d027b0a3c55c56b9d added a new version of LocalBlobAccess that can persist its state across restarts.

PR #64, opened 2020-06-08
Add support for metadata in gRPC client configuration

gorilych:
This change allows to specify custom headers in grpc metadata in blobstore configuration, for example 

```
      blobstore: {
        contentAddressableStorage: {
          grpc: {
            address: "remote.cache.tld:443",
            metadata: [
              {
                header: "authorization",
                values: [ "Bearer tokenXXXXXXXXXXXX" ]
              },
...
```

This might be useful in case if builbarn is used as remote cache frontend which requires specific metadata like auth headers.

### DIFF
--- a/pkg/grpc/add_metadata_interceptor_test.go
+++ b/pkg/grpc/add_metadata_interceptor_test.go
@@ -1,0 +1,75 @@
+package grpc_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/buildbarn/bb-storage/internal/mock"
+	bb_grpc "github.com/buildbarn/bb-storage/pkg/grpc"
+	"github.com/golang/mock/gomock"
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+func TestAddMetadataUnaryClientInterceptor(t *testing.T) {
+	ctrl, ctx := gomock.WithContext(context.Background(), t)
+	defer ctrl.Finish()
+
+	interceptor := bb_grpc.NewAddMetadataUnaryClientInterceptor([]string{"header", "value"})
+	invoker := mock.NewMockUnaryInvoker(ctrl)
+	req := &empty.Empty{}
+	resp := &empty.Empty{}
+
+	t.Run("AddHeader", func(t *testing.T) {
+		// Outgoing request metadata should be extended
+		// with pair ("header", "value").
+		invoker.EXPECT().Call(gomock.Any(), "SomeMethod", req, resp, nil).DoAndReturn(
+			func(ctx context.Context, method string, req interface{}, resp interface{}, cc *grpc.ClientConn, opts ...grpc.CallOption) error {
+				md, ok := metadata.FromOutgoingContext(ctx)
+				require.True(t, ok)
+				require.Equal(
+					t,
+					metadata.New(map[string]string{
+						"header": "value",
+					}),
+					md)
+				return nil
+			})
+
+		require.NoError(t, interceptor(ctx, "SomeMethod", req, resp, nil, invoker.Call))
+	})
+
+}
+
+func TestAddMetadataStreamClientInterceptor(t *testing.T) {
+	ctrl, ctx := gomock.WithContext(context.Background(), t)
+	defer ctrl.Finish()
+
+	interceptor := bb_grpc.NewAddMetadataStreamClientInterceptor([]string{"header", "value"})
+	streamDesc := grpc.StreamDesc{StreamName: "SomeMethod"}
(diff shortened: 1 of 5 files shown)


vors:
This is awesome, I was looking for this kind of support :)

One question: with Bearer tokens, it's desirable to have some refresh mechanics without restarting the bb-worker / bb-storage. Can it be expressed somehow in the config? I'm struggling to come up with a good expressive description for it, without relaying on inline shell script...

gorilych:
> with Bearer tokens, it's desirable to have some refresh mechanics 

Metadata values are hard-coded into configuration file which is read on the start.

I am looking for such functionality myself, but this patch is not about auth, actually.

pkg/proto/configuration/grpc/grpc.proto
```diff
@@ -35,6 +35,14 @@ message ClientConfiguration {
   //
   // Header names must be lower case.
   repeated string forward_metadata = 4;
+
+  // Map of gRPC metadata headers to set in client connection
+  // Header names must be lower case.
+  message header_values {
```
mickael-carl, line 41: Iirc best practice for message names is CamelCase and not snake_case?

EdSchouten:
Question to both @vors and @gorilych: is this feature still necessary from your end once #65 lands? Does #65 supersede this PR?

Not that I object to this feature; I just want to make sure that we don't add features that don't end up getting used.

pkg/proto/configuration/grpc/grpc.proto
gorilych: @mickael-carl it's protobuf file. See https://developers.google.com/protocol-buffers/docs/style

gorilych:
> 
> 
> Question to both @vors and @gorilych: is this feature still necessary from your end once #65 lands? Does #65 supersede this PR for your intended use?
> 

This is still useful, it allows api keys to be used.

pkg/proto/configuration/grpc/grpc.proto
mickael-carl: `Use CamelCase (with an initial capital) for message names – for example, SongServerRequest. Use underscore_separated_names for field names (including oneof field and extension names) – for example, song_name. ` -> isn't that applicable for `header_values` ? (taken from [here](https://developers.google.com/protocol-buffers/docs/style#message-and-field-names))

EdSchouten: Exactly. In this case it's a message; not a field name.

pkg/proto/configuration/grpc/grpc.proto
```diff
@@ -35,6 +35,14 @@ message ClientConfiguration {
...
+    string header = 1;
+    repeated string values = 2;
+  }
+  repeated header_values metadata = 5;
```
EdSchouten, line 45: Maybe call this `add_metadata`?

pkg/proto/configuration/grpc/grpc.proto
```diff
@@ -35,6 +35,14 @@ message ClientConfiguration {
   //
   // Header names must be lower case.
   repeated string forward_metadata = 4;
+
+  // Map of gRPC metadata headers to set in client connection
```
EdSchouten, line 39: Nit: General remark (also applies to #65). Please add all sentences in comments with a full stop (a period). Thanks!

pkg/grpc/add_metadata_interceptor_test.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/buildbarn/bb-storage/internal/mock"
+	bb_grpc "github.com/buildbarn/bb-storage/pkg/grpc"
+	"github.com/golang/mock/gomock"
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
```
EdSchouten, line 12: Nit: Please add an empty line between imports from separate second level domains (`github.com` vs `golang.org`). This is more consistent with the rest of the codebase.

pkg/proto/configuration/grpc/grpc.proto
gorilych: This makes sense. Will do

pkg/proto/configuration/grpc/grpc.proto
gorilych: done

pkg/grpc/add_metadata_interceptor_test.go
gorilych: done

pkg/proto/configuration/grpc/grpc.proto
gorilych: @mickael-carl you were right, fixed it

vors:
I probably set with #65, but I do think that this is useful.

PR #65, opened 2020-06-09
Add support for google auth in gRPC client configuration

gorilych:
This change allows to use google credentials in grpc backend of blobstore configuration, for example

      blobstore: {
        contentAddressableStorage: {
          grpc: {
            address: "remote.cache.tld:443",
            oauth: [
              {
                jsonKey: importstr '/path/to/service-account-key.json',
                // or it can be:
                //     googleDefaultCredentials: {}
                scopes: [ "https://www.googleapis.com/auth/userinfo.email" ],
              },

This might be useful in case if builbarn is used as remote cache frontend which requires google auth (like RBE).


pkg/proto/configuration/grpc/grpc.proto
```diff
@@ -51,6 +55,30 @@ message ClientKeepaliveConfiguration {
...
+  // to obtain access token.
+  // Either google_default_credentials should be set to True
+  // or json_key or json_key_file should be provided.
+  string json_key = 2;
+
+  // File with service account private key to use
+  // to obtain access token.
+  // Either google_default_credentials should be set to True
+  // or json_key or json_key_file should be provided.
+  string json_key_file = 3;
```
EdSchouten, line 75: This option is superfluous, right? It's also possible to use `jsonKey: importstr 'filename'` from within Jsonnet.

pkg/proto/configuration/grpc/grpc.proto
```diff
@@ -51,6 +55,30 @@ message ClientKeepaliveConfiguration {
   bool permit_without_stream = 3;
 }
 
+message ClientOAuthConfiguration {
+  // Use default google credentials (false by default)
```
EdSchouten, line 59: I suspect it would make more sense to model `google_default_credentials` and `json_key` as a `oneof`:
``` proto
oneof credentials {
  google.protobuf.Empty google_default_credentials = 1;
  string json_key = 2; // Or maybe call this 'service_account' to match the terminology of the OAuth library?
}
```
This ensures that they are truly orthogonal.

pkg/proto/configuration/grpc/grpc.proto
```diff
@@ -35,6 +35,10 @@ message ClientConfiguration {
   //
   // Header names must be lower case.
   repeated string forward_metadata = 4;
+
+  // Token-based authentication with Google.
+  // See https://grpc.io/docs/guides/auth/
+  ClientOAuthConfiguration credentials = 6;
```
EdSchouten, line 41: I guess it's best to call this 'oauth' instead of 'credentials' to make it explicit that OAuth is the mechanism that's used here.

pkg/proto/configuration/grpc/grpc.proto
gorilych: @EdSchouten you're correct, but it will work for jsonnet only. What if you'll switch from jsonnet to pure json or yaml after some time?

gorilych: it's ok to remove this option for me, though :)

EdSchouten: Then this won't be the only piece of code that needs adjusting. Notice how the TLS client and server also use Jsonnet `importstr` to load key material.

pkg/proto/configuration/grpc/grpc.proto
gorilych: will add

pkg/proto/configuration/grpc/grpc.proto
gorilych: ok

mickael-carl: The comment above also mentions Google specifically though, as well as the underlying message, is it wise to call it OAuth then?

EdSchouten: Yeah. Something like `// OAuth token-based authentication.` ?

mickael-carl: That'd mean moving the Google stuff out of the underlying message though, but then this can also be changed to a oneof?

EdSchouten: Google authentication is also based on OAuth, right? So it's perfectly fine to keep that in ClientOAuthConfiguration.

pkg/proto/configuration/grpc/grpc.proto
EdSchouten: Also to answer your question in more detail: it's unlikely that we ever get rid of Jsonnet and not replace it with something that's on par with it. Still, assuming we would get rid of Jsonnet (ad absurdum), then I'd rather see that we work towards a universal solution for loading key material from disk. For example, by adding a:

```proto
message KeyMaterial {
  oneof source {
    string direct = 1;
    string from_disk = 2;
  }
}
```

pkg/proto/configuration/grpc/grpc.proto
mickael-carl: I was thinking about `google_default_credentials` which is now move to a oneof so yeah, this works 👍

pkg/proto/configuration/grpc/grpc.proto
gorilych: option removed

pkg/proto/configuration/grpc/grpc.proto
gorilych: added in the next commit

pkg/grpc/grpc.go
```diff
@@ -58,25 +60,43 @@ func NewGRPCClientFromConfiguration(configuration *configuration.ClientConfigura
 		dialOptions = append(dialOptions, grpc.WithInsecure())
 	}
 
+	if oauthConfig := config.Oauth; oauthConfig != nil {
+		var perRPC credentials.PerRPCCredentials
+		var err error
+		switch oauthConfig.Credentials.(type) {
+		case (*configuration.ClientOAuthConfiguration_GoogleDefaultCredentials):
```
EdSchouten, line 67: Nit: You can remove the parentheses here.

pkg/grpc/grpc.go
```diff
@@ -58,25 +60,43 @@ func NewGRPCClientFromConfiguration(configuration *configuration.ClientConfigura
 		dialOptions = append(dialOptions, grpc.WithInsecure())
 	}
 
+	if oauthConfig := config.Oauth; oauthConfig != nil {
+		var perRPC credentials.PerRPCCredentials
+		var err error
+		switch oauthConfig.Credentials.(type) {
```
EdSchouten, line 66: If you use `switch credentials := oauthConfig.Credentials.(type)` here, you can use `credentials.ServiceAccountKey` later on to access it without any type assertions.

pkg/proto/configuration/grpc/grpc.proto
```diff
@@ -35,6 +35,10 @@ message ClientConfiguration {
   //
   // Header names must be lower case.
   repeated string forward_metadata = 4;
+
+  // Token-based authentication with Google.
```
EdSchouten, line 39: This should still be updated to no longer mention Google explicitly, right? This can also be used outside of Google.

pkg/proto/configuration/grpc/grpc.proto
```diff
@@ -51,6 +55,23 @@ message ClientKeepaliveConfiguration {
   bool permit_without_stream = 3;
 }
 
+message ClientOAuthConfiguration {
+  oneof credentials {
+    // Use default google credentials (false by default)
```
EdSchouten, line 60: The 'false by default' part can be removed now, because that's enforced by the use of `oneof`.
Nit: Capitalize 'google'.

pkg/grpc/grpc.go
gorilych: sure :)

pkg/proto/configuration/grpc/grpc.proto
gorilych: I'm not sure about it, because both methods (`oauth.NewApplicationDefault` and `oauth.NewServiceAccountFromKey`) are Google-specific. Mb in future "google.golang.org/grpc/credentials/oauth" will be extended to support other OAuth providers.

EdSchouten: Ah, got it. Well, in that case let's just stick to this then.

gorilych: From the other side, https://grpc.io/docs/guides/auth/ already mentions this fact, so I can remove it from comment.

gorilych: added in the next commit

pkg/proto/configuration/grpc/grpc.proto
gorilych: done

pkg/grpc/grpc.go
gorilych: done

PR #69, opened 2020-06-16
Adding option to dump empty base configuration model

gfleury:
Allow easy creation of a empty/clear configuration file structure. 
Understand the capabilities/features of a software sometimes can be done by its configuration files OR checking the code itself. This could be extended to the other cmd entry points for the other components of buildbarn. The idea is to ease the visibility of available features/options.

```
./bb_storage --dump_configuration_struct
{
  "blobstore": {
    "contentAddressableStorage": {

    },
    "actionCache": {

    }
  },
  "httpListenAddress": "",
  "grpcServers": [
  ],
  "schedulers": {
  },
  "allowAcUpdatesForInstances": [
  ],
  "maximumMessageSizeBytes": "0",
  "global": {
    "jaeger": {
      "agentEndpoint": "",
      "collectorEndpoint": "",
      "serviceName": "",
      "alwaysSample": false
    },
    "mutexProfileFraction": 0,
    "prometheusPushgateway": {
      "url": "",
      "job": "",
      "basicAuthentication": {
        "username": "",
        "password": ""
      },
      "grouping": {
      },
      "pushInterval": "0s"
    }
  }
}

```

### DIFF
--- a/cmd/bb_replicator/main.go
+++ b/cmd/bb_replicator/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"log"
 	"net/http"
 	"os"
@@ -20,9 +21,17 @@ import (
 
 func main() {
 	if len(os.Args) != 2 {
-		log.Fatal("Usage: bb_replicator bb_replicator.jsonnet")
+		log.Fatal("Usage: bb_replicator --dump_configuration_struct | bb_replicator.jsonnet")
 	}
 	var configuration bb_replicator.ApplicationConfiguration
+	if os.Args[1] == "--dump_configuration_struct" {
+		if confDump, err := util.MarshalExample(&configuration); err != nil {
+			log.Fatal("Failed to generate configuration example structure: %s", err)
+		} else {
+			fmt.Println(confDump)
+			return
+		}
+	}
 	if err := util.UnmarshalConfigurationFromFile(os.Args[1], &configuration); err != nil {
 		log.Fatalf("Failed to read configuration from %s: %s", os.Args[1], err)
 	}
--- a/cmd/bb_storage/main.go
+++ b/cmd/bb_storage/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"log"
 	"net/http"
 	"os"
@@ -24,9 +25,17 @@ import (
 
 func main() {
 	if len(os.Args) != 2 {
-		log.Fatal("Usage: bb_storage bb_storage.jsonnet")
+		log.Fatal("Usage: bb_storage --dump_configuration_struct | bb_storage.jsonnet")
 	}
 	var configuration bb_storage.ApplicationConfiguration
+	if os.Args[1] == "--dump_configuration_struct" {
+		if confDump, err := util.MarshalExample(&configuration); err != nil {
+			log.Fatal("Failed to generate configuration example structure: %s", err)
+		} else {
+			fmt.Println(confDump)
+			return
+		}
+	}
 	if err := util.UnmarshalConfigurationFromFile(os.Args[1], &configuration); err != nil {
 		log.Fatalf("Failed to read configuration from %s: %s", os.Args[1], err)
 	}
--- a/pkg/util/jsonnet.go
+++ b/pkg/util/jsonnet.go
@@ -3,6 +3,7 @@ package util
 import (
 	"io/ioutil"
 	"os"
+	"reflect"
 	"strings"
 
 	"github.com/golang/protobuf/jsonpb"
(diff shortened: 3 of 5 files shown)


pkg/util/jsonnet.go
```diff
@@ -49,3 +50,39 @@ func UnmarshalConfigurationFromFile(path string, configuration proto.Message) er
...
+	}
+
+	return jsonString, err
+}
+
+func initializeMessage(t reflect.Type, v reflect.Value) {
+	for i := 0; i < v.NumField(); i++ {
+		f := v.Field(i)
+		ft := t.Field(i)
+		if ft.Type.Kind() == reflect.Ptr {
```
EdSchouten, line 82: How does this work in case of recursively defined Protobuf messages? What about fields that are Protobuf implementation details? Should we use [protoreflect](https://godoc.org/google.golang.org/protobuf/reflect/protoreflect) instead?

cmd/bb_storage/main.go
```diff
@@ -24,9 +25,17 @@ import (
 
 func main() {
 	if len(os.Args) != 2 {
-		log.Fatal("Usage: bb_storage bb_storage.jsonnet")
+		log.Fatal("Usage: bb_storage [--dump_configuration_struct] bb_storage.jsonnet")
```
EdSchouten, line 28: Nit: UNIX usage/man page synopses always use `[]` to indicate optionality and `|` to indicate choice. The correct usage would therefore be:
```
Usage: bb_storage --dump_configuration_struct | filename
```

mickael-carl:
Does it make sense to add such feature directly to the binary, considering we have an [example deployments repository](https://github.com/buildbarn/bb-deployments)? Shouldn't we simply expand there as well as document what's unclear in the configuration?

gfleury:
Yeah, I have to agree that this is not something usual. Have seem such options in other softwares, like postfix. What lead me to come up with this was that some option for jaeager tracing options. Which were not very well documented. So, documentation will definitely solve this issue.

mickael-carl:
The proto should at least always be documented, pretty sure @EdSchouten guarantees that during reviews. For bb-deployments, we can strive to update faster, but iirc one of the deployments is always up-to-date pretty fast (docker-compose)?

cmd/bb_replicator/main.go
```diff
@@ -20,9 +21,17 @@ import (
...
 	if len(os.Args) != 2 {
-		log.Fatal("Usage: bb_replicator bb_replicator.jsonnet")
+		log.Fatal("Usage: bb_replicator [--dump_configuration_struct] bb_replicator.jsonnet")
 	}
 	var configuration bb_replicator.ApplicationConfiguration
+	if os.Args[1] == "--dump_configuration_struct" {
+		if confDump, err := util.MarshalExample(&configuration); err != nil {
+			log.Fatal("Failed to generate configuration example structure: %s", err)
+		} else {
+			fmt.Println(confDump)
```
mickael-carl, line 31: What if instead you'd output the conf in the specified file?

pkg/util/jsonnet.go
```diff
@@ -49,3 +50,39 @@ func UnmarshalConfigurationFromFile(path string, configuration proto.Message) er
 	}
 	return nil
 }
+
+// MarshalExample
+// marshals the  Protobuf message into a json string
```
mickael-carl, line 55:
```suggestion
// marshals the Protobuf message into a json string
```
Nit: double space

pkg/util/jsonnet.go
gfleury: 1- Good question, do we have an example of such msg?
2,3 - I don't know protoreflect but I can definitely check.

pkg/util/jsonnet.go
```diff
@@ -49,3 +50,39 @@ func UnmarshalConfigurationFromFile(path string, configuration proto.Message) er
...
+func MarshalExample(msg interface{}) (string, error) {
+
```
mickael-carl, line 57:
```suggestion
```
Nit: no need for this extra newline?

cmd/bb_storage/main.go
gfleury: Trueee, I was just checking your historical unix backgroud. You passed. =) good catch.

pkg/util/jsonnet.go
EdSchouten: BlobAccessConfiguration is a good example of that. Though right now that won't trigger, because there's a `oneof` inside of it that breaks the loop.

gfleury:
1) The idea here is definitely not replace documentation in any level. Is just to have a handy full options list to better understand of the features/capabilities.
2) .proto files are very well documented. 

As I mentioned on the previous comment, for me was just a easy way of getting a list of the available options. I did went thru the protofiles,  pointed on the README.
Totally agree with all the listed concerns, this just felt super handy for my first contact with the buildbarn components.

cmd/bb_replicator/main.go
gfleury: Doesn't look much safe, as @EdSchouten pointed out, most of the nested structures are empty.
He could pipe it to a file if wanted to create a base file to work on it.

EdSchouten:
I think if we wanted to merge this, we would at least want this to display `oneof`s properly. Right now the most interesting configuration bits live underneath `blobstore`, and this change isn't making it possible to display any of that. Displaying `oneof`s requires a fundamentally different approach, as we can't use (proto)reflect to construct a message and pass that on to jsonpb.

My hope is that someone comes along and develops a system similar to Swagger that is capable of rendering arbitrary Protobuf schemas in a fancy web UI and cross reference everything.

gfleury:
Makes totally sense. Today I was looking for such options over blobstores. I will take a look if this is possible with such approach, if so I would reopen this PR.

PR #70, opened 2020-06-24
Support instanceNames with "/" in the names

vors:
## Why

There are REAPI implementations that use `/` as a normal character in the instance names. This change should (at least partially) fix the problems with mixing them with bb.

## Test plan

Added a unit test + alternated existing tests to exercise this codepath a little.

### DIFF
--- a/pkg/blobstore/grpcservers/byte_stream_server.go
+++ b/pkg/blobstore/grpcservers/byte_stream_server.go
@@ -21,17 +21,26 @@ import (
 // - ${instance}/uploads/${uuid}/blobs/${hash}/${size}
 //
 // In the process, the hash, size and instance are extracted.
+// ${instance} could contain multiply segments.
 func parseResourceNameWrite(resourceName string) (digest.Digest, error) {
 	fields := strings.FieldsFunc(resourceName, func(r rune) bool { return r == '/' })
 	l := len(fields)
-	if (l < 5) || fields[l-5] != "uploads" || fields[l-3] != "blobs" {
+	if l < 5 || fields[l-5] != "uploads" || fields[l-3] != "blobs" {
 		return digest.BadDigest, status.Errorf(codes.InvalidArgument, "Invalid resource naming scheme")
 	}
 	size, err := strconv.ParseInt(fields[l-1], 10, 64)
 	if err != nil {
 		return digest.BadDigest, status.Errorf(codes.InvalidArgument, "Invalid resource naming scheme")
 	}
-	instance := strings.Join(fields[0:(l-5)], "/")
+	suffixLen := 0
+	for _, i := range []int{1, 2, 3, 4, 5} {
+		// 1 is for '/'
+		suffixLen = suffixLen + 1 + len(fields[l-i])
+	}
+	instance := ""
+	if suffixLen < len(resourceName) {
+		instance = resourceName[0 : len(resourceName)-suffixLen]
+	}
 	return digest.NewDigest(instance, fields[l-2], size)
 }
 
--- a/pkg/blobstore/grpcservers/byte_stream_server.go
+++ b/pkg/blobstore/grpcservers/byte_stream_server.go
@@ -33,7 +33,7 @@ func parseResourceNameWrite(resourceName string) (digest.Digest, error) {
 		return digest.BadDigest, status.Errorf(codes.InvalidArgument, "Invalid resource naming scheme")
 	}
 	suffixLen := 0
-	for _, i := range []int{1, 2, 3, 4, 5} {
+	for i := 1; i <= 5; i++ {
 		// 1 is for '/'
 		suffixLen = suffixLen + 1 + len(fields[l-i])
 	}
--- a/pkg/blobstore/grpcservers/byte_stream_server.go
+++ b/pkg/blobstore/grpcservers/byte_stream_server.go
@@ -24,17 +24,14 @@ import (
 func parseResourceNameWrite(resourceName string) (digest.Digest, error) {
 	fields := strings.FieldsFunc(resourceName, func(r rune) bool { return r == '/' })
 	l := len(fields)
-	if (l != 5 && l != 6) || fields[l-5] != "uploads" || fields[l-3] != "blobs" {
+	if (l < 5) || fields[l-5] != "uploads" || fields[l-3] != "blobs" {
 		return digest.BadDigest, status.Errorf(codes.InvalidArgument, "Invalid resource naming scheme")
 	}
 	size, err := strconv.ParseInt(fields[l-1], 10, 64)
 	if err != nil {
 		return digest.BadDigest, status.Errorf(codes.InvalidArgument, "Invalid resource naming scheme")
 	}
-	instance := ""
-	if l == 6 {
-		instance = fields[0]
-	}
+	instance := strings.Join(fields[0:(l-5)], "/")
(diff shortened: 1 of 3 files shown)


EdSchouten:
Interesting. I wasn't aware that people used instance names like these. This likely also breaks tools like bb-browser.

Do we care that this doesn't preserve things like leading/trailing slashes, and multiple slashes in a row?

vors:
Oh, good point. If we fix it we should fix it correctly and not half-way.

This came out in the context of Google's RBE, which uses the following schema:
`projects/[PROJECT_ID]/instances/[INSTANCE_ID]`

vors:
FWIW the spec is somewhat unclear about trailing and double `/`, but there is no harm in supporting it for sure.

https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L212

```
// A single server MAY support multiple instances of the execution system, each
// with their own workers, storage, cache, etc. The exact relationship between
// instances is up to the server. If the server does, then the `instance_name`
// is an identifier, possibly containing multiple path segments, used to
// distinguish between the various instances on the server, in a manner defined
// by the server. For servers which do not support multiple instances, then the
// `instance_name` is the empty path and the leading slash is omitted, so that
// the `resource_name` becomes `uploads/{uuid}/blobs/{hash}/{size}`.
// To simplify parsing, a path segment cannot equal any of the following
// keywords: `blobs`, `uploads`, `actions`, `actionResults`, `operations` and
// `capabilities`.
```

EdSchouten:
Yeah, let's assume the instance name is a path then. That seems fair. Having instance names with leading, trailing our double slashes should be forbidden if that's the case.

There is already a TODO in `digest.NewDigest()` about us not validating the instance name. Could you maybe extend that to contain a check like this?
```go
if instance != strings.Join(strings.FieldsFunc(instance, func(r rune) bool { return r == '/' }), "/") {
   return status.Error(...)
}
```
That way we can be sure that we don't introduce any undesirable aliasing of digests.

pkg/cas/byte_stream_server.go
```diff
@@ -24,17 +24,14 @@ import (
 func parseResourceNameWrite(resourceName string) (digest.Digest, error) {
 	fields := strings.FieldsFunc(resourceName, func(r rune) bool { return r == '/' })
 	l := len(fields)
-	if (l != 5 && l != 6) || fields[l-5] != "uploads" || fields[l-3] != "blobs" {
+	if (l < 5) || fields[l-5] != "uploads" || fields[l-3] != "blobs" {
```
EdSchouten, line 27: Nit: `(l < 5)` -> `l < 5`

pkg/cas/byte_stream_server.go
```diff
@@ -24,17 +24,14 @@ import (
...
 	}
 	size, err := strconv.ParseInt(fields[l-1], 10, 64)
 	if err != nil {
 		return digest.BadDigest, status.Errorf(codes.InvalidArgument, "Invalid resource naming scheme")
 	}
-	instance := ""
-	if l == 6 {
-		instance = fields[0]
-	}
+	instance := strings.Join(fields[0:(l-5)], "/")
```
EdSchouten, line 34: Nit: `fields[0:(l-5)]` -> `fields[:l-5]`. Same in digest.go.

vors:
> That way we can be sure that we don't introduce any undesirable aliasing of digests.

Oh that's a good call.
I think having string split / join on the `newDigest()` codepath is not ideal, seems like unnecessary overhead. Also it makes `digest.go` impl kind of dependent on the implementation in the `byte_stream_server.go` (because the assumption about the instance name is also in `parseResourceNameWrite`.

I think it makes sense to just re-write the parsing logic to avoid using the `strings.Join`.

EdSchouten:
Alternatively, just match for `HasPrefix("/") || HasSuffix("/") || Contains("//")`

pkg/blobstore/grpcservers/byte_stream_server_test.go
```diff
@@ -186,7 +186,7 @@ func TestByteStreamServer(t *testing.T) {
 		).Return(buffer.NewBufferFromError(status.Error(codes.NotFound, "Blob not found")))
 
 		req, err := client.Read(ctx, &bytestream.ReadRequest{
-			ResourceName: "///fedora28//blobs/09f34d28e9c8bb445ec996388968a9e8/////7/",
+			ResourceName: "fedora28/blobs/09f34d28e9c8bb445ec996388968a9e8/7",
```
vors, line 189: I'm not 100% sure what was the intention of the original test.... let me know if this should be reverted and made work as is.
As you can expect the error changed with the new parsing logic that allows `//` in the instance name.

pkg/blobstore/grpcservers/byte_stream_server_test.go
```diff
@@ -205,6 +205,35 @@ func TestByteStreamServer(t *testing.T) {
 		require.Equal(t, status.Error(codes.InvalidArgument, "Invalid resource naming scheme"), err)
 	})
 
+	t.Run("WriteSuccessTrickyInstance", func(t *testing.T) {
+		// Attempt to write a blob with an instance name.
+		blobAccess.EXPECT().Put(
+			gomock.Any(),
+			digest.MustNewDigest("/instance///default//", "581c1053f832a1c719fb6528a588ccfd", 14),
```
vors, line 212: Digest now supports `/` in any places of the name

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -21,19 +21,25 @@ import (
...
-	if (l != 5 && l != 6) || fields[l-5] != "uploads" || fields[l-3] != "blobs" {
+	if l < 5 || fields[l-5] != "uploads" || fields[l-3] != "blobs" {
 		return digest.BadDigest, status.Errorf(codes.InvalidArgument, "Invalid resource naming scheme")
 	}
 	size, err := strconv.ParseInt(fields[l-1], 10, 64)
 	if err != nil {
 		return digest.BadDigest, status.Errorf(codes.InvalidArgument, "Invalid resource naming scheme")
 	}
+	suffixLen := 0
+	for _, i := range []int{1, 2, 3, 4, 5} {
```
vors, line 36: omg, what did I write 🤦  let me replace it by a for loop

EdSchouten:
I gave all of this a bit more thought, and I think that if we want to do this right, we first need to tackle some technical debt. Right now instance names are modelled as plain strings. I think we should make an opaque type for them:

- Not all strings are valid instance names according to REv2. They may not contain certain keywords, such as `blobs` and `uploads`, as that would make ByteStream URLs ambiguous.
- It would make the `pkg/digest` API a bit cleaner if Digest creation functions were methods of such a type.

I don't want to bother you with fixing that kind of stuff. You're interested in fixing the problem at hand; not dealing with any toil. This is why I went ahead and wrote a bigger change to do cleanups to this code. I will continue testing this change on my end the upcoming week and merge it to master afterwards.

Thanks for bringing this issue to my attention and your work so far!

vors:
Cool, that sounds reasonable. No worries about the throw-away work, it works fine in my fork and once it's properly fixed, I will switch to upstream.

EdSchouten:
Hey @vors! Be sure to test 1824023b6cb68793e1e03d36a2106780b49044a5 and let me know whether that works for you. Thanks!

Issue #71, opened 2020-07-22
Unit test TestLocalDirectoryEnterSymlink fails on RHEL7, ELOOP returned instead of ENOTDIR


ehkloaj:
The unit test [TestLocalDirectoryEnterSymlink](https://github.com/buildbarn/bb-storage/blob/master/pkg/filesystem/local_directory_test.go#L69-L75) fails on our Red Hat Enterprise Linux 7 system. 
The test checks if the error code ENOTDIR is returned but on our system ELOOP is returned instead.
https://github.com/buildbarn/bb-storage/blob/master/pkg/filesystem/local_directory_test.go#L73

When reading about the issue we have found that ELOOP should be returned when too many symbolic links were encountered in the resolving pathname or when O_NOFOLLOW was specified but pathname was a symbolic link (see [source](https://www.tutorialspoint.com/unix_system_calls/open.htm)).

When investigating the code we have found that [O_NOFOLLOW](https://www.gnu.org/software/findutils/manual/html_node/find_html/O_005fNOFOLLOW.html) is specified in the syscall and therefore we think that the correct behavior should be to return ELOOP not ENOTDIR since the test tries to open a symlink to the root folder ("/").
https://github.com/buildbarn/bb-storage/blob/master/pkg/filesystem/local_directory.go#L54

We are not sure if this is behavior is due to our specific environment or if it is to be seen as a bug in the test. For those of you who got this test to pass it would be really helpful if you could provide us with which OS you are running on.

EdSchouten:
> For those of you who got this test to pass it would be really helpful if you could provide us with which OS you are running on.

The test currently passes on Ubuntu 18.04, macOS 10.15, and on Linux-based GitHub Actions workers (container image `docker://l.gcr.io/google/bazel:2.1.0`)

ehkloaj:
Thank you for the fast response @EdSchouten.

Then it seems that it has something to do with our environment. I will keep looking.

ehkloaj:
According to the [Linux Manual](https://man7.org/linux/man-pages/man2/open.2.html) ELOOP is returned when _pathname was a symbolic link, and flags specified O_NOFOLLOW but not O_PATH_. Specifying O_PATH solves the problem for us and does not break any other test.
`fd, err := unix.Openat(d.fd, name, unix.O_DIRECTORY|unix.O_NOFOLLOW|unix.O_RDONLY|unix.O_PATH, 0)`

[...]

PR #72, opened 2020-07-22
Return ENOTDIR instead of ELOOP when working with older Linux versions

ehrtro:
As discussed in https://github.com/buildbarn/bb-storage/issues/71 we're encountering issues when running TestLocalDirectoryEnterSymlink on our RHEL7 system (3.10.0-1127.8.2.el7.x86_64 Linux Kernel). Specifying O_PATH as a flag for openat solves this issue.

### DIFF
--- a/pkg/filesystem/local_directory.go
+++ b/pkg/filesystem/local_directory.go
@@ -56,6 +56,9 @@ func (d *localDirectory) enter(name string) (*localDirectory, error) {
 		if runtime.GOOS == "freebsd" && err == syscall.EMLINK {
 			// FreeBSD erroneously returns EMLINK.
 			return nil, syscall.ENOTDIR
+		} else if runtime.GOOS == "linux" && err == syscall.ELOOP {
+			// Linux 3.10 returns ELOOP, while Linux 4.15 returns ENOTDIR. Prefer the latter.
+			return nil, syscall.ENOTDIR
 		}
 		return nil, err
 	}


pkg/filesystem/local_directory.go
```diff
@@ -51,7 +51,7 @@ func (d *localDirectory) enter(name string) (*localDirectory, error) {
 	}
 	defer runtime.KeepAlive(d)
 
-	fd, err := unix.Openat(d.fd, name, unix.O_DIRECTORY|unix.O_NOFOLLOW|unix.O_RDONLY, 0)
+	fd, err := unix.Openat(d.fd, name, unix.O_DIRECTORY|unix.O_NOFOLLOW|unix.O_PATH|unix.O_RDONLY, 0)
```
EdSchouten, line 54: `O_PATH` is specific to Linux, so we shouldn't use that here.

What are your thoughts on extending the if-statement below to also contain something like this?
```go
if runtime.GOOS == "linux" && err == syscall.ELOOP {
        // Linux 3.10 returns ELOOP, while Linux 4.15 returns ENOTDIR. Prefer the latter.
        return nil, syscall.ENOTDIR
}
```

ehrtro: > `O_PATH` is specific to Linux, so we shouldn't use that here.
> 
> What are your thoughts on extending the if-statement below to also contain something like this?
> 
> ```go
> if runtime.GOOS == "linux" && err == syscall.ELOOP {
>         // Linux 3.10 returns ELOOP, while Linux 4.15 returns ENOTDIR. Prefer the latter.
>         return nil, syscall.ENOTDIR
> }
> ```

That sounds like a suitable solution to this issue, we'll update the PR accordingly!

PR #75, opened 2020-07-28
Add stackdriver trace

andyleap:
Basic implementation of stackdriver tracing.
Should we extract common fields (like always sample) into a separate structure?
Should we shove this kind of thing under a tracing or debugging field with the jaeger stuff?
I punted on those for now cause it'd be a breaking config change, but more then happy to rework them

### DIFF
--- a/pkg/global/apply_configuration.go
+++ b/pkg/global/apply_configuration.go
@@ -26,19 +26,11 @@ import (
 func ApplyConfiguration(configuration *pb.Configuration) error {
 	// Push traces to Jaeger.
 	if tracingConfiguration := configuration.GetTracing(); tracingConfiguration != nil {
-		if jaegerConfiguration := tracingConfiguration.GetJaeger(); jaegerConfiguration != nil {
-			pe, err := prometheus_exporter.NewExporter(prometheus_exporter.Options{
-				Registry:  prometheus.DefaultRegisterer.(*prometheus.Registry),
-				Namespace: "bb_storage",
-			})
-			if err != nil {
-				return util.StatusWrap(err, "Failed to create the Prometheus stats exporter")
-			}
-			view.RegisterExporter(pe)
-			if err := view.Register(ocgrpc.DefaultServerViews...); err != nil {
-				return util.StatusWrap(err, "Failed to register ocgrpc server views")
-			}
-			zpages.Handle(nil, "/debug")
+		if err := view.Register(ocgrpc.DefaultServerViews...); err != nil {
+			return util.StatusWrap(err, "Failed to register ocgrpc server views")
+		}
+
+		if jaegerConfiguration := tracingConfiguration.Jaeger; jaegerConfiguration != nil {
 			je, err := jaeger.NewExporter(jaeger.Options{
 				AgentEndpoint:     jaegerConfiguration.AgentEndpoint,
 				CollectorEndpoint: jaegerConfiguration.CollectorEndpoint,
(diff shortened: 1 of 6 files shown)


pkg/global/apply_configuration.go
```diff
@@ -24,33 +25,48 @@ import (
...
-			Namespace: "bb_storage",
-		})
-		if err != nil {
-			return util.StatusWrap(err, "Failed to create the Prometheus stats exporter")
-		}
-		view.RegisterExporter(pe)
-		if err := view.Register(ocgrpc.DefaultServerViews...); err != nil {
-			return util.StatusWrap(err, "Failed to register ocgrpc server views")
+	if tracingConfiguration := configuration.GetTracing(); tracingConfiguration != nil {
+		if jaegerConfiguration := tracingConfiguration.GetJaeger(); jaegerConfiguration != nil {
```
EdSchouten, line 29: Nit: It's perfectly fine to just use `.Jaeger` instead of `.GetJaeger()` here, as `tracingConfiguration` is guaranteed to be non-nill.

pkg/proto/configuration/global/global.proto
```diff
@@ -15,9 +15,16 @@ message JaegerConfiguration {
 
   // OpenTracing service name.
   string service_name = 3;
+}
 
-  // Whether or not all traces should be sampled.
-  bool always_sample = 4;
+message StackdriverConfiguration {
+  // GCP project to send trace data to
```
EdSchouten, line 21: Maybe it's a bit more elegant to not end this sentence with a preposition? Also make sure that all sentences are ended with a period.

> GCP project to which trace data is sent. Can be left blank when running on GCP, as it is automatically determined by the metadata server.

pkg/global/apply_configuration.go
```diff
@@ -24,33 +25,48 @@ import (
...
+				Namespace: "bb_storage",
+			})
+			if err != nil {
+				return util.StatusWrap(err, "Failed to create the Prometheus stats exporter")
+			}
+			view.RegisterExporter(pe)
+			if err := view.Register(ocgrpc.DefaultServerViews...); err != nil {
+				return util.StatusWrap(err, "Failed to register ocgrpc server views")
+			}
+			zpages.Handle(nil, "/debug")
```
EdSchouten, line 41: Is this code specific to Jaeger, by the way? What does it even do? I'm pretty much an idiot when it comes to OpenTracing.

pkg/proto/configuration/global/global.proto
```diff
@@ -46,10 +53,20 @@ message PrometheusPushgatewayConfiguration {
   google.protobuf.Duration push_interval = 5;
 }
 
-message Configuration {
+message TracingConfiguration {
```
EdSchouten, line 56: Should this be called OpenCensusTracing, or would that be too explicit?

andyleap: Yeah, I think that's a bit specific to be getting at this point, I'd say people would probably want to reevaluate their tracing if we start supporting multiple tracing frameworks, which seems a bit... excessive anyhow.

pkg/proto/configuration/global/global.proto
andyleap: Done

pkg/global/apply_configuration.go
andyleap: It's not actually, broke this out into a few different settings.  Basically, hooking up an exporter from opencensus's span stats and such to prometheus, registering stuff from grpc as needing to be tracked (it's... complex?), and hooking up a debug web page.

pkg/global/apply_configuration.go
```diff
@@ -24,33 +25,55 @@ import (
...
+				Registry:  prometheus.DefaultRegisterer.(*prometheus.Registry),
+				Namespace: "bb_storage",
+			})
+			if err != nil {
+				return util.StatusWrap(err, "Failed to create the Prometheus stats exporter")
+			}
+			view.RegisterExporter(pe)
+		}
+
+		if tracingConfiguration.EnableZpages {
```
EdSchouten, line 69: I think this functionality is even broken at this point in time, as we never expose all of /debug of the default mux. Shall we just kick this out of the tree for the time being?

andyleap: Yeah, I was just trying to not disrupt too much, but if it's not being served up, no point in preserving it...

andyleap: Done

PR #79, opened 2020-08-10
Add replication support to ReadFallback

asuffield:
Package move is necessary because //pkg/blobstore cannot depend on
//pkg/blobstore/replication

Otherwise this is just cribbed from Mirrored. Note that backwards
compatibility is maintained in the case where replicator is nil, so
existing configs keep doing the same thing (and not everybody wants
replication here, it's not always the right thing to do).

Signed-off-by: Andrew Suffield <asuffield@gmail.com>


pkg/blobstore/readfallback/read_fallback_blob_access.go~
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
```
EdSchouten, line 4: Did you unintentionally copy this file?

pkg/blobstore/readfallback/read_fallback_blob_access_test.go
```diff
@@ -0,0 +1,250 @@
+package readfallback_test
```
EdSchouten, line 1: This file also got duplicated unnecessarily?

pkg/blobstore/readfallback/read_fallback_blob_access.go~
```diff
@@ -77,13 +81,21 @@ func (eh *readFallbackErrorHandler) OnError(observedErr error) (buffer.Buffer, e
...
 	}
 	if eh.secondary == nil {
-		// Both backends returned NOT_FOUND. Don't bother
-		// prepending the name of the backend.
+		// No secondary exists, so just return the error
 		return nil, observedErr
 	}
-	ba := eh.secondary
+
+	if eh.replicator != nil {
```
EdSchouten, line 88: The Buildbarn codebase tends to stick to the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern) as much as possible.

Please add a NoopBlobReplicator, for which ReplicateMultiple() does nothing, and ReplicateSingle() does nothing more than call Get() on the source. The advantage of doing that is that we then have a universal mechanism for this, not specific to ReadFallbackBlobAccess.

pkg/blobstore/readfallback/read_fallback_blob_access.go~
```diff
@@ -77,13 +81,21 @@ func (eh *readFallbackErrorHandler) OnError(observedErr error) (buffer.Buffer, e
 		return nil, util.StatusWrap(observedErr, "Secondary")
 	}
 	if eh.secondary == nil {
-		// Both backends returned NOT_FOUND. Don't bother
-		// prepending the name of the backend.
+		// No secondary exists, so just return the error
```
EdSchouten, line 84: Please end sentences source code comments and documentation with a full stop.

pkg/blobstore/readfallback/read_fallback_blob_access.go~
asuffield: Bah this repo doesn't have .gitignore for emacs. Fixed.

pkg/blobstore/readfallback/read_fallback_blob_access.go~
asuffield: .

pkg/blobstore/readfallback/read_fallback_blob_access.go~
EdSchouten: Be sure to extend .gitignore if you want!

pkg/blobstore/readfallback/read_fallback_blob_access.go~
asuffield: Done

pkg/blobstore/configuration/new_blob_access.go
```diff
@@ -385,7 +387,17 @@ func NewNestedBlobAccess(configuration *pb.BlobAccessConfiguration, creator Blob
 		if err != nil {
 			return nil, err
 		}
-		implementation = blobstore.NewReadFallbackBlobAccess(primary, secondary)
+		var replicator replication.BlobReplicator
```
EdSchouten, line 390: Current style is not to rely on default configuration options too heavily. Just remove the if-condition below and call into NewBlobReplicatorFromConfiguration() unconditionally.

pkg/blobstore/readfallback/read_fallback_blob_access_test.go
```diff
@@ -20,7 +21,8 @@ func TestReadFallbackBlobAccessGet(t *testing.T) {
 
 	primary := mock.NewMockBlobAccess(ctrl)
 	secondary := mock.NewMockBlobAccess(ctrl)
-	blobAccess := blobstore.NewReadFallbackBlobAccess(primary, secondary)
+	replicator := replication.NewNoopBlobReplicator(secondary)
```
EdSchouten, line 24: Please use a `replicator := mock.NewBlobReplicator(ctrl)` here. NoopBlobReplicator is not the type we should be testing here.

Once you adjust all of the tests below, you can remove all of `TestReadFallbackBlobAccessReplication()`, as `TestReadFallbackBlobAccessGet()` will then provide sufficient coverage.

pkg/blobstore/replication/noop_blob_replicator.go
```diff
@@ -0,0 +1,31 @@
...
+type noopBlobReplicator struct {
+	source blobstore.BlobAccess
+}
+
+// NewNoopBlobReplicator creates a BlobReplicator that can be used to
+// access a single source without replication.
+//
+// It is useful for the BlobAccess variants where replication is optional.
+func NewNoopBlobReplicator(source blobstore.BlobAccess) BlobReplicator {
+	return &noopBlobReplicator{
```
EdSchouten, line 20: This type is sufficiently small that you don't need to return this by reference. Just do `return noopBlobReplicator{ ... }` and remove the pointer from the receiver types.

pkg/blobstore/configuration/new_blob_access.go
asuffield: Sure - I'll note that this does break existing configurations, I was just trying to make it backwards compatible.

pkg/blobstore/readfallback/read_fallback_blob_access_test.go
asuffield: Ah yes, artifact of the previous iteration.

pkg/blobstore/replication/noop_blob_replicator.go
asuffield: I don't think it makes any real difference, but done.

EdSchouten:
Thanks!

PR #81, opened 2020-08-13
Expose findMissingQueue for external usage

Qinusty:
This allows findMissing queue to be used outside of
CompletenessCheckingBlobAccess as it can be used for more general
use cases where large sets of digests need to be derived
and checked.

### DIFF
--- a/pkg/blobstore/completenesschecking/BUILD.bazel
+++ b/pkg/blobstore/completenesschecking/BUILD.bazel
@@ -2,7 +2,10 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 
 go_library(
     name = "go_default_library",
-    srcs = ["completeness_checking_blob_access.go"],
+    srcs = [
+        "completeness_checking_blob_access.go",
+        "find_missing_queue.go",
+    ],
     importpath = "github.com/buildbarn/bb-storage/pkg/blobstore/completenesschecking",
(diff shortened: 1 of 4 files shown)


EdSchouten:
What’s your intended use case? Have you considered using BatchedStoreBlobAccess from bb-remote-execution?

Qinusty:
> What’s your intended use case? Have you considered using BatchedStoreBlobAccess from bb-remote-execution?

Ensuring that all all digests returned by the asset service are available within the CAS with a CompletenessCheckingFetcher implementation. https://github.com/Qinusty/bb-asset-hub/issues/4

Qinusty:
> A custom worker could have done the downloading part. By following an approach like that, we would have gained:

> Insight in what's going on using the bb-scheduler page.
Insight in what happened in the past using bb-browser.
Global in-flight deduplication of fetch operations for the same URLs.
The ability to add concurrency and to accurately limit it.
The ability to place the Remote Asset API endpoint in a different network compared to where the actual fetching takes place.
The ability to use bb-storage to cache information about previous fetch operations.

We have discussed this quite a lot and there are quite a few reasons we chosen not to have this as the default behaviour. I've addressed this comment in further detail in the [relevant issue](https://github.com/Qinusty/bb-asset-hub/issues/18#issuecomment-676290153)

Qinusty:
Closing as this will likely be addressed via an alternative solution.

PR #82, opened 2020-08-13
Teach CloudBlobAccess to touch metadata during FindMissing calls

asuffield:
This makes cloud objects more LRU-like in their expiry behavior, and
hence more usable.

Initially implemented for S3, others will take some experimentation.


pkg/blobstore/cloud_blob_access.go
```diff
@@ -69,10 +72,18 @@ func (ba *cloudBlobAccess) Put(ctx context.Context, digest digest.Digest, b buff
 func (ba *cloudBlobAccess) FindMissing(ctx context.Context, digests digest.Set) (digest.Set, error) {
 	missing := digest.NewSetBuilder()
 	for _, blobDigest := range digests.Items() {
```
EdSchouten, line 74: What would really improve performance is if we could somehow use pipelining for these requests. Right now we have O(n) roundtrips.

pkg/blobstore/configuration/new_blob_access.go
```diff
@@ -121,7 +122,14 @@ func NewNestedBlobAccess(configuration *pb.BlobAccessConfiguration, creator Blob
 			if err != nil {
 				return nil, err
 			}
-			implementation = blobstore.NewCloudBlobAccess(bucket, backend.Cloud.KeyPrefix, storageType)
+			var refreshInterval time.Duration
+			if backendConfig.S3.RefreshInterval != nil {
```
EdSchouten, line 126: Trailing comment after I already pressed the submit button:

What are the costs (in terms of money) of sending COPY requests? If this is more expensive than plain HEAD requests, should we make this option required? I just want to make sure that existing users of CloudBlobAccess don't run into unintended expenses by upgrading to this new code.

Maybe this should also be documented.

pkg/cloud/aws/copy_mutator.go
```diff
@@ -0,0 +1,36 @@
...
+		refreshInterval: refreshInterval,
+	}
+}
+
+func (m blobCopyMutator) Mutate(asFunc func(interface{}) bool) error {
+	input := &s3.CopyObjectInput{}
+	if !asFunc(&input) {
+		return fmt.Errorf("failed to get CopyObjectInput - not an s3 bucket?")
+	}
+	input.CopySourceIfUnmodifiedSince = aws.Time(time.Now().Add(m.refreshInterval))
```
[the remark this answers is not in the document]
EdSchouten: Hmmm... Shouldn't this subtract? "Copy the file if it's unmodified since $two_weeks_ago".

pkg/cloud/aws/copy_mutator.go
```diff
@@ -0,0 +1,36 @@
...
+// NewBlobCopyMutator creates a CopyMutator which updates the "Used"
+// metadata field on s3 objects with the current time.
+func NewBlobCopyMutator(refreshInterval time.Duration) cloud.CopyMutator {
+	return blobCopyMutator{
+		refreshInterval: refreshInterval,
+	}
+}
+
+func (m blobCopyMutator) Mutate(asFunc func(interface{}) bool) error {
+	input := &s3.CopyObjectInput{}
```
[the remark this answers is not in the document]
asuffield: They actually say exactly the same thing, except my version allocates a useless extra object which is immediately GCed. Fixed accordingly.

pkg/blobstore/cloud_blob_access.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
+	"github.com/buildbarn/bb-storage/pkg/cloud"
```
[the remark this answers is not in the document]
asuffield: I borrow a line from somebody I knew at Google, which is that "cloud" expands to "I don't know".

There's probably no good name for this.

(Meanwhile I shall move pkg/cloud, that seems reasonable)

pkg/blobstore/configuration/new_blob_access.go
asuffield: S3 bills this as if it was a PUT request: https://aws.amazon.com/s3/pricing/

That is indeed more expensive, and making it optional seems reasonable. The cost is likely inconsequential.

pkg/cloud/copy_mutator.go
```diff
@@ -0,0 +1,9 @@
+package cloud
+
+// CopyMutator can be used from a
```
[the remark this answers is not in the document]
asuffield: Sure, a function type also works here.

pkg/cloud/aws/copy_mutator.go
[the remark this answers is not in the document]
asuffield: Done

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -293,6 +293,12 @@ message S3BlobAccessConfiguration {
 
   // AWS access options and credentials.
   buildbarn.configuration.cloud.aws.SessionConfiguration aws_session = 7;
+
+  // When the object becomes older than this interval, refresh it by
+  // updating the "Used" metadata field. This prevents objects from
+  // expiring. Default is to refresh every time the object is touched
+  // by FindMissing.
```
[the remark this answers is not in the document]
asuffield: It's marginal, but sure, duly adjusted.

Documentation updated, plus including a few things I've learned about configuring S3.

pkg/blobstore/cloud_blob_access.go
asuffield: Easy enough.

Writing a test for it was rather more effort, but also done.

pkg/blobstore/configuration/new_blob_access.go
asuffield: Documented in the proto comments, and now it's a required field because that seems reasonable.

The way to "disable" it is obvious after a little thought: just set this to any value larger than your expiration time. Or just any large value.

pkg/blobstore/cloud/cloud_blob_access.go
```diff
@@ -76,13 +80,58 @@ func (ba *cloudBlobAccess) Put(ctx context.Context, digest digest.Digest, b buff
 }
 
 func (ba *cloudBlobAccess) FindMissing(ctx context.Context, digests digest.Set) (digest.Set, error) {
+	// Workers send their errors here
```
EdSchouten, line 83: Does this actually use HTTP pipelining, or are we spawning O(n) goroutines to hammer S3?

I think a change like this needs more evaluation before we can follow it through. I guess it's wiser to just focus on the BeforeCopyFunc parts for now.

pkg/blobstore/cloud/cloud_blob_access.go
```diff
@@ -17,18 +19,20 @@ import (
...
-func NewCloudBlobAccess(bucket *blob.Bucket, keyPrefix string, readBufferFactory ReadBufferFactory, digestKeyFormat digest.KeyFormat) BlobAccess {
+func NewCloudBlobAccess(bucket *blob.Bucket, keyPrefix string, readBufferFactory blobstore.ReadBufferFactory, digestKeyFormat digest.KeyFormat, beforeCopy BeforeCopyFunc) blobstore.BlobAccess {
 	return &cloudBlobAccess{
 		bucket:            bucket,
 		keyPrefix:         keyPrefix,
 		readBufferFactory: readBufferFactory,
 		digestKeyFormat:   digestKeyFormat,
+		beforeCopy:        beforeCopy,
 	}
 }
```
EdSchouten, line 38: This solution only triggers copies when doing FindMissing(). Though this may be acceptable, it's important document somewhere that it only 'fixes' the CAS. Using this for the AC still causes actions to get lost, even when accessed. This is different from, say, LocalBlobAccess which touches objects both in Get() and FindMissing().

Out of interest, can we think of any ways to also fix this for Get()? Is there anything better we can do than sending two separate calls (one GetObject, one CopyObject)?

pkg/digest/set_builder.go
```diff
@@ -2,11 +2,15 @@ package digest
...
 	"sort"
+	"sync"
 )
 
 // SetBuilder is a builder for Set objects.
 type SetBuilder struct {
 	digests map[Digest]struct{}
+
+	// Protects access to digests
+	lock sync.Mutex
```
EdSchouten, line 13: This seems like a pessimisation for the majority of places where SetBuilder is used in synchronous cases.

pkg/blobstore/cloud/cloud_blob_access.go
asuffield: Very much the latter - but splitting it out is reasonable, I'll turn it into a separate PR.

(Note that we're not transferring any data, and the transport layer is delegated to the aws sdk, but I would expect the transport layer to do the right thing - spawning lots of goroutines is the normal API here)

pkg/blobstore/cloud/cloud_blob_access.go
asuffield: Good point, this isn't really useful if we don't also have AC working.

It'll have to be two separate API calls on the wire - one GET to retrieve the content because we need that anyway, and one COPY to update the metadata. Given that, the necessary change is straightforward, I'll sort it out.

pkg/digest/set_builder.go
asuffield: sync.Mutex should be cheap when not contended, but this belongs in the parallelisation PR anyway

pkg/blobstore/cloud/cloud_blob_access.go
asuffield: Duly fixed - it's easy because this COPY can be entirely async

pkg/blobstore/cloud/cloud_blob_access.go
```diff
@@ -41,6 +44,14 @@ func (ba *cloudBlobAccess) Get(ctx context.Context, digest digest.Digest) buffer
 		}
 		return buffer.NewBufferFromError(err)
 	}
+
+	go func() {
+		err := ba.touchBlob(ctx, digest)
```
EdSchouten, line 49: This approach has two downsides:
- There is no back-pressure.
- Errors are reported through log.Printf(), meaning they will only be visible in server logs, which are hardly ever checked.

I had exactly the same problem in some of the other code in the tree, which is why I added the `buffer.WithBackgroundTask()` function. With that function you can let a buffer hang right before it's closed and inject a custom error. Could you change this code to use that function instead? I guess you'll need to write something like this:

```go
b, t := buffer.WithBackgroundTask(ba.readBufferFactory.NewBufferFromReader(...))
go func() {
	err := ba.touchBlob(ctx, digest)
	if err != nil {
		err = util.StatusWrap(err, "Failed to refresh blob")
	}
	t.Finish(err)
}()
return b
```

pkg/blobstore/cloud/cloud_blob_access.go
```diff
@@ -77,16 +88,31 @@ func (ba *cloudBlobAccess) Put(ctx context.Context, digest digest.Digest, b buff
...
 	}
+
 	return missing.Build(), nil
 }
 
 func (ba *cloudBlobAccess) getKey(digest digest.Digest) string {
 	return ba.keyPrefix + digest.GetKey(ba.digestKeyFormat)
 }
+
+func (ba *cloudBlobAccess) touchBlob(ctx context.Context, blobDigest digest.Digest) error {
```
EdSchouten, line 112: Nit: If you change this function to take a `key string` instead of a `blobDigest digest.Digest`, `ba.getKey()` gets called less frequently.

pkg/blobstore/cloud/s3_copy_mutator.go
```diff
@@ -0,0 +1,34 @@
+package cloud
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/s3"
+
```
EdSchouten, line 9: Nit: Remove this empty line. The current convention is to group imports of the same 2nd level domain together.

pkg/blobstore/cloud/cloud_blob_access.go
```diff
@@ -77,16 +88,31 @@ func (ba *cloudBlobAccess) Put(ctx context.Context, digest digest.Digest, b buff
...
+	key := ba.getKey(blobDigest)
+	// Touch the object to update its modification time, so that cloud expiration policies will be LRU
```
EdSchouten, line 114: Please end all sentences in a comment with a full stop.

pkg/blobstore/cloud/cloud_blob_access_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/buildbarn/bb-storage/pkg/digest"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"gocloud.dev/blob/memblob"
+)
+
+// These tests use memblob as a fake bucket. blob.Bucket isn't
+// reasonably mockable, its methods return struct types which can't be
```
EdSchouten, line 19: This is interesting, isn't it? gocloud is supposed to offer abstractions. Meanwhile blob.Bucket is a concretion that can't easily be mocked.

pkg/blobstore/cloud/s3_copy_mutator.go
```diff
@@ -0,0 +1,34 @@
...
+	clock         clock.Clock
+}
+
+// NewS3CopyMutator creates a BeforeCopyFunc which updates the "Used"
+// metadata field on s3 objects with the current time.
+func NewS3CopyMutator(minRefreshAge time.Duration, clock clock.Clock) BeforeCopyFunc {
+	return func(asFunc func(interface{}) bool) error {
+		var input *s3.CopyObjectInput
+		if !asFunc(&input) {
+			return fmt.Errorf("failed to get CopyObjectInput - not an s3 bucket?")
```
EdSchouten, line 24: I think this can only occur in case of actual programming mistakes, right? Maybe better to put a `panic()` here, so that we crash intentionally?

pkg/blobstore/cloud/s3_copy_mutator.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+type s3CopyMutator struct {
+	minRefreshAge time.Duration
+	clock         clock.Clock
+}
+
+// NewS3CopyMutator creates a BeforeCopyFunc which updates the "Used"
+// metadata field on s3 objects with the current time.
+func NewS3CopyMutator(minRefreshAge time.Duration, clock clock.Clock) BeforeCopyFunc {
```
EdSchouten, line 20: Could we at least add a single unit test for this type, for the success case?

pkg/blobstore/cloud/cloud_blob_access.go
asuffield: Simple enough, done.

pkg/blobstore/cloud/cloud_blob_access.go
asuffield: I believe this evaporates when compiled anyway, but sure, it makes little difference

pkg/blobstore/cloud/s3_copy_mutator.go
asuffield: Or API drift in gocloud, since we went through an interface. crash vs failing all operations is marginal, but I suppose there's no harm in a panic.

pkg/blobstore/cloud/cloud_blob_access_test.go
asuffield: It would be reasonable if memblob was a little more complete - it isn't quite rich enough to be a full fake, there's no way to test the error paths. I might send them a PR later.

pkg/blobstore/cloud/cloud_blob_access.go
asuffield: You should probably express that opinion in the form of a linter.

pkg/blobstore/cloud/s3_copy_mutator.go
asuffield: Done

pkg/blobstore/cloud/s3_copy_mutator.go
asuffield: Done

timuralp:
I stumbled across this PR while investigating a different issue and wanted to make one comment: can the metadata update behavior be optional? I may be misreading it, but it seems the metadata update request itself is unconditional, while including an `x-amz-copy-source-if-unmodified-since` header? I suspect there may be consumers who do not want the behavior of always making a conditional update request.

asuffield:
It's certainly possible - but why would anybody want that behaviour? Conditional updates don't do anything if the condition is not met, so if you set the refresh age to be sufficiently high (longer than your expiry time will suffice) then it should be identical behavior to what it did before.

timuralp:
> It's certainly possible - but why would anybody want that behaviour?

I see two reasons. One is that some S3-clones do not implement the `x-amz-copy-source-if-unmodified` headers, while supporting the bulk of the API. In that case the conditional updates will either be rejected or turn into unconditional updates.

The second reason is that this double the total number of requests per object. For example, if there is a caching proxy that can satisfy most of the GET/HEAD requests and greatly reduce the build times, it is no longer as effective as these conditional updates have to be made to the source object store. In these cases, being able to disable this behavior would be helpful.

asuffield:
Some performance testing later: the parallelisation change is really important to this feature, there is too much end-to-end latency otherwise.

asuffield:
> Some performance testing later: the parallelisation change is really important to this feature, there is too much end-to-end latency otherwise.

More detail: the parallelisation change is really important to using cloud as the single backend in general. I dropped the small-object redis layer because that doesn't have the same behaviour, and that's what messed it all up.

asuffield:
> > Some performance testing later: the parallelisation change is really important to this feature, there is too much end-to-end latency otherwise.
> 
> More detail: the parallelisation change is really important to using cloud as the single backend in general. I dropped the small-object redis layer because that doesn't have the same behaviour, and that's what messed it all up.

Conclusion: this change is correct but CloudBlobAccess is too slow to be useful for objects smaller than multiple-MB, which also means it isn't useful for AC. This change doesn't appear to make the problem any better or worse. My CAS happens to be almost all small objects, which CloudBlobAccess is terrible for.

A following PR will add the same behaviour to RedisBlobAccess, since that looks easy.

pkg/blobstore/cloud/s3_copy_mutator.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/buildbarn/bb-storage/pkg/clock"
+)
+
+type s3CopyMutator struct {
```
EdSchouten, line 11: This struct can be removed now, right?

pkg/blobstore/cloud/s3_copy_mutator.go
```diff
@@ -0,0 +1,32 @@
...
+	minRefreshAge time.Duration
+	clock         clock.Clock
+}
+
+// NewS3CopyMutator creates a BeforeCopyFunc which updates the "Used"
+// metadata field on s3 objects with the current time.
+func NewS3CopyMutator(minRefreshAge time.Duration, clock clock.Clock) BeforeCopyFunc {
```
EdSchouten, line 18: Nit: Call this NewS3LRURefreshingBeforeCopyFunc() or something? Current Buildbarn convention is to name constructor-like functions after the {interface,function type} they create.

pkg/blobstore/cloud/s3_copy_mutator_test.go
```diff
@@ -0,0 +1,38 @@
...
+		// does it this way, so we follow.
+		switch v := i.(type) {
+		case **s3.CopyObjectInput:
+			*v = &input
+			return true
+		}
+		return false
+	}))
+
+	require.Equal(t, now.String(), *input.Metadata["Used"])
```
EdSchouten, line 35: I would suggest that something like this is done:
```go
require.Equal(t, s3.CopyObjectInput{
    ...
}, input)
```
The testify package is capable of printing diffs between structs pretty well.

pkg/blobstore/cloud/cloud_blob_access_test.go
```diff
@@ -0,0 +1,108 @@
...
+		require.NoError(t, err)
+		require.Equal(t, existsBefore, true)
+
+		_, err = blobAccess.Get(ctx, blobDigest).ToByteSlice(100)
+		require.Equal(t, status.Error(codes.Internal, "Buffer has checksum 787ec76dcafd20c1908eb0936a12f91edd105ab5cd7ecc2b1ae2032648345dff, while 64ec88ca00b268e5ba1a35678a1b5316d212f4f366b2477232534a8aeca37f3c was expected"), err)
+
+		// Postcondition: repair deletes the incorrect object
+		existsAfter, err := bucket.Exists(ctx, bucketKey)
+		require.NoError(t, err)
+		require.Equal(t, existsAfter, false)
```
EdSchouten, line 66: require.False()

EdSchouten (approved):
This is starting to look really good. One more round of feedback and then we should be ready to land this.

pkg/blobstore/cloud/s3_copy_mutator.go
asuffield: Done

PR #83, opened 2020-08-14
WIP: Add basic query logging

asuffield:
This is some minimal logging to aid debugging - for now, mostly only useful for tracking down issues with http routing.

### DIFF
--- a/pkg/global/BUILD.bazel
+++ b/pkg/global/BUILD.bazel
@@ -2,7 +2,10 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library")
 
 go_library(
     name = "go_default_library",
-    srcs = ["apply_configuration.go"],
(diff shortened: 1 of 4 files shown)


EdSchouten:
My concern with changes like these is that over time people will need logging in al sorts of places, and that in each of those places we het more boilerplate code to log an arbitrary set of things.

Can’t Jaeger/Stackdriver be used for stuff like this?

asuffield:
That's an interesting question which I need to dig into a bit more - there probably is a reasonable solution based around those libraries (opentelemetry is likely the way to go now). Something like this PR will still be needed, to generate the data, but we can probably do better than what I have here right now.

asuffield:
This has turned into several other PRs, and the actual log-writing piece can be done by the opentelemetry-collector

PR #85, opened 2020-08-20
Add trace span details

asuffield:
This injects new spans in MetricsBlobAccess so that traces show the
path through BlobAccess layers, and inserts annotations into
BuildQueue spans to identify the digests involved.

This isn't the last word in tracing, but it does appear to be the
critical pieces of data for these operations.


pkg/blobstore/metrics_blob_access.go
```diff
@@ -86,12 +89,18 @@ func (ba *metricsBlobAccess) updateDurationSeconds(vec prometheus.ObserverVec, c
 }
 
 func (ba *metricsBlobAccess) Get(ctx context.Context, digest digest.Digest) buffer.Buffer {
+	ctx, span := trace.StartSpan(ctx, ba.name+".Get")
```
EdSchouten, line 92: My suggestion would be to add a separate TracingBlobAccess. There is no need to let this be part of MetricsBlobAccess, right?

The advantage of having a separate TracingBlobAccess is that it allows us to prevent the overhead on systems where tracing is not enabled.

pkg/blobstore/metrics_blob_access.go
```diff
@@ -127,20 +150,31 @@ func (ba *metricsBlobAccess) FindMissing(ctx context.Context, digests digest.Set
...
 }
 
 func (eh *metricsErrorHandler) Done() {
 	eh.blobAccess.updateDurationSeconds(eh.blobAccess.getDurationSeconds, eh.errorCode, eh.timeStart)
+	eh.span.End()
+}
+
+func errToStatus(err error) trace.Status {
+	if err == nil {
+		return trace.Status{}
```
EdSchouten, line 177: I've seen us juggle with trace.Status in a couple of other places. Should we add a proper utility function for this to pkg/util/status.go?

pkg/builder/tracing_build_queue.go
```diff
@@ -0,0 +1,42 @@
+package builder
```
EdSchouten, line 1: Similar to what I stated in https://github.com/buildbarn/bb-remote-execution/pull/67, I don't think stuff like this should be added. Some generic gRPC middleware could do this for us.

pkg/blobstore/metrics_blob_access.go
asuffield: I was suspecting the same thing, let me work this bit into the opentelemetry conversion PR.

pkg/blobstore/metrics_blob_access.go
asuffield: There is indeed no need to put it in here, we can add another wrapper for tracing if you prefer.

This code is already near-zero-overhead. The opencensus and opentelemetry libraries handle this explicitly so that applications using them don't need to add extra logic for it. All of the calls here are no-ops if tracing is not enabled. The worst thing we do here is compute ba.name+".Get" and I can smooth that out if you think it's important (I expect the compiler to optimise it away).

There's a general principle here that if any overhead should be removed, it should be removed in the tracing library, rather than expecting applications to try to reproduce it.

asuffield:
I have a cleaned up version of this, but it wants to go in after the opentelemetry migration. Parking it for the moment.

EdSchouten:

moroten:
@EdSchouten In https://github.com/buildbarn/bb-remote-execution/pull/67 you were skeptical to adding this detailed information, is that still the case? I think this information would be useful and would still be limited compared to the amount of other data exported.

I suggest to expose the requested digest when reading a blob, but for `FindMissingBlobs` I suggest to only log the number digests. It might be difficult to draw the line about what data is interesting and not, maybe that can be adjusted in the future.

EdSchouten:
Hey! Just to make it clear: I'm not sceptical of adding this information. The only thing I'd like us to do is to have it as generic middleware, as opposed to annotating the entire codebase. Some config where you can say: "Of these RPCs, I want to dump the following Protobuf fields: ...."

EdSchouten:
FYI: This was fixed as part of https://github.com/buildbarn/bb-storage/commit/3f0259c1cd286522e4fc5c19642ce2d1fb819c2a

PR #86, opened 2020-08-20
Teach RedisBlobAccess to update TTL during Get and FindMissing calls

asuffield:
This makes the key_ttl feature be LRU-like instead of FIFO.

### DIFF
--- a/pkg/blobstore/redis_blob_access.go
+++ b/pkg/blobstore/redis_blob_access.go
@@ -26,18 +26,20 @@ type redisBlobAccess struct {
 	readBufferFactory  ReadBufferFactory
 	digestKeyFormat    digest.KeyFormat
 	keyTTL             time.Duration
+	refreshOnUse       bool
 	replicationCount   int64
 	replicationTimeout int
 }
 
 // NewRedisBlobAccess creates a BlobAccess that uses Redis as its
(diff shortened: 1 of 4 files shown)


pkg/blobstore/redis_blob_access.go
```diff
@@ -26,18 +26,20 @@ type redisBlobAccess struct {
 	readBufferFactory  ReadBufferFactory
 	digestKeyFormat    digest.KeyFormat
 	keyTTL             time.Duration
+	refreshOnUse       bool
```
EdSchouten, line 29: refreshTTLOnUse?

EdSchouten (commented):
To me it isn't really clear what the use of TTLs in Redis is in our case. Isn't it just sufficient to set the eviction policy to LRU on the server globally?

I read the Redis source code some time ago. As far as I know, EXISTS and GET operations also affect the eviction policy state as expected.

asuffield:
Interesting question, I will have to run performance tests to find out. If correct, we could maybe remove keyTTL entirely.

asuffield:
Some experimentation later, I have concluded that redis's own LRU is at least as good as this, so keyTTL is an antifeature. I'll follow up by removing it.

PR #88, opened 2020-08-22
Migrate to OpenTelemetry

asuffield:
This inserts the wiring for opentelemetry-collector into both the oc
and otel bindings, and inserts the opentelemetry instrumentation into
the places which it immediately supports.

The necessary incantations to map contexts are in pkg/util/trace for
calling libraries which use the other mode.

Switching buildbarn over to use otel natively will have to be a
follow-up change because it spans multiple git repositories.


asuffield:
Much refactoring later, pasting the new commit message here:

This inserts the wiring for opentelemetry-collector into both the oc
and otel bindings, and inserts the opentelemetry instrumentation into
the places which it immediately supports.

Existing uses of OpenCensus are updated to OpenTelemetry. The OC
exporters are retained for libraries which still use OC (notably
gocloud.dev), and contexts are mapped as needed.

Direct support for jaeger and stackdriver is dropped in favor of
configuring this in opentelemtry-collector, which supports far more
backends. I'll follow up with examples in bb-deployments.

A matching PR to bb-remote-execution will need to be merged along with
this - there turns out to be no neat way to do the transition.

pkg/blobstore/configuration/new_blob_access.go
```diff
@@ -240,6 +241,7 @@ func newNestedBlobAccessBare(configuration *pb.BlobAccessConfiguration, creator
 					ReadTimeout:     readTimeout,
 					WriteTimeout:    writeTimeout,
 				})
+
```
EdSchouten, line 244: Nit: unnecessary whitespace change? Better revert this to prevent merge conflicts for others.

pkg/global/apply_configuration.go
```diff
@@ -70,9 +42,20 @@ func ApplyConfiguration(configuration *pb.Configuration) error {
 			view.RegisterExporter(pe)
 		}
 
-		if tracingConfiguration.AlwaysSample {
-			trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
+		sampling := float64(tracingConfiguration.SampleProbability)
+		if otelConfiguration := tracingConfiguration.GetOtel(); otelConfiguration != nil {
```
EdSchouten, line 46: Nit: Only use `.GetFoo()` methods in case the Protobuf message may be nil. In this case it's guaranteed to be non-nil.

pkg/global/apply_configuration.go
```diff
@@ -70,9 +42,20 @@ func ApplyConfiguration(configuration *pb.Configuration) error {
 			view.RegisterExporter(pe)
 		}
 
-		if tracingConfiguration.AlwaysSample {
-			trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
+		sampling := float64(tracingConfiguration.SampleProbability)
```
EdSchouten, line 45: This is only used inside the if-statement. Was your intention to also use it for the ApplyConfig() call below?

pkg/global/apply_configuration.go
```diff
@@ -106,3 +89,107 @@ func ApplyConfiguration(configuration *pb.Configuration) error {
 
 	return nil
 }
+
+func applyOtel(config *pb.OpenTelemetryConfiguration, sampleProbability float64) error {
```
EdSchouten, line 93: applyOpenTelemetryConfiguration?

pkg/global/apply_configuration.go
```diff
@@ -70,9 +42,20 @@ func ApplyConfiguration(configuration *pb.Configuration) error {
...
+			}
 		}
+
+		if ocConfiguration := tracingConfiguration.GetOc(); ocConfiguration != nil {
+			if err := applyOc(ocConfiguration); err != nil {
+				return err
+			}
+		}
+
+		octrace.ApplyConfig(octrace.Config{DefaultSampler: octrace.ProbabilitySampler(float64(tracingConfiguration.SampleProbability))})
```
EdSchouten, line 58: I guess this casting could be removed, if the Protobuf configuration used `double` instead of `float`?

Why do we need to call into octrace here? Isn't this code supposed to be decoupled from OpenCensus/OpenTelemetry?

pkg/grpc/server.go
```diff
@@ -7,7 +7,7 @@ import (
 	configuration "github.com/buildbarn/bb-storage/pkg/proto/configuration/grpc"
 	"github.com/buildbarn/bb-storage/pkg/util"
 	"github.com/golang/protobuf/ptypes"
-	"github.com/grpc-ecosystem/go-grpc-prometheus"
+	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
```
EdSchouten, line 10: I've seen changes along these lines in the past. Why is this being done? It's not as if our imports are ambiguous. Could you please revert this?

pkg/proto/configuration/global/global.proto
```diff
@@ -56,18 +57,95 @@ message PrometheusPushgatewayConfiguration {
...
+  // Connection address for the collector. Default is localhost:55680.
+  string address = 1;
+
+  // TLS configuration. TLS is not enabled when left unset.
+  buildbarn.configuration.tls.ClientConfiguration tls = 2;
+
+  // Resource detectors. Labels will be merged, with earlier things in
+  // this entries being preferred over later ones.
+  enum ResourceDetector {
+    ENV = 0;  // Use OTEL_RESOURCE_ATTRIBUTES environment variable
```
EdSchouten, line 85: Do we need to use names this short? I'm fine with abbreviating "Amazon Web Services" to just AWS, but at least this configuration option could use a name that's more descriptive. USE_ENVIRONMENT_VARIABLES?

pkg/proto/configuration/global/global.proto
```diff
@@ -56,18 +57,95 @@ message PrometheusPushgatewayConfiguration {
...
+  repeated ResourceDetector resource_detectors = 3;
+
+  // Queue size for buffering spans. Default is 2048.
+  int32 max_queue_size = 4;
+
+  // If true, then span Start() and End() functions will block when
+  // the queue is full. This will significantly degrade performance,
+  // but does guarantee no spans are lost, which can be useful when
+  // debugging with the sampling probability set to 1.0.
+  bool block_on_queue_full = 5;
```
EdSchouten, line 99: block_when_queue_is_full or block_on_full_queue?

pkg/proto/configuration/global/global.proto
```diff
@@ -56,18 +57,95 @@ message PrometheusPushgatewayConfiguration {
...
+  // },
+  // otel: {
+  //   resourceDetectors: ['ENV'],
+  // },
+  //
+  // And set the environment variable OTEL_RESOURCE_ATTRIBUTES to
+  // 'service.name=buildbarn'.
+
+  // Export data to an OpenCensus agent.
+  OpenCensusConfiguration oc = 6;
```
EdSchouten, line 145: open_census

pkg/proto/configuration/global/global.proto
```diff
@@ -56,18 +57,95 @@ message PrometheusPushgatewayConfiguration {
...
+
+  // Export data to an OpenTelemetry collector.
+  OpenTelemetryConfiguration otel = 7;
```
EdSchouten, line 148: open_telemetry

pkg/proto/configuration/global/global.proto
```diff
@@ -56,18 +57,95 @@ message PrometheusPushgatewayConfiguration {
...
+  // The intended usage is to run the opentelemetry-collector agent as
+  // a sidecar, with both the opencensus and otel receivers enabled,
+  // and configure that agent to send traces to your preferred
+  // backend.
+
+  // In order to fully capture all traces, it is currently necessary
+  // to configure both oc and otel, because not all libraries have
+  // migrated to OpenTelemtry yet. A sufficient minimal configuration is:
+  //
+  // oc: {
```
EdSchouten, line 134: Please use Jsonnet syntax for configuration examples.

pkg/util/trace.go
```diff
@@ -0,0 +1,54 @@
...
+	oc "go.opencensus.io/trace"
+	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/propagation"
+	otel "go.opentelemetry.io/otel/api/trace"
+	"go.opentelemetry.io/otel/codes"
+	"google.golang.org/grpc/status"
+)
+
+// PropagateOtelToOpencensus arranges for spans that let us trace into libraries using opencensus.
+func PropagateOtelToOpencensus(ctx context.Context) context.Context {
```
EdSchouten, line 15: PropagateOpenTelemetryToOpenCensus?

pkg/blobstore/circular/circular_blob_access.go
```diff
@@ -7,15 +7,18 @@ import (
 	"io/ioutil"
```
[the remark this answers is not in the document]
asuffield: It wouldn't reasonably compile without it, this seemed easiest. I don't expect it to get used.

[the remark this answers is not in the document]
EdSchouten: Feel free to remove it then. :-)

cmd/bb_replicator/main.go
```diff
@@ -66,6 +67,7 @@ func main() {
 
 	// Web server for metrics and profiling.
 	router := mux.NewRouter()
+	router.Use(muxtrace.Middleware("bb_replicator"))
```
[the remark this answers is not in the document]
asuffield: Done, will split this out

pkg/global/apply_configuration.go
asuffield: Yes, fixed

pkg/proto/configuration/global/global.proto
asuffield: I went for matching the names used in otel: https://pkg.go.dev/go.opentelemetry.io/otel/sdk@v0.11.0/trace?tab=doc#BatchSpanProcessorOptions

While the grammar is questionable, consistency seems better.

pkg/proto/configuration/global/global.proto
asuffield: Done, I think

pkg/util/trace.go
asuffield: Sure, a few more characters don't make it more comical than it already is

pkg/grpc/server.go
asuffield: The reason is specifically that goimports doesn't believe you and wants to remove it, even though it does in fact compile.

pkg/global/apply_configuration.go
asuffield: Ah yes, protobuf does have a double type.

It doesn't matter much, but I moved this into applyOpenTelemetry. There's no semantic difference.

pkg/proto/configuration/global/global.proto
asuffield: It's not necessary, I was just matching the names of the detectors in the otel library. Changed.

pkg/grpc/server.go
asuffield: I've fumbled these through for now, we can puzzle out goimports independently.

pkg/util/http_client.go
```diff
@@ -0,0 +1,12 @@
+package util
+
+import (
+	"net/http"
+
+	otelhttp "go.opentelemetry.io/contrib/instrumentation/net/http"
+)
+
+// NewHTTPClient constructs an HTTP client with instrumentation.
+func NewHTTPClient() *http.Client {
```
EdSchouten, line 10: What are your thoughts on using this occasion to move the HTTPClient interface we have in pkg/blobstore to this file as well?

```go
// HTTPClient is an interface around Go's standard HTTP client type. It          
// has been added to aid unit testing.                                           
type HTTPClient interface {
        Do(req *http.Request) (*http.Response, error)
}
```

The convention so far is also to not use NewFoo() functions in case singleton instances suffice. We could provide a default HTTP client object using something like this:

```go
var DefaultHTTPClient HTTPClient = &http.Client{
        Transport: otelhttp.NewTransport(http.DefaultTransport),
}
```

Or are you dealing with certain pieces of code that strongly depend on *http.Client instead of using an interface type? If so, we could declare it without the `HTTPClient` in between and add a compile-time type assertion next to it.

pkg/proto/configuration/global/global.proto
```diff
@@ -56,18 +57,96 @@ message PrometheusPushgatewayConfiguration {
...
+  // Jaeger configuration for tracing has been replaced by oc and otel.
+  reserved 1;
 
-  // Stackdriver Trace configuration for tracing.
-  StackdriverConfiguration stackdriver = 2;
+  // Stackdriver Trace configuration has been replaced by oc and otel.
+  reserved 2;
 
   // Export stats on traces to Prometheus.
   bool enable_prometheus = 3;
```
EdSchouten, line 117: Is my observation correct that this option only applies to just OpenCensus and not OpenTracing? If so, would it make sense to move it into the relevant nested message?

pkg/proto/configuration/global/global.proto
```diff
@@ -56,18 +57,96 @@ message PrometheusPushgatewayConfiguration {
...
+
+  // Maximum number of spans to batch before sending. Default is 512.
+  int32 max_export_batch_size = 7;
+}
+
 message TracingConfiguration {
-  // Jaeger configuration for tracing.
-  JaegerConfiguration jaeger = 1;
+  // Jaeger configuration for tracing has been replaced by oc and otel.
+  reserved 1;
```
EdSchouten, line 111: Nit: This is still referring to the old field names.

EdSchouten (approved):
This is starting to look really good. I think this is (almost) good to go.

Thanks for working on this, Andrew. Really appreciated!

pkg/proto/configuration/global/global.proto
asuffield: Hmm, interesting question. It is currently true that it's only wired up to OpenCensus. That's mostly because the OpenTelemetry version of this isn't very complete yet (tracing is much further ahead) so I didn't have a sensible migration path right now. At some point in the future, it should be handled on the otel side... does it make more sense to leave it here as a generic config knob that's somehow implemented, or to move it around as the code does?

pkg/util/http_client.go
asuffield: This turns out to work, done.

pkg/proto/configuration/global/global.proto
EdSchouten: Sorry for not responding to this earlier.

I'd suggest we move it around as the code does, right? That both keeps the code self-contained. One added 'advantage' is that it causes startup to fail if people refer to OpenCensus by the time we remove support for it entirely. It's a good way to signal people of changes.

pkg/util/trace.go
```diff
@@ -0,0 +1,56 @@
...
+
+	oc "go.opencensus.io/trace"
+	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/propagation"
+	otel "go.opentelemetry.io/otel/api/trace"
+	"go.opentelemetry.io/otel/codes"
+	"google.golang.org/grpc/status"
+)
+
+// PropagateOpenTelemetryToOpenCensus arranges for spans that let us trace into libraries using opencensus.
```
EdSchouten, line 14: Nit: s/opencensus/OpenCensus/

EdSchouten (approved):
This looks good to land after the enable_prometheus boolean is moved. :tada:

pkg/proto/configuration/global/global.proto
asuffield: Done

asuffield:
Rebased and finished off. Now we just need to puzzle out the last bits of https://github.com/buildbarn/bb-remote-execution/pull/69 before merging this (merging either one without the other will inevitably break tracing - we could beat them into compiling but it wouldn't actually work).

pkg/util/trace.go
```diff
@@ -0,0 +1,56 @@
...
+	oc "go.opencensus.io/trace"
+	"go.opentelemetry.io/otel/api/global"
+	"go.opentelemetry.io/otel/api/propagation"
+	otel "go.opentelemetry.io/otel/api/trace"
+	"go.opentelemetry.io/otel/codes"
+	"google.golang.org/grpc/status"
+)
+
+// PropagateOpenTelemetryToOpenCensus arranges for spans that let us trace into libraries using OpenCensus.
+func PropagateOpenTelemetryToOpenCensus(ctx context.Context) context.Context {
```
EdSchouten, line 15: I thought I'd help you out with this PR by removing CloudBlobAccess from the tree.

This should allow you to trim down this PR by removing any OpenCensus related bits, right?

asuffield: Rebased and cleaned up.

EdSchouten:
Hi @asuffield! With OpenTelemetry almost being released, I thought now would be the right moment to switch to it. Unfortunately, this PR didn't apply anymore, due to massive API changes between v0.20.0 and v1.0.0-RC2, which is why I went ahead and effectively wrote my own version, using this PR as inspiration.

Thanks again for your contribution! I hope the changes I made also work for you!

asuffield:
Woo! We were just talking about needing to revive this one to get rid of our patch, that solves most of the problem.

I think the only thing we'll need to slot in is the otlp exporter, which should now be trivial.

EdSchouten:
Great! That's nice to hear. Feel free to submit a PR to get that added. Enjoy!

PR #89, opened 2020-08-24
Prefer WaitForReady on grpc clients

asuffield:
This CallOption causes client calls to prefer to block when the
underlying transport is broken (any form of tcp connection issue)
rather than fail the call instantly with Unavailable. Deadlines are
still respected, but this will make grpc wait until the call deadline
is reached if a backend is not responding.

For bazel builds, we consistently prefer to take the latency hit of
waiting for a connection to recover rather than instantly failing the
build when any glitch happens.

### DIFF
--- a/pkg/grpc/base_client_factory.go
+++ b/pkg/grpc/base_client_factory.go
@@ -34,6 +34,10 @@ func (cf baseClientFactory) NewClientFromConfiguration(config *configuration.Cli
 
 	dialOptions := []grpc.DialOption{
 		grpc.WithStatsHandler(&ocgrpc.ClientHandler{}),
+		// We consistently prefer to wait for a connection rather than
+		// abort builds when there is a tcp error. Can be overridden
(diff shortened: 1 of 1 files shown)


EdSchouten:
I don't know whether I'm in favour of changes like these.

Right now Buildbarn always fails immediately and relies on Bazel to do retrying. The advantage of this approach is that the client is continuously informed of what's going on in the backend. It can display helpful messages during a build that show why things are going slowly. By enabling options like the one you proposed, this property no longer holds.

asuffield:
I can dig a little deeper if you like. The observed behaviour is that bazel failed instantly when faced with this particular error code, and never retried. Something in there is clearly not right, this might not be the correct solution.

asuffield:
I still haven't tracked down what causes those intermittent failures, but this doesn't seem to solve them anyway. I think I'm hunting a bazel bug....

PR #90, opened 2020-08-25
update to go-redis v8

asuffield:
Also Inject the necessary wiring for OpenTelemetry, which isn't
enabled by default. This won't do anything immediately (the hook is a
no-op if the otel library doesn't think the span is sampled, so we
need the otel PR to enable it), but this seems to belong here rather
than in the otel PR.


pkg/blobstore/configuration/new_blob_access.go
```diff
@@ -239,9 +240,15 @@ func newNestedBlobAccessBare(configuration *pb.BlobAccessConfiguration, creator
 					DialTimeout:     dialTimeout,
```
EdSchouten, line 240: Now that we're switching to a version of go-redis that has support for context, is there still any value in providing config options for DialTimeout, ReadTimeout and WriteTimeout?

In an ideal world, all of Buildbarn's components should continue to go on, as long as Bazel on the client side is willing to do so.

asuffield: This is an excellent question which I am currently in the middle of learning the answer to, because something in here is misbehaving. I'll probably know the answer in the next day or two.

asuffield: So the answer turns out to be that these are the settings for net.Dialer, mirroring those names, but somewhat misleadingly. They're the timeouts for a single attempt at a dial/read/write, which are inside the retry loop. MaxRetries * DialTimeout needs to be significantly smaller than the context deadline, so they are both useful and important.

I shall document them better.

pkg/blobstore/configuration/new_blob_access.go
```diff
@@ -239,9 +240,15 @@ func newNestedBlobAccessBare(configuration *pb.BlobAccessConfiguration, creator
...
 					ReadTimeout:     readTimeout,
 					WriteTimeout:    writeTimeout,
+					NewClient: func(opt *redis.Options) *redis.Client {
```
EdSchouten, line 243: If you were to place this function into a free function somewhere at the top of the file, you could also call into it below to construct the single Redis client, right?

Wouldn't it make more sense to make this change as part of the OpenTelemetry PR?

EdSchouten (approved):

pkg/blobstore/configuration/new_blob_access.go
asuffield: The OpenTelemetry PR doesn't have a redis version which includes redisext.OpenTelemetryHook, so there isn't an obviously correct ordering here.

I can split the function out, although it then ends up quite distant from this code. (Tempting to split up new_blob_access, it's something of a skyscraper)

pkg/blobstore/configuration/new_blob_access.go
asuffield: Documented in the .proto accordingly. They turn out to be important in the clustered configuration, because the defaults are way too long to be hanging on a single node when you could be trying a different one.

PR #91, opened 2020-09-01
util.StatusWrapfWithCode will panic when err == nil

asuffield:


EdSchouten (approved):
Would it be possible to somehow introduce a unit test for this? If the answer is: "It's too much of a hassle", then I'm also fine with that. :-)

asuffield:
I had a go but it turns out to be quite fiddly to test go-redis, because results get returned by mutating redis.NewIntCmd and there's no public API for doing the same thing.

PR #94, opened 2020-09-02
configurable logging for aws-sdk-go

asuffield:
Wire up the aws-sdk-go logging options to the config file.

### DIFF
--- a/pkg/cloud/aws/session.go
+++ b/pkg/cloud/aws/session.go
@@ -31,5 +31,27 @@ func NewSessionFromConfiguration(configuration *aws_pb.SessionConfiguration) (*s
 			staticCredentials.SecretAccessKey,
 			"")
 	}
+	logLevel := aws.LogOff
+	if configuration.GetLogDebug() {
+		logLevel |= aws.LogDebug
+	}
+	if configuration.GetLogSigning() {
+		logLevel |= aws.LogDebugWithSigning
+	}
+	if configuration.GetLogHttpBody() {
+		logLevel |= aws.LogDebugWithHTTPBody
+	}
+	if configuration.GetLogRequestRetries() {
+		logLevel |= aws.LogDebugWithRequestRetries
+	}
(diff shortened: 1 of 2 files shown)


pkg/proto/configuration/cloud/aws/aws.proto
```diff
@@ -30,4 +30,13 @@ message SessionConfiguration {
   // will search the default credential provider chain (e.g.,
   // environment variables, EC2 instance IAM roles).
   StaticCredentials static_credentials = 5;
+
+  // Log levels are documented at
+  // https://docs.aws.amazon.com/sdk-for-go/api/aws/#LogLevelType
+  bool log_debug = 6;
```
EdSchouten, line 36: Would it make sense to use a `repeated enum` instead?
```go
enum LogLevelType {
  DEBUG = 1;
  DEBUG_WITH_WHATEVER = 2;
  ...
}
repeated LogLevelType log_level = 6;
```
That way it's easier to parameterize/forward/... log levels without needing to resort to merging objects.

pkg/cloud/aws/session.go
```diff
@@ -31,5 +31,27 @@ func NewSessionFromConfiguration(configuration *aws_pb.SessionConfiguration) (*s
...
+		logLevel |= aws.LogDebugWithRequestRetries
+	}
+	if configuration.GetLogRequestErrors() {
+		logLevel |= aws.LogDebugWithRequestErrors
+	}
+	if configuration.GetLogEventStreamBody() {
+		logLevel |= aws.LogDebugWithEventStreamBody
+	}
+	if logLevel != aws.LogOff {
+		cfg.LogLevel = &logLevel
```
EdSchouten, line 54: LogOff is the default, right? In that case it wouldn't hurt to just do this assignment unconditionally I suppose?

asuffield:
If we're ditching cloud then I guess there's no real need to finish this off.

PR #96, opened 2020-09-07
Satiate 'go mod tidy' by providing package comments in otherwise "empty" packages

asuffield:
go mod tidy fails to finish its job of cleaning up unused dependencies
because it is confused by the packages which contains only protos
(since these will be generated by bazel, there are no .go
files). Satisfy it that these packages do exist by adding package
comments to all of them.

At this point, go mod tidy removes ineffassign as unused. Following
the pattern in
https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md
we create a dummy tools dependency which is otherwise ignored, so that
go mod tidy understands we want to keep that one.


asuffield:
golint hates the package names. How about proto/configuration/storage instead of proto/configuration/bb_storage?

Or we could drop golint in favour of a linter that wants to be used this way

tools.go
```diff
@@ -0,0 +1,5 @@
+// +build tools
+
+package tools
+
+import _ "github.com/gordonklaus/ineffassign"
```
EdSchouten, line 5: Would it make sense to remove this, but add something similar to #97? Could we slap an underscore import into //tools/tidy itself?

pkg/proto/replicator/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package replicator contains the grpc protocol for interacting with bb_replicator
```
EdSchouten, line 1: What's the convention for documentation strings like these? Does it always have to start with `"Package %s"`, or would it be a bit more tidy to remove that prefix entirely?
```go
// The gRPC protocol for that BlobReplicator uses to communicate with instances of bb_replicator.
```

Nits:
- Please add a full stop to the end of every sentence.
- Make sure to wrap lines at <80 columns.
- s/grpc/gRPC/.

EdSchouten (commented):
So I assume the package level documentation is what shows up in the Synopsis column on this page:

https://pkg.go.dev/github.com/buildbarn/bb-storage

Is there some automated way we can check that we don't forget to add these going forward? For example, is there some kind of golint option? By enabling that, we can ensure that we never forget to add these dummy files to the tree for Protobufs.

tools.go
asuffield: That's also possible - the unbuildable tools.go is the conventional approach for this case, but we could converge it with #97. We'd probably have to merge #97 first before that's feasible.

pkg/proto/replicator/doc.go
asuffield: > So I assume the package level documentation is what shows up in the Synopsis column on this page:
> 
> https://pkg.go.dev/github.com/buildbarn/bb-storage
> 
> Is there some automated way we can check that we don't forget to add these going forward? For example, is there some kind of golint option? By enabling that, we can ensure that we never forget to add these dummy files to the tree for Protobufs.

A little digging around indicates that the Synopsis column gets the first sentence of the package comment, and the whole comment shows up in the Overview section: https://pkg.go.dev/go.opentelemetry.io/otel@v0.11.0/api/correlation?tab=doc#pkg-overview

golint doesn't have any options, as part of its policy of being borderline unusable in CI (https://github.com/golang/lint#purpose), but https://staticcheck.io/docs/checks#ST1000 would cover it.

I am tempted to say: put golint in the bin, use https://golangci-lint.run/ and enable the interesting stuff, which also solves the tedious opinions it has about the bb_storage package name. However, that's probably a different PR, and I have a simpler solution in the short term: go mod tidy exits 1 on master because it can't find the empty packages. That's sufficient to test the thing we care about here, which is that go mod tidy can run to completion.

asuffield: > What's the convention for documentation strings like these? Does it always have to start with `"Package %s"`, or would it be a bit more tidy to remove that prefix entirely?

Interesting question, I had a look around to find out.

The standard library almost, but not quite, always uses this form: https://pkg.go.dev/std@go1.15.2?tab=packages

Other libraries are a bit more variable: https://pkg.go.dev/go.opentelemetry.io/otel@v0.11.0?tab=subdirectories

I don't have a strongly held opinion here, because realistically, these are probably placeholders which never get read by anybody. Protobuf packages aren't going to have their documentation in godoc, it would be produced by https://github.com/pseudomuto/protoc-gen-doc instead.

asuffield: > Nits:
> 
> * Please add a full stop to the end of every sentence.
> * Make sure to wrap lines at <80 columns.
> * s/grpc/gRPC/.

Duly tidied.

asuffield:
go mod tidy is added to CI and correctly identified the two new packages, which are now fixed to also have comments.

pkg/proto/replicator/doc.go
EdSchouten: These two files still need to be tidied up:

- pkg/proto/icas/doc.go
- pkg/proto/replicator/doc.go

asuffield: Got those two now.

golangci-lint is going to take more effort to beat into working with bazel, so I'll stick a pin in that idea for now.

> > What's the convention for documentation strings like these? Does it always have to start with `"Package %s"`, or would it be a bit more tidy to remove that prefix entirely?
> 
> Interesting question, I had a look around to find out.

It turns out that golint has strongly held opinions on this topic, but I'm rapidly losing interest in its opinions. I haven't puzzled out a workaround for it yet though.

EdSchouten:
This should be ready for merge, as soon as the CI failures are fixed.

EdSchouten (approved):
LGTM! Will merge as soon as CI has completed.

asuffield:
> This should be ready for merge, as soon as the CI failures are fixed.

Swatted the strays, but it's going to fail on golint objecting to the package names bb_storage and bb_replicator still. What do we want to do there? I could rename them to remove the underscores, or try harder to replace golint with something more flexible.

EdSchouten:
> Swatted the strays, but it's going to fail on golint objecting to the package names bb_storage and bb_replicator still. What do we want to do there? I could rename them to remove the underscores, or try harder to replace golint with something more flexible.

You can't pick an arbitrary name, because it has to match up with the name used in Protobuf generated code, right?

I saw in some Gazelle/rules_go bug report that it's also possible to just add a dummy.go with contents like these:
```go
// +build ignore
package dummy
```
Providing package-level documentation becomes a bit useless then, because it's likely not picked up because of the `// +build ignore`. I'm fine with omitting the docs then.

asuffield:
> > Swatted the strays, but it's going to fail on golint objecting to the package names bb_storage and bb_replicator still. What do we want to do there? I could rename them to remove the underscores, or try harder to replace golint with something more flexible.
> 
> You can't pick an arbitrary name, because it has to match up with the name used in Protobuf generated code, right?
> 
> I saw in some Gazelle/rules_go bug report that it's also possible to just add a dummy.go with contents like these:
> 
> ```go
> // +build ignore
> package dummy
> ```

I tried this idea and it almost works, but then `go mod tidy` doesn't believe in the package because it can't find any non-excluded source files, so we're back where we started.

It does indeed need to match the protobuf code. It's unpretty, but my last remaining idea is to just rename the packages. I have grudgingly removed the bb_ prefix from them and fixed up the references.

EdSchouten:
I basically gave up on getting `go mod` to work properly with *.pb.go files being absent. It doesn't work well for bb-storage, but it's even worse now for repositories that depend on bb-storage. The *.pb.go files are now being checked in.

In other words, there is less need for a change like this. Marking this as closed. Thanks for your contribution regardless!

PR #97, opened 2020-09-07
Convert the style and lint bits of github workflows into a runnable command

asuffield:
It's not pretty, and it's very much writing a shell script in golang,
but it lets me run the whole thing locally instead of repeatedly
pushing and waiting for the github workflow to tell me that I don't
quite have the formatting the way it wants.

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -9,36 +9,12 @@
                "uses": "actions/checkout@v1"
             },
             {
-               "name": "Buildifier",
-               "run": "bazel run @com_github_bazelbuild_buildtools//:buildifier"
-            },
-            {
-               "name": "Gazelle",
-               "run": "bazel run //:gazelle -- update-repos -from_file=go.mod -to_macro go_dependencies.bzl%go_dependencies -prune && bazel run //:gazelle"
-            },
-            {
-               "name": "Gofmt",
-               "run": "bazel run @go_sdk//:bin/gofmt -- -s -w ."
-            },
-            {
-               "name": "Clang format",
-               "run": "find . -name '*.proto' -exec bazel run @llvm_toolchain//:bin/clang-format -- -i {} +"
-            },
-            {
-               "name": "GitHub workflows",
-               "run": "bazel build //tools/github_workflows && cp bazel-bin/tools/github_workflows/* .github/workflows"
+               "name": "Tidy and lint",
+               "run": "bazel run //tools/tidy"
             },
             {
                "name": "Test style conformance",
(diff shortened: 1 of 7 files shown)


.github/workflows/master.yaml
```diff
@@ -9,37 +9,13 @@
...
-            },
-            {
-               "name": "Clang format",
-               "run": "find . -name '*.proto' -exec bazel run @llvm_toolchain//:bin/clang-format -- -i {} +"
-            },
-            {
-               "name": "GitHub workflows",
-               "run": "bazel build //tools/github_workflows && cp bazel-bin/tools/github_workflows/* .github/workflows"
+               "name": "Tidy and lint",
+               "run": "bazel run //tools/tidy"
```
EdSchouten, line 13: This should likely be prefixed with `@com_github_buildbarn_bb_storage`, right? Otherwise it'll break for the other repositories.

BUILD.bazel
```diff
@@ -6,8 +6,10 @@ load("@com_github_bazelbuild_buildtools//buildifier:def.bzl", "buildifier")
 # gazelle:resolve proto go build/bazel/remote/execution/v2/remote_execution.proto @com_github_bazelbuild_remote_apis//build/bazel/remote/execution/v2:go_default_library
 gazelle(
     name = "gazelle",
+    visibility = ["//tools/tidy:__pkg__"],
 )
```
EdSchouten, line 10: Would it make sense to now just move these into //tools/tidy?

tools/tidy/main.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+		if !info.IsDir() && strings.HasSuffix(info.Name(), ".proto") {
+			run("external/llvm_toolchain/bin/clang-format", "-i", path)
+		}
+		return nil
+	}); err != nil {
+		log.Fatalf("walking %s failed: %v", root, err)
+	}
+	runGo("../org_golang_x_lint/golint", "golint", "-set_exit_status", root+"/...")
+	runGo("../com_github_gordonklaus_ineffassign", "ineffassign", root)
```
EdSchouten, line 89: It's annoying that both golint and ineffassign have such massive entry points. Otherwise these could have been some plain `import`s.

EdSchouten (commented):
Could you make sure that this also works for bb-remote-execution, etc. when dragged in through an external repo?

tools/tidy/main.go
asuffield: I'm increasingly tempted to cut them both in favour of golangci-lint anyway.

EdSchouten:

Issue #93, opened 2020-09-02
Failed to update object metadata with AWS S3


timuralp:
Following the change to unconditionally call `blob.Copy` in FindMissing, objects bigger than 5GB result in an error from AWS (https://github.com/buildbarn/bb-storage/pull/82). This is because they require a multipart upload and to update metadata on a multipart upload multiple calls have to be made:
1. initiate multipart upload
2. upload part copy with the same source
3. complete multipart upload with the new metadata

The limitation is documented here: https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html. Specifically:
> You create a copy of your object up to 5 GB in size in a single atomic operation using this API. However, to copy an object greater than 5 GB, you must use the multipart upload Upload Part - Copy API.

This also means that objects > 5GB will require multiple calls on every pass through FindMissing (since the x-amz-copy-source-if-unmodified-since has to be passed to every UploadPartCopy call -- https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html).

Lastly, this needs to be fixed in gocloud, but currently, bb-storage will not work as intended for objects > 5GB.

The error will appear as follows from AWS:
```
---[ REQUEST POST-SIGN ]-----------------------------
PUT /junk HTTP/1.1
Host: 1space-test.s3.amazonaws.com
User-Agent: aws-sdk-go/1.34.5 (go1.14.4; linux; amd64)
Content-Length: 0
Authorization: AWS4-HMAC-SHA256 Credential=<key>/20200902/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-copy-source;x-amz-copy-source-if-unmodified-since;x-amz-date;x-amz-meta-used;x-amz-metadata-directive, Signature=5f21ff6a41f0aae38c1a562d7d43f195deb0d572755fbc8ab93151147309fee4
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Copy-Source: <object>
X-Amz-Copy-Source-If-Unmodified-Since: Wed, 02 Sep 2020 16:39:02 GMT
X-Amz-Date: 20200902T163902Z
X-Amz-Meta-Used: 2020-09-02 16:39:02.003101746 +0000 UTC
X-Amz-Metadata-Directive: REPLACE
Accept-Encoding: gzip


-----------------------------------------------------
[...]

PR #98, opened 2020-09-08
Evict entries from existence cache when NotFound errors are seen

asuffield:
Storage backends signal that a cache item has been evicted by
returning NotFound. We can use this signal to clear the existence
cache immediately. Since bazel does retry FindMissing on each
execution attempt, this will result in the retry getting a correct
result and being able to proceed.

Where multiple parallel existence caches are used, --remote_retries will need to be larger than the number of caches for this to work.

Obviously this doesn't help in build-without-the-bytes mode if the lost cache entry is an execution result, but it should work for input files which bazel has available.

### DIFF
--- a/pkg/blobstore/existence_caching_blob_access.go
+++ b/pkg/blobstore/existence_caching_blob_access.go
@@ -3,7 +3,10 @@ package blobstore
 import (
 	"context"
 
+	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
 	"github.com/buildbarn/bb-storage/pkg/digest"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 type existenceCachingBlobAccess struct {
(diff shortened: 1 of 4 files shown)


EdSchouten:
Sorry, but I'm going to have to decline a change like this. First of all, this requirement is already documented:

```proto
// The storage backend must also be robust
// enough to guarantee that objects don't disappear shortly after
// calling ContentAddressableStorage.FindMissingBlobs(), as that
// would cause this decorator to cache invalid data.
```

Second, quoting #93:

> I'm willing to keep it in the tree, as long as it's not in the way in terms of development and its ugliness doesn't spill into the rest of the codebase.

I think this qualifies as ugliness spilling into the rest of the codebase.

I used to hate Bazel's requirements on blob retention as much as you, but at some point I've come to realise it's for the better. How can you expect a build cluster to build a 10 GB project, if your storage isn't capable of holding on to 10 GB of data in the first place? Bazel could surely be extended to do even more retrying, backtracking, but that's just silly. It blows up build times for what could be seen as the build cluster being misconfigured. If Bazel fails explicitly, you as an administrator know that there's an underlying problem that needs fixing.

I didn't write [ADR#2](https://github.com/buildbarn/bb-adrs/blob/master/0002-storage.md) as a hobby. I wrote it for people to use it as a starting point for building their own storage clusters that work well. Sure, there are things that could be improved to it:

- A MirroredBlobAccess that supports three or more backends (e.g., a tie-breaker), so that downtime is permitted.
- A version of LocalBlobAccess that supports persistency.

But at the same time, what's there already allows you to build some pretty awesome setup that can easily scale to petabytes in size. Changes like the ones you proposed are just workarounds that in my opinion don't have a place in this source tree.

As I stated before, I'm getting a bit tired of dealing with stuff like this. So far half of the time I have available to work with the Open Source community is being wasted on fruitless discussions on how to make S3 do something it wasn't designed to do in the first place. This PR has been the tipping point for me, which is why I've decided that I will remove CloudBlobAccess in the nearby future.

PR #100, opened 2020-09-15
enable golangci-lint

asuffield:
Let's see what this looks like if we turn on their github integration.


asuffield:
It's angry about the missing packages, so rebasing over https://github.com/buildbarn/bb-storage/pull/96

asuffield:
Annoyance: this isn't going to work until we can supply it with the generated .go files, which is going to require some creativity.

EdSchouten:
Hi Andrew,

Sorry for letting you wait. The last weeks have been *very* busy on my end.

Question: What are your thoughts on not making any code changes (i.e., checking bogus source files), but simply creating some temporary source files prior to running `go mod tidy`? That way we don't need to care about bad interactions with Gazelle, spurious golint warnings, etc.

asuffield:
> Hi Andrew,
> 
> Sorry for letting you wait. The last weeks have been _very_ busy on my end.
> 
> Question: What are your thoughts on not making any code changes (i.e., checking bogus source files), but simply creating some temporary source files prior to running `go mod tidy`? That way we don't need to care about bad interactions with Gazelle, spurious golint warnings, etc.

Interesting variation on the theme... I've also been a bit buried the past few weeks, but let me see if I can find time to pull together another attempt at this. It has occurred to me that we could probably take advantage of the rules_go go_path rule to have the temporary source files be the actual source files, which would be enough to get golangci-lint working as well as go mod tidy. A bit of creative wrapping will be needed but that might be a useful exercise anyway.

EdSchouten:

PR #102, opened 2020-11-26
build: upgrade rules_go and gazelle

tdyas:
Upgrade rules_go to v0.24.7 and gazelle to v0.22.2. The main change is that the generated `go_library` and other targets are now named after the package and do not default to `go_default_library`. The benefit is that it unlocks upgrading to Go v1.15.5. (`go_test` targets are now named `PACKAGE_test` where `PACKAGE` is the package name.)

Most of the changes were made by invoking `bazel run //:gazelle -- fix`. I then had to revert the changes to restore `@com_github_bazelbuild_remote_apis//build/bazel/remote/execution/v2:go_default_library` and `"@com_github_bazelbuild_remote_apis//build/bazel/semver:go_default_library"` because the new naming scheme does not seem to apply to them for a reason that I have not yet ascertained.


tdyas:
And CI fails because of the protobuf target naming convention issue (among other reasons). Will have to dig in a bit on why Gazelle wants to use the new naming convention but the protobuf targets for remote-apis want to use the `go_default_library` convention.

EdSchouten:
> Most of the changes were made by invoking `bazel run //:gazelle -- fix`. I then had to revert the changes to restore `@com_github_bazelbuild_remote_apis//build/bazel/remote/execution/v2:go_default_library` and `"@com_github_bazelbuild_remote_apis//build/bazel/semver:go_default_library"` because the new naming scheme does not seem to apply to them for a reason that I have not yet ascertained.

I guess it may make sense to just add a local patch for @com_github_bazelbuild_remote_apis to our repository. That way Gazelle will do the right thing out of the box.

> And CI fails because of the protobuf target naming convention issue (among other reasons). Will have to dig in a bit on why Gazelle wants to use the new naming convention but the protobuf targets for remote-apis want to use the `go_default_library` convention.

That's because unlike most other Go repositories, the remote-apis one has BUILD files with Go targets checked in. Those use the old naming scheme, thereby causing Gazelle to leave them alone.

EdSchouten:
This has happened in the meantime. We’re now using a recent version of Gazelle that uses the new naming scheme.

PR #106, opened 2021-01-28
allow configuration of trace sampling policy

tdyas:
Add a new `sampling_policy` item to the `global` configuration block to allow configuration of trace sampling to more than just the default and always.

The sampling policies in the proto come from the [`AlwaysSample`](https://github.com/census-instrumentation/opencensus-go/blob/13369a4547045c729dd938a203157a208afc244a/trace/sampling.go#L64), [`NeverSample`](https://github.com/census-instrumentation/opencensus-go/blob/13369a4547045c729dd938a203157a208afc244a/trace/sampling.go#L71) and [`ProbabilitySampler`](https://github.com/census-instrumentation/opencensus-go/blob/13369a4547045c729dd938a203157a208afc244a/trace/sampling.go#L43) Sampler implementations available in the Open Census Go library.


tdyas:
This also removes the `always_sample` as moot now that there is `sampling_policy`.

pkg/global/BUILD.bazel
```diff
@@ -19,5 +19,7 @@ go_library(
         "@io_opencensus_go_contrib_exporter_jaeger//:jaeger",
         "@io_opencensus_go_contrib_exporter_prometheus//:prometheus",
         "@io_opencensus_go_contrib_exporter_stackdriver//:stackdriver",
+        "@org_golang_google_grpc//codes",
```
EdSchouten, line 22: I just saw you pushed an additional commit to move these dependencies to the right place in the BUILD file. Life pro tip: you don't need to modify BUILD files by hand for changes like this. Just run:
```
bazel run //:gazelle
```
And it will add the missing dependencies for you!

pkg/global/apply_configuration.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/client_golang/prometheus/push"
+	"google.golang.org/grpc/codes"
```
EdSchouten, line 21: Nit: Add an empty line above this. The current convention in the Buildbarn codebase is to group imports by domain. github.com != golang.org.

pkg/proto/configuration/global/global.proto
```diff
@@ -68,6 +88,9 @@ message TracingConfiguration {
 
   // Whether or not all traces should be sampled.
   bool always_sample = 4;
+
+  // Policy to use to decide to sample traces.
+  SamplingPolicy sampling_policy = 5;
```
EdSchouten, line 93: As it is right now, the schema adds some ambiguity. There is no requirement that a `oneof` is set, meaning that `samplingPolicy: {}` is also valid, schema-wise. It means there is a SamplingPolicy set, but that none of the children are set. What are your thoughts on simply doing something like this?

```proto
message TracingConfiguration {
  ...
  oneof sample_policy {
    google.protobuf.Empty sample_always = 5;
    google.protobuf.Empty sample_never = 6;
    double sample_probability = 7;
  }
}
```

EdSchouten:
cc @asuffield

asuffield:
Ah yes, this is covered by https://github.com/buildbarn/bb-storage/pull/88 - we should probably just finish that one off.

pkg/proto/configuration/global/global.proto
tdyas: That suggestion is fine. Will switch it.

pkg/global/apply_configuration.go
tdyas: Will fix.

pkg/global/BUILD.bazel
tdyas: Nice. For these sorts of commands, probably worth documenting them in the README? I am more often in projects with the regular Go toolchain, so I always end up forgetting Bazel invocations like this and have to relearn every time ...

PR #107, opened 2021-02-08
tracing: add constant rate sampler

tdyas:
Add a constant rate trace sampler using on a token bucket algorithm. This is useful in lieu of probability-based trace sampler when the number of requests is low.


pkg/util/tracing.go
```diff
@@ -0,0 +1,96 @@
...
+	if s.AvailableTokens >= s.TokensPerSample {
+		s.AvailableTokens -= s.TokensPerSample
+		return trace.SamplingDecision{Sample: true}
+	}
+
+	return trace.SamplingDecision{Sample: false}
+}
+
+// Return a new constant rate trace sampler. The period exposed to callers is to one second.
+func NewConstantRateTraceSampler(tokensPerSecond, maxTokens, tokensPerSample int64) *ConstantRateTraceSampler {
```
EdSchouten, line 81: Just let this function return `trace.Sampler`. That way `ConstantRateTraceSampler` may remain private.

pkg/util/tracing.go
```diff
@@ -0,0 +1,96 @@
...
+// ConstantRateTraceSampler uses a token bucket algorithm to sample traces. This is useful as an alternative
+// to probability-based sampling when the service has a low number of requests per second.
+//
+// Inspired in part by https://github.com/jaegertracing/jaeger-client-go/blob/master/utils/rate_limiter.go
+
+type ConstantRateTraceSampler struct {
+	lock sync.Mutex
+
+	// The number of available tokens in the bucket.
+	AvailableTokens int64
```
EdSchouten, line 19: I think that all of these fields may remain private.

I would also suggest to group these fields by locking requirement. The general style in the Buildbarn codebase is to put constant fields at the very top of the struct, followed by placing all fields that require locking right underneath the lock declaration. That way it becomes self-documenting what the locking pattern is.

pkg/util/tracing.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Number of tokens put into the bucket per time period.
+	TokensPerPeriod int64
+
+	// Number of tokens required to sample a trace.
+	TokensPerSample int64
+
+	LastRefillTime time.Time
+	NextRefillTime time.Time
+	ClockNow       func() time.Time
```
EdSchouten, line 35: Be sure to use this interface instead:
https://pkg.go.dev/github.com/buildbarn/bb-storage@v0.0.0-20210207101039-9507e33a5caf/pkg/clock#Clock
We already create mocks for it in internal/mock, so inside your tests you can just say
```go
clock := mock.NewMockClock(ctrl)
clock.EXPECT().Now().Return(time.Unix(1000, 0))
```

pkg/util/tracing.go
```diff
@@ -0,0 +1,96 @@
+package util
```
EdSchouten, line 1: Considering that we can reasonably expect that this will only ever be used by pkg/global, I would suggest that this code lives there: `pkg/global/constant_rate_trace_sampler{,_test}.go` or something.

EdSchouten (commented):
Not saying we shouldn't merge a change like this, but have you considered getting something like this added upstream?

How does this fit in with the entire migration from OpenCensus to OpenTelemetry? It would be a shame if we invested a lot in adding features to this code if we're going to switch to OpenTelemetry in the long run.

pkg/util/tracing.go
EdSchouten: Ah! I see that your concern is that `trace.Sampler` is not an interface, but a plain callback type. No worries, you can either do this:
```go
func NewConstantRateTraceSampler(tokensPerSecond, maxTokens, tokensPerSample int64) trace.Sampler {
    return func(trace.SamplingParameters) trace.SamplingDecision {
       ...
    }
}
```
Or let this function return `s.sample`.

tdyas:
> Not saying we shouldn't merge a change like this, but have you considered getting something like this added upstream?

I didn't really. My primary concern for now is being able to get tracing spans from bb-storage in a low traffic situation. Upstreaming to the ultimate project is a "nice-to-have" given that goal.

> How does this fit in with the entire migration from OpenCensus to OpenTelemetry? It would be a shame if we invested a lot in adding features to this code if we're going to switch to OpenTelemetry in the long run.

My research into prior art only turned up the [rate-limited sampler in the Jaeger client library](https://github.com/jaegertracing/jaeger-client-go/blob/fe3fa553c313b32f58cc684a59a4d48f03e07df9/sampler.go#L177-L239). OpenTelemetry didn't look like it had a rate-limited sampler. See https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/sampling.go. (The [rate limiting code](https://github.com/jaegertracing/jaeger-client-go/blob/master/utils/rate_limiter.go) from Jaeger inspired part of this code, although I diverged to use `int64` vs `float64` to avoid any need for floating point operations.)

Thus, it appears that a constant rate sampler would still need to be developed for OpenTelemetry.

pkg/util/tracing.go
tdyas: > I think that all of these fields may remain private.

These fields were private in an earlier iteration of this change (in a version before the one seen in this PR). I switched the fields to public as, because the test is in the `util_test` package and this code is in the `util` package, the test originally constructed the `constantRateTraceSampler` struct directly in order to control the clock mock. It had to do this because the code originally returned the `sample` method (and not the struct).

I then switched to returning the struct and not the `Sample` method from the construction method so that the test could avoid needing to construct the struct directly, but that in turn still required the field to be public for the test to be able to control the clock mock.

So the weird structure of this code is due to mocking the clock for the test. Your comment to use `NewMockClock` seems like it could potentially remove the need for these "visibility gymnastics."

pkg/util/tracing.go
tdyas: Will do. That might obviate the need for some of the field visibility issues.

pkg/util/tracing.go
tdyas: Will do.

pkg/util/tracing.go
tdyas: See https://github.com/buildbarn/bb-storage/pull/107#discussion_r572412082. This code originally did just return the `sample` method, but had to be changed given the way I tried to mock the clock and how this code puts tests into a different package from the code under test (which is not idiomatic Go).

pkg/util/tracing.go
tdyas: > I would also suggest to group these fields by locking requirement. The general style in the Buildbarn codebase is to put constant fields at the very top of the struct, followed by placing all fields that require locking right underneath the lock declaration. That way it becomes self-documenting what the locking pattern is.

Will do.

tdyas:
Rebased the PR and updated based on the comments.

pkg/global/constant_rate_trace_sampler.go
```diff
@@ -0,0 +1,86 @@
...
+		return a
+	}
+	return b
+}
+
+// Sample decides whether to sample a trace. The trace.Sampler type is unfortunately a function and
+// not an interface. This `Sample` function should be provided to the OpenCensus tracing library as
+// the `DefaultSampler`, and not the ConstantRateTraceSampler struct.
+func (s *ConstantRateTraceSampler) sample(params trace.SamplingParameters) trace.SamplingDecision {
+	if params.ParentContext.IsSampled() {
```
EdSchouten, line 38: Does this mean that if there's a process in front of the cluster (e.g., an on-prem instance of bb-storage for caching) that reports that a part of the trace is sampled, we always sample without respecting the maximum token count? Is this desirable? May it be abused?

pkg/global/constant_rate_trace_sampler.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/buildbarn/bb-storage/pkg/clock"
+
+	"go.opencensus.io/trace"
+)
+
+// ConstantRateTraceSampler uses a token bucket algorithm to sample traces. This is useful as an alternative
```
EdSchouten, line 12: Please wrap comments at <80 columns.

pkg/proto/configuration/global/global.proto
```diff
@@ -57,6 +57,18 @@ message PrometheusPushgatewayConfiguration {
...
+// Parameters for a "token bucket" for deciding when to sample.
+message ConstantRateSamplingPolicy {
+  // Number of tokens to put in the bucket per second.
+  int32 tokens_per_second = 1;
+
+  // Maximum number of tokens that may be held in the bucket.
+  int32 max_tokens = 2;
+
+  // Number of tokens required to consume from the bucket to sample a trace.
+  int32 tokens_per_sample = 3;
```
EdSchouten, line 69: I guess this was added to support some form of fractions. Is that really a realistic use case? My intuition says that a schema that's more like this would have been adequate for most users:
```
message ConstantRateSamplingPolicy {
  int64 samples_per_period = 1;
  google.protobuf.Duration period = 2;
}
```

pkg/global/constant_rate_trace_sampler.go
tdyas: Will do.

I'll make the manual edit for now, but would love to know if there is any automated tooling for Go that can do this consistently.

tdyas: and if so, maybe incorporate into the CI flow?

pkg/proto/configuration/global/global.proto
tdyas: Yes, for example the ability to do 1 sample every 2 or 10 seconds. Configuring the period would be an alternate way to accomplish that. Will switch to that. I definitely overcomplicated that.

pkg/global/constant_rate_trace_sampler.go
tdyas: > Is this desirable? 

In my view, that question is one for the user. One risk is incomplete traces across the entire life cycle of a request because only certain services choose to log their spans. The benefit is getting complete traces through the system more consistently than if each service made the decision in isolation randomly or at a low rate.

The existing probability-based sampler respects the decision. See https://github.com/census-instrumentation/opencensus-go/blob/49838f207d61097fc0ebb8aeef306913388376ca/trace/sampling.go#L52-L54. 

A more nuanced configuration would be to have a separate sampler that only checks whether the upstream made a trace decision. OpenCensus unfortunately does not have such a sampler; OpenTelemetry does. See https://github.com/open-telemetry/opentelemetry-go/blob/875a25835fcfcddd61e94d87b1a0b14db4d5a0e5/sdk/trace/sampling.go#L157-L162. 

So maybe we should make the parent-sampling part of this PR configurable? The better tracing architecture in my view is to let the user decide what the questions of when to sample at all and when to respect the upstream as separate decisions.

I made the decision here for consistency with the probability-based sampler in OpenCensus, but that is not my preferred choice if consistency is not a requirement.

> May it be abused?

Like anything else, yes. A malicious client could send a `grpc-trace-bin` and now the request will have a tracing context if trace header propagation is enabled if it always respects the parent decision. I've seen multi-service traces across multiple BuildBarn services for a request in the tracing store I'm using, so I suspect something might be enabling that, but I still need to figure out what since I want to disable it if it is enabled on the bb-storage fronting our cluster. (I still see missing spans but that may be occurring for other reasons.)

Do you have any additional context on how BuildBarn's grpc library is setup or not for tracing? 

Any way, I would say it is definitely good practice for any entry point to ignore the `grpc-trace-bin` header and originate the tracing decision itself if it cannot trust its clients.

pkg/global/constant_rate_trace_sampler.go
```diff
@@ -0,0 +1,86 @@
...
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	// Update the number of available samples based on the time since the last check.
+	now := s.clock.Now()
+	if now.Equal(s.nextRefillTime) || now.After(s.nextRefillTime) {
+		durationSinceLastRefillTime := now.Sub(s.lastRefillTime)
+		numPeriodsSinceLastRefillTime := max(0, durationSinceLastRefillTime.Nanoseconds()/s.periodDurationNanos)
+
+		s.availableTokens += numPeriodsSinceLastRefillTime * s.tokensPerPeriod
```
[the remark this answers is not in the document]
tdyas: The concern here is whether to allow some measure of bursting if there are lower amounts of traffic. `maxTokens` lets the user make that choice for themself. I personally wouldn't set `maxTokens` higher than 2 or 3 times the tokens per period number. I'm not worried about flooding my tracing provider if the system will burst with 3 seconds worth of tracing in the morning. Of course, a user could make a mistake here and set maxTokens to a really large number. How much is it worth protecting that user from their own ignorance of the tracing policy versus giving up burst capability?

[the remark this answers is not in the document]
tdyas: Also, my main goal of this sampler is to have a sampler for low traffic situations where the probability-based sampler is ineffective at generating a sufficient number of traces to be useful. At higher traffic levels, I'd probably switch to the probability-based sampler.

pkg/global/constant_rate_trace_sampler.go
```diff
@@ -0,0 +1,86 @@
...
+	if params.ParentContext.IsSampled() {
+		return trace.SamplingDecision{Sample: true}
+	}
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	// Update the number of available samples based on the time since the last check.
+	now := s.clock.Now()
+	if now.Equal(s.nextRefillTime) || now.After(s.nextRefillTime) {
```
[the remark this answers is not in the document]
tdyas: Yup that would work as well.

pkg/global/constant_rate_trace_sampler.go
```diff
@@ -0,0 +1,86 @@
...
+
+func max(a, b int64) int64 {
+	if a > b {
+		return a
+	}
+	return b
+}
+
+// Sample decides whether to sample a trace. The trace.Sampler type is unfortunately a function and
+// not an interface. This `Sample` function should be provided to the OpenCensus tracing library as
```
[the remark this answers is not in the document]
tdyas: Just my opinion on API design. Will remove the commentary.

pkg/global/constant_rate_trace_sampler.go
```diff
@@ -0,0 +1,86 @@
...
+// NewConstantRateTraceSampler returns a new constant rate trace sampler.
+// The period exposed to callers is one second.
+func NewConstantRateTraceSampler(tokensPerSecond, maxTokens, tokensPerSample int64, clock clock.Clock) trace.Sampler {
+	now := clock.Now()
+
+	s := &ConstantRateTraceSampler{
+		clock:               clock,
+		availableTokens:     maxTokens,
+		maxTokens:           maxTokens,
+		periodDurationNanos: int64(time.Second),
```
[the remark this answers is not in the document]
tdyas: I'm trying to recall why I did this, but I've forgotten the reason unfortunately.

pkg/global/constant_rate_trace_sampler.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	// Update the number of available samples based on the time since the last check.
+	now := s.clock.Now()
+	if now.Equal(s.nextRefillTime) || now.After(s.nextRefillTime) {
+		durationSinceLastRefillTime := now.Sub(s.lastRefillTime)
+		numPeriodsSinceLastRefillTime := max(0, durationSinceLastRefillTime.Nanoseconds()/s.periodDurationNanos)
```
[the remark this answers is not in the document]
tdyas: >  As far as I know, Go's time.Time values embed a monotonic clock value as well, meaning that calls to time.Now() always return timestamps for which expressions like the one you wrote are guaranteed to be non-negative.

I wasn't aware of that. Good to know.

pkg/global/constant_rate_trace_sampler.go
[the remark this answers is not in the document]
tdyas: As for why "unfortunate," maybe not in this case, but in the more general sense, I find it complicates the ability of test code that needs to be able to control parts of the underlying struct (which I originally thought had been needed for this). Returning the function gives up that possibility since the struct is no longer accessible to anything outside the function including test code.

(Of course, that would only be feasible any way in cases where the code and tests are in the same package which is not the case with this code base.)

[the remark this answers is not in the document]
EdSchouten: I think that if it's necessary for tests to modify the struct, then it's more of an indicator that decomposition of your type is needed.

PR #113, opened 2021-03-16
Add local blob storage comment to proto file

charlesoconor:
This should make it easier to find. Pulled from https://github.com/buildbarn/bb-storage/blob/ea0ae7cb574af99b1aac53b6b4d547b0a2262f16/pkg/blobstore/local/old_current_new_location_blob_map.go#L34

I changed some wording in the first paragraph to make more sense for the proto. Otherwise should be unchanged.

### DIFF
--- a/pkg/proto/configuration/blobstore/blobstore.proto
+++ b/pkg/proto/configuration/blobstore/blobstore.proto
@@ -308,6 +308,70 @@ message MirroredBlobAccessConfiguration {
   BlobReplicatorConfiguration replicator_b_to_a = 4;
 }
 
+// LocalBlobAccess stores all data onto disk in block sizes. A block
+// cannot span multiple blocks, meaning that blocks generally need to
+// be large in size (gigabytes). The number of blocks may be relatively
+// low. For example, for a 512 GiB cache, it is acceptable to create 32
+// blocks of 16 GiB in size.
+//
+// Blocks are partitioned into three groups based on their creation
+// time, named "old", "current" and "new". Blobs provided to Put() will
+// always be stored in a block in the "new" group. When the oldest block
+// in the "new" group becomes full, it is moved to the "current" group.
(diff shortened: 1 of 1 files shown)


pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -308,6 +308,70 @@ message MirroredBlobAccessConfiguration {
...
+//
+// Blocks are partitioned into three groups based on their creation
+// time, named "old", "current" and "new". Blobs provided to Put() will
+// always be stored in a block in the "new" group. When the oldest block
+// in the "new" group becomes full, it is moved to the "current" group.
+// This causes the oldest block in the "current" group to be displaced
+// to the "old" group. The oldest block in the "old" group is discarded.
+//
+// The difference between the "current" group and the "old" group is
+// that the needRefresh value returned by Get() differs.
```
EdSchouten, line 325: I think that for OldCurrentNewLocationBlobmap this phrasing makes sense, because we want to document how it interacts with other classes. For blobstore.proto I think it would make more sense to just explain the implications of this from a user's perspective:

```
// The difference between the "current" group and the "old" group is
// that data is refreshed when accessed.
```

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -308,6 +308,70 @@ message MirroredBlobAccessConfiguration {
...
+// Data in the "old" group is at risk of being removed in the nearby
+// future, which is why it needs to be copied into the "new" group when
+// requested to be retained. Data in the "current" group is assumed to
+// remain present for the time being, which is why it is left in place.
+// This copying is performed by KeyBlobMapBackedBlobAccess.
```
EdSchouten, line 330: I think we can also remove this sentence for the same reason.

EdSchouten (approved):

PR #115, opened 2021-03-29
Add configurable window size

jlaxson:
In my struggle to understand/optimize cache performance, I added these as configuration options to pass through to grpc.  I'm not sure if they make a difference or not, or if anyone else would use them, but cost in code is slight and will at least document here for anyone else troubleshooting.  Take or leave at your pleasure.

See also:
* https://github.com/jlaxson/bb-storage/pull/new/add-grpc-window-size
* https://pkg.go.dev/google.golang.org/grpc#WithInitialConnWindowSize and friends


jlaxson:
Whew that was easy :p

PR #116, opened 2021-03-31
Add Clonefile implementation to directory interface

jlaxson:
This is to support using Clonefile in the remote-executor repo

### DIFF
--- a/pkg/filesystem/directory.go
+++ b/pkg/filesystem/directory.go
@@ -74,7 +74,7 @@ type Directory interface {
 
 	// Link is the equivalent of os.Link().
 	Link(oldName path.Component, newDirectory Directory, newName path.Component) error
-	// Clonefile is the equivalent of unix.CloneFile on macOS
+	// Clonefile is the equivalent of unix.Clonefile on macOS.
 	Clonefile(oldName path.Component, newDirectory Directory, newName path.Component) error
 	// Lstat is the equivalent of os.Lstat().
 	Lstat(name path.Component) (FileInfo, error)
--- a/pkg/filesystem/directory.go
+++ b/pkg/filesystem/directory.go
@@ -74,6 +74,8 @@ type Directory interface {
 
 	// Link is the equivalent of os.Link().
 	Link(oldName path.Component, newDirectory Directory, newName path.Component) error
+	// Clonefile is the equivalent of unix.CloneFile on macOS
+	Clonefile(oldName path.Component, newDirectory Directory, newName path.Component) error
 	// Lstat is the equivalent of os.Lstat().
 	Lstat(name path.Component) (FileInfo, error)
 	// Mkdir is the equivalent of os.Mkdir().
--- a/pkg/filesystem/local_directory_darwin.go
+++ b/pkg/filesystem/local_directory_darwin.go
@@ -6,8 +6,8 @@ import (
 	"os"
 
 	"github.com/buildbarn/bb-storage/pkg/filesystem/path"
-	"golang.org/x/sys/unix"
 
+	"golang.org/x/sys/unix"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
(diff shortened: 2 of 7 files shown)


pkg/filesystem/local_directory_unix.go
```diff
@@ -106,6 +106,15 @@ func (d *localDirectory) Link(oldName path.Component, newDirectory Directory, ne
 	})
 }
 
+func (d *localDirectory) Clonefile(oldName path.Component, newDirectory Directory, newName path.Component) error {
+	defer runtime.KeepAlive(d)
```
jlaxson, line 110: Don't know what this does, just copying the pattern - can you educate me?

pkg/filesystem/local_directory_darwin.go
```diff
@@ -17,3 +18,7 @@ type deviceNumber = int32
 func (d *localDirectory) Mknod(name path.Component, perm os.FileMode, dev int) error {
 	return status.Error(codes.Unimplemented, "Creation of device nodes is not supported on Darwin")
 }
+
+func clonefileImpl(oldFd int, oldName string, newFd int, newName string) error {
+	return unix.Clonefileat(oldFd, oldName, newFd, newName, 0)
```
EdSchouten, line 23: Please use `CLONE_NOFOLLOW` for the flags. This API isn't supposed to follow symlinks at all, as that would allow you to escape the directory.

pkg/filesystem/local_directory_darwin.go
```diff
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/buildbarn/bb-storage/pkg/filesystem/path"
+	"golang.org/x/sys/unix"
```
EdSchouten, line 9: Imports are grouped by second level domains in the Buildbarn tree. Please move this import next to the google.golang.org ones.

pkg/filesystem/local_directory_unix.go
EdSchouten: To make sure that file descriptors don't leak when code forgets to call Close(), we have a finalizer set up. This means the Go garbage collector will call Close for us()... at some point.

This specific piece of code unboxes `d` and passes the containing file descriptor to `newDirectory.Apply()`. What happens if the garbage collector is run:
- after extracting `d.fd`, and
- before finishing the call to `newDirectory.Apply()`?

Answer: Go may see `d` is no longer live, meaning it closes `d.fd`. By putting the `runtime.Keepalive()` call at the top (which is just a dummy function that Go's garbage collector treats like a black box), we can ensure the finalizer isn't run before this function completes.

pkg/filesystem/directory.go
```diff
@@ -74,6 +74,8 @@ type Directory interface {
 
 	// Link is the equivalent of os.Link().
 	Link(oldName path.Component, newDirectory Directory, newName path.Component) error
+	// Clonefile is the equivalent of unix.CloneFile on macOS
```
EdSchouten, line 77: s/CloneFile/Clonefile/

Also make sure to end comments with a period, as the rest of the code base does this fairly consistently.

EdSchouten (approved):
This looks almost good to land. Thanks!

pkg/filesystem/local_directory_darwin.go
```diff
@@ -17,3 +18,7 @@ type deviceNumber = int32
 func (d *localDirectory) Mknod(name path.Component, perm os.FileMode, dev int) error {
 	return status.Error(codes.Unimplemented, "Creation of device nodes is not supported on Darwin")
 }
+
+func clonefileImpl(oldFd int, oldName string, newFd int, newName string) error {
```
EdSchouten, line 22: s/oldFd/oldFD/; same for newFd.

Go's style guide suggests that abbreviations (HTTP, TCP, TLS) are always capitalized. We do it consistently in the Buildbarn tree, except when dealing with generated Protobuf field names.

pkg/filesystem/local_directory_unix.go
jlaxson: Interesting, thanks 👍

jlaxson:
> Sorry for bothering you once more; could you maybe add some basic testing coverage for this to local_directory_test.go? Just testing the success case and ENOENT on macOS is sufficient.

Done.  By the way, TestLocalDirectoryIsWritable and TestLocalDirectoryIsWritableChild don't pass for me on Darwin.  I notice they're skipped in CI, maybe something has gotten out of whack?

EdSchouten:
> Done. By the way, TestLocalDirectoryIsWritable and TestLocalDirectoryIsWritableChild don't pass for me on Darwin. I notice they're skipped in CI, maybe something has gotten out of whack?

They don't pass in case actions run as root, or at least having root-like permissions on disk. GitHub Workflows seems to run  stuff that way, which is why I've disabled these tests. When I just run `bazel test` on my Mac, the tests do pass. How did you run these tests?

PR #117, opened 2021-04-16
Update remote-apis patch to fix asset buildfile

jlaxson:


EdSchouten:
Is this change needed for bb-asset-hub? If so, my suggestion is to have that patch over there.

EdSchouten:
Or wait, never mind. We also rename this target in the same patch. LGTM!

PR #122, opened 2021-08-21
Clarify the BlobAccessConfiguration.remote field comment

mostynb:
The previous phrasing was ambiguous, since bazel supports two different cache protocols (HTTP and gRPC) and this field only
supports one of them (the term "HTTP caching protocol" is used in bazel's docs).

https://docs.bazel.build/versions/main/remote-caching.html#http-caching-protocol


EdSchouten:
Hey! Thanks for the patch! I used the occasion to also just rename this backend to HTTPBlobAccess, so it's more explicit.

PR #123, opened 2021-08-27
Fix infinite loop in authorizingBuildQueue.WaitExecution

mostynb:


EdSchouten (approved):

EdSchouten:
Great find, @mostynb. Thanks for the fix! I just added some unit testing coverage there as well to make sure it'll never regress.

PR #124, opened 2021-09-01
bolt on spire-based TLS auth when SPIFFE_ENDPOINT_SOCKET is in enviro…

srago:
…nment

First cut at using spire agent to get TLS auth creds.


srago:
Wrong repo, sorry

PR #125, opened 2021-10-11
add spifffe support

sevki:


sevki:
hey @EdSchouten this is still missing more docs and tests, but I'd use some early feedback about the direction :smile:

sevki:
shoot, it seems I don't know how to rebase from upstream.

pkg/grpc/authenticator.go
```diff
@@ -46,7 +46,9 @@ func NewAuthenticatorFromConfiguration(policy *configuration.AuthenticationPolic
 		}
 		return NewTLSClientCertificateAuthenticator(
 			clientCAs,
-			clock.SystemClock), nil
+			clock.SystemClock,
```
sevki, line 49: add roughtime support

pkg/proto/configuration/auth/auth.proto
```diff
@@ -19,6 +20,9 @@ message AuthorizerConfiguration {
 
     // Deny all requests.
     google.protobuf.Empty deny = 3;
+
+    // Spiffee
```
sevki, line 24: this is an example of horrible documentation. I''ll improve on this

pkg/proto/configuration/auth/auth.proto
```diff
@@ -27,3 +31,8 @@ message InstanceNameAuthorizer {
   // The empty string may be used to indicate all instance names.
   repeated string allowed_instance_name_prefixes = 1;
 }
+
+message SpiffeAuthorizer {
```
sevki, line 35: same docs problem

pkg/grpc/authenticator.go
EdSchouten: The main goal of the Clock interface right now is just to stub out the clock as part of unit tests. Well, there's also the I/O time compensating clock we use in bb_worker to compensate for delays caused by FUSE, but that's just a detail.

I think that stuff like roughtime is best implemented by the operating system itself, through some timekeeping daemon, right? Pretty exciting stuff. I hadn't heard of it before. Thanks for sharing!

pkg/proto/configuration/auth/auth.proto
```diff
@@ -27,3 +31,20 @@ message InstanceNameAuthorizer {
...
+//  {
+// 	spiffe:{
+// 		instanceNameSubjectMap:{
+// 			"foo":{
+// 				allowedSpiffeIds:{"example.com":"workload-1"}
+// 			}
+// 		}
+// 	}
+// }
+message SpiffeAuthorizer {
```
EdSchouten, line 47: SPIFFE is supposed to be an abbreviation, right? If so, could we call this SPIFFEAuthorizer? The same applies to any of the associated types in Go code.

pkg/proto/configuration/grpc/grpc.proto
```diff
@@ -201,4 +202,6 @@ message TLSClientCertificateAuthenticationPolicy {
   // PEM data for the certificate authorities that should be used to
   // validate the remote TLS client.
   string client_certificate_authorities = 1;
+  // Spiffe
+  buildbarn.configuration.spiffe.SubjectMatcher spiffe = 10;
```
EdSchouten, line 206: Buildbarn's story around authentication & authorization is far from perfect (mainly due to historical reasons). That said, do we think it actually makes sense to do subject matching as part of the authentication code? Couldn't we argue it's purely an authorization problem?

pkg/auth/spiffe_authorizer.go
```diff
@@ -0,0 +1,90 @@
...
+		return fillErrors(status.Error(codes.Unauthenticated, "Connection was not established using gRPC"))
+	}
+	tlsInfo, ok := p.AuthInfo.(credentials.TLSInfo)
+	if !ok {
+		return fillErrors(status.Error(codes.Unauthenticated, "Connection was not established using TLS"))
+	}
+	certs := tlsInfo.State.PeerCertificates
+	if len(certs) == 0 {
+		return fillErrors(status.Error(codes.Unauthenticated, "Client provided no TLS client certificate"))
+	}
```
EdSchouten, line 57: I think what we should aim for is to keep the implementations of Authorizer decoupled from the network protocols used. For example, if we ever need to implement certain parts of Buildbarn using different protocols (REST calls? GraphQL? Thrift?), it would be nice if we could still use the existing authorization framework. This is the reason why Authenticator is placed in pkg/grpc and Authorizer in pkg/auth.

My recommendation would be to let TLSClientCertificateAuthenticator wrap `ctx` upon success using `context.WithValue()`, attaching some kind of custom struct to it that contains the X.509 certificate. This Authenticator could then extract it without depending on any gRPC code.

pkg/auth/spiffe_authorizer.go
```diff
@@ -0,0 +1,90 @@
...
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+)
+
+// SpiffeAuthorizer  authorizes based on Spiffe-IDs
+// https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE-ID.md
+//
+// see spiffe.proto for configuration options
+type SpiffeAuthorizer struct {
```
EdSchouten, line 21: Be sure to make this type private, as NewSpiffeAuthorizer() can just return the interface type. This means that you don't need to document the struct and the Authorize() method. You can instead document NewSpiffeAuthorizer() (note: NewSPIFFEAuthorizer()) in lots of detail. This will then look pretty elegant if you go to pkg.go.dev/ to read the docs.

pkg/auth/spiffe_authorizer.go
```diff
@@ -0,0 +1,90 @@
...
+			continue
+		}
+		subjectMatchers, ok := instanceMatcher.AllowedSpiffeIds[id.TrustDomain().String()]
+		if !ok {
+			allowedTrustDomains := []string{}
+			for k := range instanceMatcher.AllowedSpiffeIds {
+				allowedTrustDomains = append(allowedTrustDomains, k)
+			}
+			errs[i] = status.Error(codes.PermissionDenied,
+				fmt.Sprintf("trust domain not availavle. availabe trust domains are %s", allowedTrustDomains))
```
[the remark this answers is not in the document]
sevki: great catch! I think I deserve some type of commendation from oxford dictionary for misspelling the same word wrong twice in the same sentence :rofl:

EdSchouten:
Hi @sevki @srago! This PR seems to be mostly inactive for the last nine months. Let's close this. Be sure to reopen/file a new PR if you wish to get this functionality added. Thanks again for your contribution!

srago:
Thanks for the head's up Ed.  I'll be rebasing and sending you a new PR
hopefully next month.  I want to make sure the code we submit is working on
the top of your tree first.

Steve

PR #126, opened 2021-10-12
Windows Port for local_directory

mou-hao:
Hi,

This change intends to port `local_directory` to Windows and is part of buildbarn/bb-remote-execution#7. I want to make sure this is the right approach before I go ahead and implement a Windows worker.

I saw some previous PRs related to this issue and understand there is a need for a file descriptor based approach to avoid race conditions and to have multiple conceptual cwd's, so in my port, `local_directory` is totally windows file handle based.

The implementation passes all existing tests (except for `IsWritable` and `IsWritableChild` because of the very different file access control model).

There are some caveats though:
- To support POSIX-like rename and delete, this implementation uses features only available on Windows 10 / Windows Server 2016 or above.
- This assumes NTFS, which I would say is a safe assumption on Windows.
- If you are creating a relative symlink to a target that is not a child/grandchild/etc. of the local directory, then race condition is possible, the reason being you need to know whether the target is a directory when creating symlink on Windows.
- You need to have Developer Mode on or run as administrator to be able to create symlinks on Windows.

Cheers,
mh

I am requested by my employer to include the following disclaimer here:

> THIS SOFTWARE IS CONTRIBUTED SUBJECT TO THE TERMS OF THE APACHE LICENSE V.2.0. YOU MAY OBTAIN A COPY OF THE LICENSE AT https://www.apache.org/licenses/LICENSE-2.0.
THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. THIS SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING THIS OR ANOTHER EQUIVALENT DISCLAIMER IN ADDITION TO ANY OTHER REQUIRED LICENSE TERMS.

### DIFF
--- a/pkg/filesystem/local_directory_test.go
+++ b/pkg/filesystem/local_directory_test.go
@@ -3,7 +3,6 @@ package filesystem_test
 import (
 	"io"
 	"os"
-	"path/filepath"
 	"runtime"
 	"syscall"
 	"testing"
@@ -244,7 +243,7 @@ func TestLocalDirectoryReadlinkSuccess(t *testing.T) {
 	require.NoError(t, d.Symlink("/foo/bar/baz", path.MustNewComponent("symlink")))
 	target, err := d.Readlink(path.MustNewComponent("symlink"))
 	require.NoError(t, err)
-	require.Equal(t, filepath.FromSlash("/foo/bar/baz"), target)
+	require.Equal(t, "/foo/bar/baz", target)
 	require.NoError(t, d.Close())
 }
 
--- a/pkg/filesystem/local_directory_windows.go
+++ b/pkg/filesystem/local_directory_windows.go
@@ -13,52 +13,11 @@ import (
 	"unsafe"
 
 	"github.com/buildbarn/bb-storage/pkg/filesystem/path"
+	"github.com/buildbarn/bb-storage/pkg/filesystem/windowsext"
 
 	"golang.org/x/sys/windows"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-)
-
-var (
-	modntdll = windows.NewLazySystemDLL("ntdll.dll")
-
-	procNtSetInformationFile = modntdll.NewProc("NtSetInformationFile")
-)
-
-func ntSetInformationFile(handle windows.Handle, iosb *windows.IO_STATUS_BLOCK, fileInfo *byte, len, class uint32) error {
-	r0, _, _ := syscall.Syscall6(procNtSetInformationFile.Addr(), 5, uintptr(handle),
-		uintptr(unsafe.Pointer(iosb)), uintptr(unsafe.Pointer(fileInfo)), uintptr(len), uintptr(class), 0)
-	if r0 != 0 {
-		return windows.NTStatus(r0)
-	}
-	return nil
-}
-
-const (
-	// these constants are not present in x/sys/windows
-	FILE_READ_ATTRIBUTES  = 0x80
-	FILE_READ_DATA        = 1
-	FILE_READ_EA          = 8
-	FILE_WRITE_ATTRIBUTES = 0x100
-	FILE_WRITE_DATA       = 2
-	FILE_WRITE_EA         = 0x10
-	FILE_GENERIC_READ     = windows.STANDARD_RIGHTS_READ | windows.SYNCHRONIZE | FILE_READ_ATTRIBUTES | FILE_READ_DATA | FILE_READ_EA
-	FILE_GENERIC_WRITE    = windows.STANDARD_RIGHTS_WRITE | windows.SYNCHRONIZE | FILE_WRITE_ATTRIBUTES | FILE_WRITE_DATA | FILE_WRITE_EA
-
-	SYMLINK_FLAG_RELATIVE        = 1
-	FSCTL_SET_REPARSE_POINT      = 0x900A4
-	FSCTL_QUERY_ALLOCATED_RANGES = 0x940CF
-	FSCTL_SET_SPARSE             = 0x900C4
-
-	FILE_RENAME_REPLACE_IF_EXISTS = 1
-	FILE_RENAME_POSIX_SEMANTICS   = 2
-
-	FILE_DISPOSITION_DELETE          = 1
-	FILE_DISPOSITION_POSIX_SEMANTICS = 2
-
-	FileRenameInformation        = 10
-	FileLinkInformation          = 11
-	FileDispositionInformationEx = 64
 )
 
 func ntCreateFile(handle *windows.Handle, access uint32, root windows.Handle, path string, disposition, options uint32) error {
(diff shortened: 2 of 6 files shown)


pkg/filesystem/local_directory_windows.go
```diff
@@ -0,0 +1,1105 @@
...
+	r0, _, _ := syscall.Syscall6(procNtSetInformationFile.Addr(), 5, uintptr(handle),
+		uintptr(unsafe.Pointer(iosb)), uintptr(unsafe.Pointer(fileInfo)), uintptr(len), uintptr(class), 0)
+	if r0 != 0 {
+		return windows.NTStatus(r0)
+	}
+	return nil
+}
+
+const (
+	// these constants are not present in x/sys/windows
```
EdSchouten, line 39: I'm fine with keeping stuff like this in the tree in the short term, but could you please also file a PR against golang.org/x/sys to get this stuff added and add a comment to link to it? Thanks!

pkg/filesystem/local_directory_windows.go
```diff
@@ -0,0 +1,1105 @@
...
+	})
+}
+
+func (d *localDirectory) Clonefile(oldName path.Component, newDirectory Directory, newName path.Component) error {
+	return status.Error(codes.Unimplemented, "Clonefile is only supported on Darwin")
+}
+
+var executableExts = [...]string{".COM",".EXE",".BAT",".CMD",".VBS",".VBE",".JS",".JSE",".WSF",".WSH",".MSC",".CPL"}
+
+func isExecutable(fileName string) bool {
```
EdSchouten, line 437: Another option would be to always return true on Windows, right? I've noticed that Bazel itself already makes almost everything in input roots executable, regardless of permissions on the host (even on Linux). It wouldn't really harm us if we did the same thing on the remote side.

pkg/filesystem/local_directory_windows.go
```diff
@@ -0,0 +1,1105 @@
...
+	FileAttributes  uint32
+	FileNameLength  uint32
+	EaSize          uint32
+	FileName        [1]uint16
+}
+
+func readdirnames(handle windows.Handle) ([]string, error) {
+	outBufferSize := uint32(512)
+	outBuffer := make([]byte, outBufferSize)
+	for true {
```
EdSchouten, line 523: Nit: `for {`
Also in a couple of other places.

pkg/filesystem/local_directory_windows.go
```diff
@@ -0,0 +1,1105 @@
...
+		case windows.STATUS_FILE_IS_A_DIRECTORY:
+			return syscall.EISDIR
+		default:
+			return ntstatus.(windows.NTStatus).Errno()
+		}
+	}
+	return nil
+}
+
+type FILE_ATTRIBUTE_TAG_INFO struct {
```
EdSchouten, line 97: Could we make types like these private? Or if the goal is to get all of these added to x/sys/windows at some point, move them into a separate package for the time being? pkg/filesystem/windowsext?

EdSchouten (commented):
Hey there!

Thanks for this awesome contribution. It looks like you've given all of this a lot of thought. It's nice to see that you even managed to implement the more exotic features like GetNextRegionOffset(). Well done!

I've only got a couple of small remarks. I've also just enable CI builds for you. Make sure that all style conformance and golint warnings are addressed and we're good to go.

pkg/filesystem/local_directory_test.go
```diff
@@ -234,7 +244,7 @@ func TestLocalDirectoryReadlinkSuccess(t *testing.T) {
 	require.NoError(t, d.Symlink("/foo/bar/baz", path.MustNewComponent("symlink")))
 	target, err := d.Readlink(path.MustNewComponent("symlink"))
 	require.NoError(t, err)
-	require.Equal(t, "/foo/bar/baz", target)
+	require.Equal(t, filepath.FromSlash("/foo/bar/baz"), target)
```
EdSchouten, line 247: I guess it should be the responsibility of LocalDirectory on Windows to call filepath.ToSlash() here. The reason being that the Remote Execution protocol requires us to capture symlink targets using / as a path separator:

https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto#L1233-L1238

pkg/filesystem/local_directory_windows.go
```diff
@@ -0,0 +1,1043 @@
...
+	err := ntCreateFile(&handle, windows.FILE_LIST_DIRECTORY, d.handle, name.String(),
+		windows.FILE_CREATE, windows.FILE_DIRECTORY_FILE|windows.FILE_OPEN_REPARSE_POINT)
+	return err
+}
+
+func (d *localDirectory) Mknod(name path.Component, perm os.FileMode, dev int) error {
+	return status.Error(codes.Unimplemented, "Creation of device nodes is not supported on Windows")
+}
+
+type fileFullDirInfo struct {
```
EdSchouten, line 443: I suppose structs like these should also be moved into windowsext, as those are also candidates for upstreaming into golang.org/x/sys, right?

mou-hao: I am more certain the consts and NtSetFileInformation will fit in x/sys, but not so sure about the structs. There are many win32 APIs that can take different struct types and an enum to indicate which type is passed (and sometimes another int for the size of the out buffer). In x/sys/windows, they have consts for the enums but use *byte for the struct passed in. I think it's sensible for them to do so because oftentimes the structs are really there just to tell you the offsets in a byte buffer and do not encapsulate, like when the last member of the struct is the first character of a C string and rest of the string should follow in memory..

But let me move them to windowsext, even if x/sys doesn't want them, it will look cleaner.

pkg/filesystem/local_directory_windows.go
mou-hao: I will start with creating PRs for adding the constants. This should be quick.

Adding NtSetFileInformation and the structs will likely need raising an issue and see what go dev thinks. I will do that as well.

PR #130, opened 2021-11-15
Update documentation for minimum_epoch_interval

jaronsummers:
google.protobuf.Duration's JSON behavior is not super intuitive so updating the example to be reflective of real usage will hopefully save someone some time

### DIFF
--- a/pkg/proto/configuration/blobstore/blobstore.proto
+++ b/pkg/proto/configuration/blobstore/blobstore.proto
@@ -576,12 +576,12 @@ message LocalBlobAccessConfiguration {
     // Care should be taken that this value is not set too low. Every
     // epoch that still references valid data consumes 16 bytes of
     // memory and increases the size of the state file by a similar
-    // amount. This means that if this option is set to 5m, epoch
+    // amount. This means that if this option is set to '300s', epoch
     // bookkeeping consumes up to 12*24*365*16 B = ~1.68 MB of space if
     // the system were to operate for a full year without blocks being
-    // released. Setting this to 1s would blow this up by a factor 300.
+    // released. Setting this to '1s' blows this up by a factor 300.
     //
-    // Recommended value: 5m
+    // Recommended value: '300s'
     google.protobuf.Duration minimum_epoch_interval = 2;
   }
 


pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -576,12 +576,14 @@ message LocalBlobAccessConfiguration {
     // Care should be taken that this value is not set too low. Every
     // epoch that still references valid data consumes 16 bytes of
     // memory and increases the size of the state file by a similar
-    // amount. This means that if this option is set to 5m, epoch
+    // amount. This means that if this option is set to "300s", epoch
     // bookkeeping consumes up to 12*24*365*16 B = ~1.68 MB of space if
     // the system were to operate for a full year without blocks being
-    // released. Setting this to 1s would blow this up by a factor 300.
+    // released. Setting this to "1s" blows this up by a factor 300.
```
jaronsummers, line 582: I rephrased here just to keep the line length short.

EdSchouten:
Ah, great. Thanks for addressing this. Older versions of the Protobuf library would allow you to fill in any duration string supported by Go's `time.Duration`. So `5m`, `10h`, etc. was all valid. Later on they restricted it to just accept seconds, for consistency with the Protobuf implementations of other languages. This documentation hasn't been updated to reflect that.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -576,12 +576,14 @@ message LocalBlobAccessConfiguration {
...
     //
-    // Recommended value: 5m
+    // Recommended value: "300s"
+    // In JSON google.protobut.Duration is a string, time in seconds,
```
EdSchouten, line 585: My recommendation would remove this sentence, though, as we have many google.protobuf.Duration fields in our configuration and this isn't the most logical place to document it.

Maybe we should just link to https://developers.google.com/protocol-buffers/docs/proto3#json from README.md?

jaronsummers: That makes sense to me. Apologies for not checking the rest of the codebase for similar things.

jaronsummers:
I didn't see any really obvious places to stick a link in the readme, but I removed the suggested sentence, and also updated my quotes to be single-quotes like the other examples in that file.

PR #131, opened 2021-11-21
Add support for compiling on Apple M1

Gormo:
It seems like only a small change is required in order to compile on darwin_arm64 i.e Apple M1 HW.

The node package supports darwin_arm64 from version 16+ so this PR will increment the node_version
to the latest supported version by the currently pointed out rules_nodejs.

The same change will also need to be made in the bb-remote-execution repository but I suppose
this should go in first and then a bump of the go_dependencies in the bb-remote-execution after that.


EdSchouten:
Thanks a lot!

PR #133, opened 2022-01-18
Fix uninitialized field in http_blob_access

scele:
Initialize httpClient to make httpBlobAccess functional.


EdSchouten:
Oh man. HTTP testing coverage for this code would have been lovely. You can see that I'm not using this backend myself.

Thanks for the fix!

PR #141, opened 2022-05-15
Synchronize disk storage on SIGTERM for graceful shutdown

moroten:
This change lets all gRPC calls finish in case of SIGTERM or SIGINT. This means that the client will not notice when the service is moving around in e.g. Kubernetes. Additionally, the blob stores are flushed to avoid loss of the latest data.

This change also introduces `/-/ready` to show when the gRPC sockets are listening.


moroten:
This PR is missing implementation of shutdown for local blob access. It is also missing early return from `Execute` and `WaitExecution` streaming calls. Unit tests are missing.

Please comment if this looks like a reasonable feature to merge. I also don't know what to do about integration testing.

moroten:
@EdSchouten Is this in line with your expectations on a graceful shutdown mechanism?

pkg/blobstore/blob_access.go
```diff
@@ -17,6 +17,11 @@ type BlobAccess interface {
 	Get(ctx context.Context, digest digest.Digest) buffer.Buffer
 	Put(ctx context.Context, digest digest.Digest, b buffer.Buffer) error
 	FindMissing(ctx context.Context, digests digest.Set) (digest.Set, error)
+
+	// Shutdown flushes buffers so that no data is lost when the terminating.
+	// No further calls will be done to any BlobAccess methods after Shutdown
+	// has been called.
+	Shutdown(ctx context.Context) error
```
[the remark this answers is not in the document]
moroten: Thank you for your response, Ed. I agree with your suggestion. How do I ensure that the shutdown methods are called in order? I can let the function pass through `BlobAccessInfo`, which effectively will become similar code complexity, or I can simply ignore that problem. I'll work more on this next week.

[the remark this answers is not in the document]
EdSchouten: Is there a strong requirements that shutdown methods are called in order? For LocalBlobAccess we want to force a flush, but we don't necessarily need to truly shut it down, right? The same holds for gRPC clients. They are safe to keep open if we shut down.

If we really care about ordering of things, it's always possible to use multiple errgroups, where the termination of one errgroup causes the context of the second errgroup to be cancelled.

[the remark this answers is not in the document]
moroten: Thank you, then I've got the idea of a pattern to use.

pkg/blobstore/configuration/new_blob_access.go
```diff
@@ -43,6 +44,7 @@ func newRedisClient(opt *redis.Options) *redis.Client {
 }
 
 func newNestedBlobAccessBare(configuration *pb.BlobAccessConfiguration, creator BlobAccessCreator) (BlobAccessInfo, string, error) {
+	lifeCycle := global.NewLifeCycleManager(0)
```
moroten, line 47: Shall I put the `lifeCycleManager` inside the `BlobAccessCreator` or as a separate argument?

pkg/blobstore/local/periodic_syncer.go
```diff
@@ -147,7 +147,15 @@ func (ps *PeriodicSyncer) ProcessBlockPut() {
 		<-ch
 		_, t = ps.clock.NewTimer(ps.minimumEpochInterval)
 	}
-	ps.lastSynchronizationTime = <-t
+	ps.syncNow(<-t)
+}
+
+func (ps *PeriodicSyncer) SyncNow() {
```
moroten, line 153: Should add test and provide a shorter `errorRetryInterval`.

moroten (commented):
Updated the code, but some details still left to sort out. It introduces a new `LifeCycleManager` to handle the contexts during the running and shutdown phases.

moroten:
@EdSchouten The latest commit is ready for review. I've tried it on a simple docker-compose example where I restart docker-compose immediately after building. This gives cache hits after this patch but not for the current master. I'm planning to add this test to bb-deployments in the future.

pkg/proto/configuration/global/global.proto
```diff
@@ -260,6 +260,21 @@ message Configuration {
...
+  // terminationGracePeriodSeconds in Kubernetes (default 30s) and the timeout
+  // in Docker (default 10s).
+  //
+  // Most unary RPCs are handled quickly, whereas streaming calls might be
+  // waiting for long, e.g. minutes for uploading files or even hours for
+  // executing test actions. A reasonable trade-off is to allow the unary
+  // calls to finish, within a few seconds, and let the client retry long
+  // streaming or blocking calls.
+  //
+  // Recommended value: 5s
```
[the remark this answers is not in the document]
moroten: Another part is that HTTP requests to the diagnostics server might also be interrupted now, but the clients should cope with it. Shall bb-scheduler and bb-browser allow for graceful shutdown on their web interface?

moroten:
@EdSchouten Please review the update. All the new commits shall be seen as fixups to the commit [Implement graceful shutdown](https://github.com/buildbarn/bb-storage/pull/141/commits/a4ae3bef669c73da7a39c90f64a8f1b8932144d2)

EdSchouten (changes requested):
Hi there! Thanks again for working on this!

My recommendation would still be to see if we can't just solve this by using `context.Context` and `errgroup.Group` as much as possible, and not resort to coming up with custom APIs like LifeCycleManager. To give some insight in what I mean, it would be nice if `main()` just had a construct like this:

```go
terminationWaitGroup, terminationContext := InstallTerminationSignalHandler()
```

With:

```go
func InstallTerminationSignalHandler() (*errgroup.Group, context.Context)
```

And that the errgroup and context would get passed along the codebase, getting used in  places where we need to hook into the termination path. In places where you want to detect whether termination should take place, you just do a `<-terminationContext.Done()`. If you want to ensure a piece of work can extend the lifetime of the process until graceful termination has taken place, you call `terminationWaitGroup.Go()`..

For example, for LocalBlobAccess, we could come up with a construct like this:
```go
			periodicSyncer := local.NewPeriodicSyncer(
				terminationContext,
				...)
			terminationWaitGroup.Go(func() {
				for periodicSyncer.ProcessBlockRelease() {}
			})
			terminationWaitGroup.Go(func() {
				for periodicSyncer.ProcessBlockPut() {}
 			})
```
So we don't create a third thread of execution that tries to do the flushing. We just make it so that the existing set of threads can detect termination and respond to it in a meaningful way.

Though we'll have to see how such an approach plays out, I guess that extending PersistentStateSource.NotifySyncStarting() to take a boolean argument (`isFinalSync bool`) also seems logical, so that ProcessBlockPut() can initiate termination.

pkg/blobstore/local/closable_block_list.go
```diff
@@ -0,0 +1,80 @@
...
+// just in the final call to the BlockListPutFinalizer that ErrClosedForWriting
+// will be returned.
+func (bl *ClosableBlockList) Put(index int, sizeBytes int64) BlockListPutWriter {
+	actualWriter := bl.BlockList.Put(index, sizeBytes)
+
+	return func(b buffer.Buffer) BlockListPutFinalizer {
+		// Copy data into the block without holding any locks.
+		actualFinalizer := actualWriter(b)
+
+		return func() (int64, error) {
```
EdSchouten, line 68: `flatBlobAccess` and `PeriodicSyncer` already ensure that this part of the storage backend is adequately locked. What I mean to say is that `bl.lock` can be eliminated entirely, assuming you set this flag while holding the lock as well.

My recommendation would also be to change the very top up `Put()` to check for `closedForWriting` as well. Otherwise you will end up allocating space and ingesting objects, even though you know you will end up returning an error. Something like this?
```go
if bl.closedForWriting {
	return func(b buffer.Buffer) BlockListPutFinalizer {
		b.Discard()
		return func() (int64, error) {
			return errClosedForWriting
		}
	}
}

File: pkg/blobstore/local/closable_block_list.go [hunk 6]
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"errors"
+	"sync"
+
+	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
+)
+
+// ErrClosedForWriting is an error code to indicate that the BlockList does not
+// accept any more write operations and ongoing write operations will fail.
+var ErrClosedForWriting = errors.New("Location record invalid")
```
EdSchouten, line 12: Please make sure to return a `status.Error()`. here. `errors.New()` doesn't yield errors with a proper gRPC status code. I would also recommend making it private, so that it doesn't pollute the godocs.

moroten:
Thank you Ed for the comments, I've updated the code and force pushed.

One reason for the `lifecycleManager` was to pass a single argument around instead of two, don't know what's best in the end. Are there any better names to pass around than `terminationContext` and `terminationGroup`? I though that they should not be mixed up with the gRPC context but maybe that's not any risk.

pkg/blobstore/configuration/icas_blob_access_creator.go
```diff
@@ -1,12 +1,15 @@
...
 import (
+	"context"
+
 	"github.com/buildbarn/bb-storage/pkg/blobstore"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/grpcclients"
 	"github.com/buildbarn/bb-storage/pkg/capabilities"
 	"github.com/buildbarn/bb-storage/pkg/digest"
 	"github.com/buildbarn/bb-storage/pkg/grpc"
 	pb "github.com/buildbarn/bb-storage/pkg/proto/configuration/blobstore"
+	"golang.org/x/sync/errgroup"
```
EdSchouten, line 12: Nit: empty line above this one, so that golang.org imports are distinct from github.com.

Rule of thumb inside the Buildbarn codebase is that all imports are grouped by second level domain, sorted in alphabetical order. I'm not sure how good of a rule it is, but it's simply the way the code is structured right now.

pkg/blobstore/configuration/new_blob_access.go
```diff
@@ -381,16 +384,17 @@ func newNestedBlobAccessBare(configuration *pb.BlobAccessConfiguration, creator
 				minimumEpochInterval,
 				keyLocationMapHashInitialization,
 				dataSyncer)
-			go func() {
-				for {
-					periodicSyncer.ProcessBlockRelease()
+			terminationGroup.Go(func() error {
+				for periodicSyncer.ProcessBlockRelease(terminationContext) {
```
EdSchouten, line 388: This looks great. Thanks!

pkg/blobstore/local/closable_block_list.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// errClosedForWriting is an error code to indicate that the BlockList does not
+// accept any more write operations and ongoing write operations will fail.
+var errClosedForWriting = status.Error(codes.Unavailable, "Closed for writing")
```
EdSchouten, line 12: Keep in mind that this error message gets sent back to the user. Maybe it's better to return something like: "Cannot write object to storage, as storage is shutting down"

pkg/blobstore/local/closable_block_list.go
```diff
@@ -0,0 +1,81 @@
...
+// ClosableBlockList is an implementation of BlockList that can be blocked for
+// writing. This is useful when shutting down to know that no more data is
+// written after the last sync.
+type ClosableBlockList struct {
+	BlockList
+
+	// Writes to closedForWriting must be synchronized with all write methods.
+	// This requirement is pushed down on the user of this struct.
+	closedForWriting bool
+}
```
EdSchouten, line 23: Could you maybe add a couple of lines here along the lines of:
```go
var _ BlockList = &ClosableBlockList{}
var _ PersistentStateSource = &ClosableBlockList{}
```
That documents which interfaces are implemented, and ensures that the build fails immediately when the contract is violated. This as opposed to causing a failure at the place of use.

pkg/blobstore/local/closable_block_list.go
EdSchouten: LocalBlobAccess is a bit of a complex beast, so the locking is also a bit wacky. To clarify it here:
```go
func (bl *ClosableBlockList) Put(index int, sizeBytes int64) BlockListPutWriter {
 	// This is where the exclusive lock is held, so that we can allocate space.
 	return func(b buffer.Buffer) BlockListPutFinalizer {
 		// This is where no locks are held, so that data can be ingested at the client's pace.
 		return func() (int64, error) {
 			// This is where the exclusive lock is held once again, so that we can record that ingestion failed/succeeded.
 		}
 	}
}
```
So comparing against `bl.closedForWriting` inside the middle layer is an illegal chess move here.

Hmmm... Now that I'm taking a step back: in case we have `bl.closedForWriting == true`, we never want to call into `actualFinalizer()`. The reason being that this would:
- Cause the epoch counter to be increased, even though we're terminating/syncing.
- Cause the object to become available for FindMissingBlobs()/Read() to other clients, even though we offer no guarantees it gets preserved.

Unfortunately, you can't get away with _not_ calling `actualFinalizer()`. You have to call it, as PersistentBlockList leaks otherwise. Maybe we should simply integrate this functionality into PersistentBlockList?

pkg/blobstore/local/periodic_syncer.go
```diff
@@ -144,10 +155,32 @@ func (ps *PeriodicSyncer) ProcessBlockPut() {
 		// The system was idle for some time. Wait a bit, so
 		// that the current epoch gets a meaningful amount of
 		// data.
-		<-ch
+		select {
+		case <-ctx.Done():
+			return false
```
EdSchouten, line 160: This would be the place to do final syncs to storage, right? Right now it just terminates without doing that.

pkg/global/apply_configuration.go
```diff
@@ -43,43 +47,94 @@ import (
...
 )
 
-// LifecycleState is returned by ApplyConfiguration. It can be used by
+const (
+	stateNotServing int32 = iota
+	stateServing
+)
+
+// ExitCodeInterrupted is used to signal a successful controlled shutdown.
+var ExitCodeInterrupted int = 8
```
EdSchouten, line 56: My preference would be to remove the signal handler and raise SIGTERM again. That way the calling process gets accurate process shutdown information.

EdSchouten:
> Are there any better names to pass around than `terminationContext` and `terminationGroup`?

I think for now it's good. If it turns out to become messy, we'll at some point try to refactor `NewBlobAccessFromConfiguration()` and `NewNestedBlobAccess` to be methods on a struct type.

pkg/blobstore/local/closable_block_list.go
moroten: Comparing `bl.closedForWriting` in the middle layer should still be valid. What I did wrong is that actualWriter must still be called to guarantee no leakage. I'll make the code simpler by removing that case.

I see your point, integrated it.

pkg/blobstore/local/periodic_syncer.go
EdSchouten: Exactly! So that's why my recommendation is to alter PersistentStateSource to have this method:
```go
NotifySyncStarting(isFinalSync bool)
```
And that would cause PersistentBlockList to set the flag.

moroten: `isFinalSync` has now been added and the (not) potential race of `bl.closedForWriting` has been removed.

moroten:
`isFinalSync` has now been added and the (not) potential race of `bl.closedForWriting` has been removed.

pkg/blobstore/local/periodic_syncer.go
```diff
@@ -144,13 +156,25 @@ func (ps *PeriodicSyncer) ProcessBlockPut() {
...
+			break
+		case <-ch:
+			_, t = ps.clock.NewTimer(ps.minimumEpochInterval)
+		}
+	}
+	var isFinalSync bool
+	select {
+	case <-ctx.Done():
+		ps.lastSynchronizationTime = ps.clock.Now()
+		isFinalSync = true
```
EdSchouten, line 171: Do we think it's necessary to do two syncs in this case?
- Once with `isFinalSync` set to `false`,
- Another one with `isFinalSync` set to `true`.

That way we first do a 'catch-up', where we try to flush as much data as possible. Only once that's done, there will be a brief point in time in which the service doesn't accept writes.

pkg/global/apply_configuration.go
```diff
@@ -43,43 +47,94 @@ import (
...
+// SetNotServing updates the health probe to report healthy but not ready.
+func (ds *DiagnosticsServer) SetNotServing() {
+	ds.state = stateNotServing
+}
+
+// ServeDiagnostics is a wrapper that calls DiagnosticsServer.Serve inside
+// a goroutine, managed by the provided errgroup.Group, and returns
+// immediately.
+func ServeDiagnostics(terminationContext context.Context, terminationGroup *errgroup.Group, diagnosticsServer *DiagnosticsServer) {
+	terminationGroup.Go(func() error {
```
EdSchouten, line 126: To me it's not completely clear why we need to use `terminationGroup` here.

- `terminationGroup` is what we need to register tasks that need to extend the lifetime of the process in case of ^C.
- `terminationContext` is what we need to detect that termination has started.

I think we only need to store the `terminationContext`, and let `/-/healthy` do this:
```go
switch {
case <-ctx.Done():
    // Return a HTTP 500, as we're shutting down.
default:
    // Return a HTTP 200.
}
```

pkg/grpc/server.go
```diff
@@ -120,8 +152,32 @@ func NewServersFromConfigurationAndServe(configurations []*configuration.ServerC
...
+// grpc_health_v1.HealthCheckResponse_SERVING.
+func (sg *ServerGroup) SetReady() {
+	for i, h := range sg.healthServers {
+		h.SetServingStatus(sg.healthCheckServices[i], grpc_health_v1.HealthCheckResponse_SERVING)
+	}
+}
+
+// Shutdown stops the gRPC services. It cancels all active RPCs. Clients are
+// supposed to retry the operations whenever the service is up again.
+func (sg *ServerGroup) shutdown() {
```
EdSchouten, line 176: I think this part can now be removed, right? As in, we'll focus on making sure that storage gets flushed properly, but we won't actually tear down the RPC server.

moroten: Kubernetes will reroute new connections to other pods during shutdown, but what will happen with open channels? They will continue as usual and then suddenly get "closed for writing" errors back, instead of "go away". Will Bazel retry on the same channel or make a new connection? I'm in favour of you suggestion as I don't see the point of a "perfect" shutdown, but just want to make sure we don't get into a bad state.

pkg/global/apply_configuration.go
moroten: If either the diagnostics or gRPC servers fail, then the `terminationGroup` will be cancelled and the data will be synced. Using a standard goroutine together with `log.Fatal` there might be data loss. I don't know why they would fail, but for me that sounds reasonable. What do you think?

pkg/blobstore/local/periodic_syncer.go
moroten: Wired it up with two syncs this time.

moroten:
Forgot to print the returned error from `terminationGroup`, sorted out that now.

moroten:
I went through the code again and reduced the diffs compared to the `master` branch. Renaming `LifecycleState` to `DiagnosticsServer` in the first commit is not necessary, but I think it is still a better name. The rest of the commit chain should be squashed but I keep it non-squashed in case it's wanted for review.

EdSchouten:
Hi Fredrik, as discussed on Slack, I have gone ahead and partially merged this change with some minor cleanups applied:

- Even though I initially hinted towards using an errgroup, it looks like a sync.WaitGroup is sufficient. We don't necessarily care about propagating errors from LocalBlobAccess upwards; the existing code that logs such errors is sufficient.
- Giving it some more thought, ProcessBlockRelease shouldn't stop during shutdown. Stopping this could cause LocalBlobAccess to stop accepting writes prematurely due to new blocks being unavailable. It's only ProcessBlockPut that needs special treatment.
- I've moved all of the signal handling code into `pkg/global/graceful_termination_handler.go`.

I've left this PR open, because there are also parts that I didn't incorporate. Most notably, changes to the diagnostics HTTP server. Be sure to rebase your PR on top of latest master and we can discuss what to do with those changes.

moroten:
Thank you Ed for cleaning it all up. The leftovers are:

1. I would like to postpone `/-/healthy` to after sockets are ready for the gRPC servers. We can just skip introducing the `/-/ready` probe path as startup is so quick.
2. A couple of places prints errors using `log.Fatal`, e.g. in `grpc/server.go`. Should the disk storage synchronize then as well or are all those error cases something that will never happen?

I'll create a separate PR for 1. What about 2?

EdSchouten:
> 2. A couple of places prints errors using `log.Fatal`, e.g. in `grpc/server.go`. Should the disk storage synchronize then as well or are all those error cases something that will never happen?

I wouldn't worry about the `log.Fatal()`s too much. I mean, these typically only happen _very_ closely after startup, generally before (significant) amounts of traffic may be processed. Storage pods will typically only crashloop on startup in case of misconfigurations.

moroten:
Opened #142 for delaying the healthy service status a bit during startup.

PR #142, opened 2022-07-15
Postpone healthy service status response a bit

moroten:
The healthy and serving status, i.e. HTTP `/-/healthy` and `grpc_health_v1.HealthCheckResponse_SERVING`, will now be postponed until the whole service is up and running. Before, the healthy status was potentially reported before starting to listen to the gRPC ports.

This removes an error mode where Kubernetes would believe that the service was ready just before it crashed. The previous behaviour could lead to a cluster being upgraded to an erroneous state even for a simple jsonnet configuration error.

(I think this was the problem, don't remember exactly how it did behave.)

### DIFF
--- a/cmd/bb_replicator/main.go
+++ b/cmd/bb_replicator/main.go
@@ -57,15 +57,13 @@ func main() {
 		log.Fatal("Failed to create replicator: ", err)
 	}
 
-	go func() {
-		log.Fatal(
-			"gRPC server failure: ",
-			bb_grpc.NewServersFromConfigurationAndServe(
-				configuration.GrpcServers,
-				func(s grpc.ServiceRegistrar) {
-					replicator_pb.RegisterReplicatorServer(s, replication.NewReplicatorServer(replicator))
-				}))
-	}()
+	if err := bb_grpc.NewServersFromConfigurationAndServe(
+		configuration.GrpcServers,
+		func(s grpc.ServiceRegistrar) {
+			replicator_pb.RegisterReplicatorServer(s, replication.NewReplicatorServer(replicator))
+		}); err != nil {
+		log.Fatal("gRPC server failure: ", err)
+	}
 
 	lifecycleState.MarkReadyAndWait()
 }
--- a/cmd/bb_storage/main.go
+++ b/cmd/bb_storage/main.go
@@ -143,56 +143,54 @@ func main() {
 		capabilitiesProviders = append(capabilitiesProviders, buildQueue)
 	}
 
-	go func() {
-		log.Fatal(
-			"gRPC server failure: ",
-			bb_grpc.NewServersFromConfigurationAndServe(
-				configuration.GrpcServers,
-				func(s grpc.ServiceRegistrar) {
-					if contentAddressableStorage != nil {
-						remoteexecution.RegisterContentAddressableStorageServer(
-							s,
-							grpcservers.NewContentAddressableStorageServer(
-								contentAddressableStorage,
-								configuration.MaximumMessageSizeBytes))
-						bytestream.RegisterByteStreamServer(
-							s,
-							grpcservers.NewByteStreamServer(
-								contentAddressableStorage,
-								1<<16))
-					}
-					if actionCache != nil {
-						remoteexecution.RegisterActionCacheServer(
-							s,
-							grpcservers.NewActionCacheServer(
-								actionCache,
-								int(configuration.MaximumMessageSizeBytes)))
-					}
-					if indirectContentAddressableStorage != nil {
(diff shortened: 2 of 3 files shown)


cmd/bb_replicator/main.go
```diff
@@ -57,15 +58,18 @@ func main() {
 		log.Fatal("Failed to create replicator: ", err)
 	}
 
+	readyGroup := &sync.WaitGroup{}
+	readyGroup.Add(1)
 	go func() {
 		log.Fatal(
 			"gRPC server failure: ",
 			bb_grpc.NewServersFromConfigurationAndServe(
+				readyGroup,
```
moroten, line 67: The alternative I see is to do
```golang
startupCtx, cancleStarupCtx := context.WithCancel()
grpcServerErrorChannel := bb_grpc.NewServersFromConfigurationAndServe(startupCtx, ...)
select {
case err := <-grpcServerErrorChannel:
    // Busy ports or other configuration errors will be reported here.
    log.Fatal("gRPC server failure: ", err)
default:
    break
}
go func() {
    log.Fatal("gRPC server failure: ", <-grpcServerErrorChannel)
}()
cancelStartup()
lifecycleState.MarkReadyAndWait()
```
This seems to be a bit bloated code compared to how it looks like today, but it is more clear what is happening. To make things more tidy, the error wrapping should be moved to a helper function.

moroten:
@EdSchouten friendly ping.

cmd/bb_storage/main.go
```diff
@@ -193,7 +197,7 @@ func main() {
 					}
 				}))
 	}()
-
+	readyGroup.Wait()
 	lifecycleState.MarkReadyAndWait()
```
EdSchouten, line 201: Could we do this as part of MarkReadyAndWait()?

I think that the main annoyance right now is that we have one instance of `health.Server` per RPC server, which is why you're trying to solve this as part of NewServersFromConfigurationAndServe(). But what if we eliminated that? I think it's completely safe to call `health.NewServer()` exactly once as part of `global.ApplyConfiguration()`, and somehow get that passed along as we create gRPC servers.

What are your thoughts?

EdSchouten (commented):
Sorry for the delays. Completely overlooked this PR!

cmd/bb_storage/main.go
moroten: It would be great, but at the moment `grpc.proto` defines individual service names for each server. I don't know who uses such configuration. Would you like to remove it or do you have other ideas?
```protobuf
  // Service name for health check requests. The gRPC server will
  // report itself healthy for this service via the grpc.health.v1
  // protocol.
  string health_check_service = 7;
```

moroten: Alternatively, `global.ApplyConfiguration` can provide a factory function which is passed to `NewServersFromConfigurationAndServe`. Then everything will still work.

EdSchouten: Indeed! That sounds like a good solution. We already provide a factory for clients as well.

moroten: I now realised that the `readyGroup.Wait()` call is there to make sure that `NewServersFromConfigurationAndServe` has finished with all the initialisation, e.g. started listening to all the ports.

The error returned from `NewServersFromConfigurationAndServe` serves dual purposes at the moment, both startup errors and errors during runtime (from `s.Serve()`). It is wrapped with "xyz gRPC server failure: " in `main.go` and I don't want to duplicate that error wrapping. Any ideas?

moroten:
@EdSchouten Should `NewServersFromConfigurationAndServe` change name so we don't forget to update it wherever it is used? Not sure what name then.

pkg/grpc/server.go
```diff
@@ -117,24 +116,28 @@ func NewServersFromConfigurationAndServe(configurations []*configuration.ServerC
 
 		// TCP sockets.
 		for _, listenAddress := range configuration.ListenAddresses {
+			// Make a copy for the go routine.
+			listenAddress := listenAddress
```
EdSchouten, line 120: Nit: I'm generally not a fan of shadowing of existing variables. Could you change the loop itself to use `listenAddressIter`? No need to add the comment above; It's sort of unwritten 'common practice' in the Buildbarn tree to use `Iter` as the suffix.

EdSchouten:
> @EdSchouten Should `NewServersFromConfigurationAndServe` change name so we don't forget to update it wherever it is used? Not sure what name then.

Name is good. Thanks!

pkg/grpc/server.go
moroten: Done that. Waiting for the GitHub actions to run.

PR #145, opened 2022-09-29
feat: add tls client auth jmes patterns

joeljeske:
This change adds 2 new options to the grpc.TLSClientCertificateAuthenticationPolicy message: 

* `validation_jmespath_expression` which is a boolean expression to approve the TLS connection and optionally sending an `Unauthenticated`.

* `metadata_extraction_jmespath_expression` which is a metadata extractor to be used in downstream authorization decisions

Both have the same context at their disposal, which are the set of usable Subject Alternative Names in the Certificate. Notably, this does not include the Common Name in the certificate.

```jsonnet
{
    // Contains every DNS Subject Alt Name provided
     "dnsNames": ["example.com", "*.sub.example.com"],
   
     // Contains every Email Subject Alt Name provided
     "emailAddresses": ["me@example.com"] 
}
```

Config Migrations: 

* Change usages of `TLSClientCertificateAuthenticationPolicy.metadata` to `TLSClientCertificateAuthenticationPolicy.metadata_extraction_jmespath_expression`
* Set `TLSClientCertificateAuthenticationPolicy.validation_jmespath_expression` to ``` `true` ```  to maintain existing behavior


pkg/grpc/tls_client_certificate_authenticator.go
```diff
@@ -61,5 +65,49 @@ func (a *tlsClientCertificateAuthenticator) Authenticate(ctx context.Context) (*
...
+
+	return auth.NewAuthenticationMetadataFromRaw(metadataRaw)
+}
+
+func getClientCertificateJMESPathSearchContext(cert *x509.Certificate) (map[string]any, error) {
+	// We have to go through this copying and json dance in order to 
+	// ensure that we don't replace [] with null, and that we have the proper
+	// types needed for JMESPath to search over without typing failures
+
+	dnsNames := make([]interface{}, 0, len(cert.DNSNames))
```
EdSchouten, line 97: Nit: Use `any` instead of `interface{}`

pkg/grpc/tls_client_certificate_authenticator.go
```diff
@@ -61,5 +65,49 @@ func (a *tlsClientCertificateAuthenticator) Authenticate(ctx context.Context) (*
...
 		return nil, util.StatusWrapWithCode(err, codes.Unauthenticated, "Cannot validate TLS client certificate")
 	}
-	return a.metadata, nil
+
+	searchContext, err := getClientCertificateJMESPathSearchContext(certs[0])
+	if err != nil {
+		return nil, util.StatusWrapWithCode(err, codes.Unauthenticated, "Failed to create a search context for extracting metadata")
+	}
+
+	// Validate the client cert matches our expectations
```
EdSchouten, line 74: Please end all sentences in comments with a full stop (period), also in the Protobuf source files.

pkg/grpc/tls_client_certificate_authenticator.go
```diff
@@ -61,5 +65,49 @@ func (a *tlsClientCertificateAuthenticator) Authenticate(ctx context.Context) (*
...
+	// Extract metadata from the client cert
+	metadataRaw, err := a.metadataExtractor.Search(searchContext)
+	if err != nil {
+		return nil, util.StatusWrapWithCode(err, codes.Unauthenticated, "Cannot extract metadata from TLS client certificate")
+	}
+
+	return auth.NewAuthenticationMetadataFromRaw(metadataRaw)
+}
+
+func getClientCertificateJMESPathSearchContext(cert *x509.Certificate) (map[string]any, error) {
```
EdSchouten, line 92: This function can't fail, right? If so, please remove the `error`.

pkg/grpc/tls_client_certificate_authenticator_test.go
```diff
@@ -99,16 +104,21 @@ func TestTLSClientCertificateAuthenticator(t *testing.T) {
...
-				"username": structpb.NewStringValue("John Doe"),
+				"email": structpb.NewStringValue("me@example.com"),
 			},
 		}),
 	})
-	authenticator := bb_grpc.NewTLSClientCertificateAuthenticator(clientCAs, clock, expectedMetadata)
+
+	aValidator := jmespath.MustCompile(`contains(dnsNames, 'a.example.com')`)
+	bValidator := jmespath.MustCompile(`contains(dnsNames, 'b.example.com')`)
+	metadataExtractor := jmespath.MustCompile(`{"public": {"email": emailAddresses[?ends_with(@, '@example.com')] | [0] } }`)
```
EdSchouten, line 114: My recommendation would be to use a simpler expression here like `{"public": @}`, because that would allow us to validate the format of everything that comes out of this.

pkg/grpc/tls_client_certificate_authenticator.go
```diff
@@ -61,5 +65,49 @@ func (a *tlsClientCertificateAuthenticator) Authenticate(ctx context.Context) (*
...
+		return nil, util.StatusWrapWithCode(err, codes.Unauthenticated, "Failed to create a search context for extracting metadata")
+	}
+
+	// Validate the client cert matches our expectations
+	validationResult, err := a.validator.Search(searchContext)
+	if err != nil {
+		return nil, util.StatusWrapWithCode(err, codes.Unauthenticated, "TLS client certificate failed JMESPath validation")
+	}
+	if validationResult != true {
+		return nil, util.StatusWrapWithCode(errors.New("JMESPath validation result was not true"), codes.Unauthenticated, "TLS client certificate failed JMESPath validation")
```
EdSchouten, line 80: You can just return `status.Error(codes.Unauthenticated, "....")` here.

pkg/proto/configuration/grpc/grpc.proto
```diff
@@ -277,11 +277,67 @@ message TLSClientCertificateAuthenticationPolicy {
...
+  // with code UNAUTHENTICATED; not PERMISSION_DENIED.
+  string validation_jmespath_expression = 3;
+
+  // JMESPath expression for converting Certificates information to a
+  // Protobuf message of type buildbarn.auth.AuthenticationMetadata.
+  // The resulting metadata can be used for purposes such as authorization.
+  //
+  // The context data has the following structure:
+  //
+  // {
```
EdSchouten, line 326: Instead of documenting the input data structure once again, I guess it's sufficient to just refer to the description above.

EdSchouten:
Thanks for working on this, Joel. Looking forward to see this land!

pkg/grpc/tls_client_certificate_authenticator.go
EdSchouten: Also please make sure that these error messages don't contain too many technicalities. Keep in mind that these error messages are returned to the user. The JMESPath expressions are something that an administrator of the cluster specifies; not the user.

PR #146, opened 2022-10-17
Make java_proto_library work

Jonpez2:
Remap the outer classname so that an invocation of java_proto_library doesn't get a classname collision


EdSchouten:
Looks good! Could you address the CI failure? Will merge upon success.

Jonpez2:
I think I did it

Jonpez2:
Apparently not. What is that test complaining about? I haven’t touched any byte arrays…

EdSchouten:
We check in the resulting .pb.yo files, to make 'go get' work. Take a look at the .github/workflows files to see what command you need to run to regenerate them.

Jonpez2:
OK, sorry, I'm not seeing any commands in the workflows that do anything but clang-format on the protos?  I'm sorry I'm being dumb!

EdSchouten:
```
            find . bazel-bin/pkg/proto -name '*.pb.go' -delete || true
            bazel build $(bazel query 'kind("go_proto_library", //...)')
            find bazel-bin/pkg/proto -name '*.pb.go' | while read f; do
              cat $f > $(echo $f | sed -e 's|.*/pkg/proto/|pkg/proto/|')
            done
```

Jonpez2:
Oh duh my fork is out of date, I'm sorry

Jonpez2:
shall we try this one?  Maybe I got it right...

Issue #120, opened 2021-05-26
Allow us to point all the grpc servers at files to pick up credentials


Jonpez2:
I'm currently trying to plug together SPIFFE (https://github.com/spiffe) and buildbarn in a deployment on GKE, to give me well-managed mTLS.  My current approach is to use the spiffe-helper (https://github.com/spiffe/spiffe-helper) to keep pem files up to date on the in one container, with spiffe-helper just invoking 'sleep infinity', and then having the buildbarn container share a filesystem so that it can read the resultant certs.  Note that SPIFFE rotates the leaf certificates very frequently - on the order of once per hour.  This is pretty painful however, as buildbarn uses jsonnet to pipe creds through to the process, which means that a hot reload of the rotated certs is impossible without a full restart.

I'm thinking that the best option here would be for buildbarn to allow specification of a set of files for certs, and then just use grpc's hot reloading code as demonstrated here - https://github.com/grpc/grpc-go/blob/master/security/advancedtls/examples/credential_reloading_from_files/server/main.go

Does that seem right?  Or am I missing something obvious?

Thank you!

Jonpez2:
Any thoughts on this please?  Would you prefer me to propose a code change?

EdSchouten:
Hey! Sorry for letting you wait.

I'm fine with supporting use cases like these. I think the cleanest way to support this is to factor out the keypair string fields we have in our config files to a separate Protobuf message. We can then use a `oneof` in there to choose between providing the key material inline, vs. using other frameworks.

Instead of using stuff like spiffe-helper, I'm also fine with integrating https://github.com/spiffe/go-spiffe/tree/main/v2. That way you don't need to run all sorts of sidecars just to get TLS working.

Jonpez2:
My turn to apologize to you!

So actually an even better approach (for me anyway!) would be to upgrade to grpc 1.38.0 and then allow use of xds, like this: https://cloud.google.com/traffic-director/docs/security-proxyless-setup.  That still requires a bit of changes inside buildbarn - specifically, all the servers will have to do this (https://github.com/grpc/grpc-go/blob/master/examples/features/xds/server/main.go#L82) - but it means that all the certificate specification and discovery and so forth becomes someone else's problem.  Wdyt?

EdSchouten:
Oh, that would be pretty sweet. Feel free to submit a PR to add a gRPC configuration for enabling xDS!

Jonpez2:
On it

[...]

PR #148, opened 2022-10-26
[wip] add server xDS support

joeljeske:
Basic integration for server support. 

Tests to add: 
- Validate if the various serverOptions are compatible when using xDS server or there are runtime errors
- Validate basic static xDS bootstrap config reading


pkg/proto/configuration/grpc/grpc.proto
```diff
@@ -164,8 +164,8 @@ message ServerConfiguration {
   // inside directories that have access controls set up properly.
   repeated string listen_paths = 2;
 
-  // TLS configuration. TLS is not enabled when left unset.
-  buildbarn.configuration.tls.ServerConfiguration tls = 3;
+  // was TLS, moved tl TLS option in transport_credentials.
+  reserved 3;
```
EdSchouten, line 168: No need to mark 3 as reserved. Just let `tls` below use field number 3 again. With Protobuf it's safe to upgrade a single field to a `oneof`, provided you preserve the field number.

joeljeske:
TL;DR Closing this in favor of #149 

I initially started down this path of integrating into xDS with the goal of supporting certificate rotation from cert files on disk. In reading #120 thought it best to use xDS to support this use case. I did not have plans to actually integrate with a running xDS server, but hoping use some of the client-only grpc golang features of xDS. I have decided that this is not a good approach for the following reasons. 

First, I've found that xDS integration with go-gRPC does not appear to function without the use of an accessible xDS server to manage configuration. I was originally under the impression that after integration with go-grpc/xds, we would be able to specify a xDS bootstrap file that is static on only read from disk without the need for a xDS server. Whereas the xDS spec does allow for a file_watcher[ certificate provider](https://github.com/grpc/proposal/blob/master/A29-xds-tls-security.md#file_watcher-certificate-provider), there must still be an xDS server that requests clients to use such a provisioned certificate.

Second, I have found that there are a variety of expectations made on an application that claims to support xDS. If Buildbarn was xDS compatible, one would expect that they could configure the following bits on their xDS server: Service Discovery, Route Authorization, Load Balancing, Retry/Backoff. I will not be using any of those features, and am not the right person to integrate those bits into Buildbarn. Obviously there is significant overlap with this feature set with the existing configuration of Buildbarn. It is probably not desirable to support this alternate strategy without deeper analysis.

PR #149, opened 2022-10-28
Support TLS Server/Client certificates read from a file, and refreshe…

joeljeske:
…d on an interval.

Incompatible Changes:

This change moves the existing `{client,server}_certificate` and `{client,server}_private_key` into `{client,server}_key_pair.inline.certificate`  and `{client,server}_key_pair.inline.private_key`

This change adds an alternate strategy for specifying certificates/private keys by using `{client,server}_key_pair.files.certificate_path` and `{client,server}_key_pair.files.private_key_path`  which should reference PEM files on disk. A refresh_interval must also be specified to dictate the interval at which the files should be read and used in place of the existing certificate/key if such files have been changed.

In kubernetes for example, this allows a side-car to rotate certificate based on user's needs.


pkg/proto/configuration/tls/tls.proto
```diff
@@ -36,14 +38,26 @@ message ClientConfiguration {
   // addresses. This field can be used to specify the expected DNS name
   // of the server certificate.
   string server_name = 5;
+
+  // x509 Key-Pair to use for client TLS. No client
+  // certificate/private key is used when left unset.
+  oneof client_key_pair {
```
EdSchouten, line 44: Could we move this into its own message?
```
message X509KeyPair {
   message Inline {
   }

   message Files {
   }

   oneof { ... }
}
```
That way client and server certificates will have the same schema. I guess that would simplify the code a bit as well?

pkg/proto/configuration/tls/tls.proto
```diff
@@ -52,4 +66,32 @@ message ServerConfiguration {
...
+  oneof server_key_pair {
+    // Use inline data for the key-pair.
+    X509KeyPairData server_key_pair_data = 4;
+
+    // Load key pair from referenced files, with a refresh interval.
+    X509KeyPairFiles server_key_pair_files = 5;
+  }
+}
+
+message X509KeyPairData {
```
EdSchouten, line 80: InlineX509KeyPair? Or maybe just `message Inline` embeded into `message X509KeyPair`.

pkg/proto/configuration/tls/tls.proto
```diff
@@ -52,4 +66,32 @@ message ServerConfiguration {
...
+  // PEM data for the certificate used for TLS.
+  string certificate = 1;
+  // PEM data for the private key used for TLS.
+  string private_key = 2;
+}
+
+message X509KeyPairFiles {
+  // PEM file path for the certificate used for TLS.
+  string certificate = 1;
```
EdSchouten, line 89: `certificate_path`?

pkg/proto/configuration/tls/tls.proto
joeljeske: Done!

pkg/proto/configuration/tls/tls.proto
joeljeske: Done!

pkg/util/tls_certificate.go
```diff
@@ -0,0 +1,118 @@
...
+// StopRefresh stops reloading the certificate from files.
+func (r *RotatingTLSCertificate) StopRefresh() {
+	r.refreshStop <- true
+}
+
+func (r *RotatingTLSCertificate) refreshOnInterval(refreshInterval time.Duration) {
+	t := time.NewTicker(refreshInterval)
+	for {
+		select {
+		case <-r.refreshStop:
```
EdSchouten, line 108: My recommendation would be to remove the spawning of the goroutine from this type. Just make it so that there is a single `Refresh()` method on this, and let tls.go spawn the goroutine. That way the tests remain simple. We also won't need to care about the refreshing then, as tls.go can simply let it run indefinitely.

pkg/util/tls.go
```diff
@@ -117,21 +180,30 @@ func NewTLSConfigFromServerConfiguration(configuration *configuration.ServerConf
...
-		tlsServerCertificateNotBeforeTimeSeconds.WithLabelValues(dnsName).Set(float64(leaf.NotBefore.UnixNano()) / 1e9)
-		tlsServerCertificateNotAfterTimeSeconds.WithLabelValues(dnsName).Set(float64(leaf.NotAfter.UnixNano()) / 1e9)
+	// Require the use of server-side certificates.
+	switch keyPair := configuration.ServerKeyPair.KeyPair.(type) {
+	case *pb.X509KeyPair_Inline_:
+		var cert tls.Certificate
+		if cert, err = tls.X509KeyPair([]byte(keyPair.Inline.Certificate), []byte(keyPair.Inline.PrivateKey)); err != nil {
+			return nil, StatusWrapWithCode(err, codes.InvalidArgument, "Invalid server certificate or private key")
+		}
+		updateServerTLSCertificateExpiry(&cert)
```
EdSchouten, line 194: What is the difference between the client & server code? Is it just the Prometheus metrics that need to be used? What are your thoughts on  the existing metrics to `buildbarn_tls_certificate_not_before_time_seconds` (i.e., removing the `client` midfix)? We could just let that be the value of some kind of label that's attached to them.

Once that's done, it should be pretty trivial to move this code into a common implementation that both the client & server code can use.

pkg/util/tls_certificate.go
joeljeske: Sounds good to me, that makes sense!

pkg/util/tls.go
joeljeske: I'm ok to rename the metric, I was approaching it this way to avoid a more annoying breaking change by renaming the metric. If thats not a concern, then I will update to use a metric with label `certificate_usage=server|client`

EdSchouten: Sounds perfect!

pkg/util/tls.go
```diff
@@ -75,13 +126,29 @@ func NewTLSConfigFromClientConfiguration(configuration *configuration.ClientConf
...
 	tlsConfig.ServerName = configuration.ServerName
 
-	if configuration.ClientCertificate != "" && configuration.ClientPrivateKey != "" {
-		// Serve a client certificate when provided.
-		cert, err := tls.X509KeyPair([]byte(configuration.ClientCertificate), []byte(configuration.ClientPrivateKey))
-		if err != nil {
-			return nil, StatusWrapWithCode(err, codes.InvalidArgument, "Invalid client certificate or private key")
+	// No client TLS is used when this is unset.
+	if configuration.ClientKeyPair != nil {
+		// Require the use of server-side certificates.
```
EdSchouten, line 131: Could we factor out this code, so that client/server creation can use the same code?

pkg/util/tls_certificate.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"bytes"
+	"crypto/tls"
+	"os"
+	"sync"
+)
+
+// RotatingTLSCertificate provides an up-to-date certificate given file paths and a refresh interval.
+type RotatingTLSCertificate struct {
+	sync.RWMutex
```
EdSchouten, line 12: Nit: Could you make this RWMutex named? There is no need for this type to offer a `Lock()`/`Unlock()` as part of the public API, right?

Furthermore, general convention in the Buildbarn tree is to group struct elements by locking requirements. So maybe best to write it like this:
```go
type RotatingTLSCertificate struct {
    certFile string
    keyFile string

    lock sync.RWMutex
    cert *tls.Certificate
    certData []byte
    keyData []byte
}
```
This makes it more obvious that certFile and keyFile are immutable.

EdSchouten (commented):
Looking slick, Joel!

pkg/util/tls_test.go
```diff
@@ -207,6 +267,14 @@ func TestTLSConfigFromClientConfiguration(t *testing.T) {
 }
 
 func TestTLSConfigFromServerConfiguration(t *testing.T) {
+	tempDir := t.TempDir()
+	exampleCertFile := tempDir + "/example-cert.pem"
```
EdSchouten, line 271: filepath.Join() ?

pkg/util/tls_certificate.go
joeljeske: Done! Thanks for the tip :)

joeljeske:
Thanks Ed! Very helpful tips, I appreciate it. Changes are made! :)

pkg/util/tls_test.go
EdSchouten: Whoops! It looks like we can also use filepath.Join() here.

EdSchouten (approved):
One small change requested, and then we're good with merging this.

PR #155, opened 2022-12-06
Update completenessChecking section in example configuration

ferraith:
The config scheme changed with the commit https://github.com/buildbarn/bb-storage/commit/1b84fa824dc60e77776ce50e05c549fdf20c089b. This change should be also reflected in the example configuration.


EdSchouten:
Thanks a lot, Andreas!

Issue #153, opened 2022-12-01
Missing shell in bb-storage docker image


ferraith:
From your statement https://github.com/buildbarn/bb-storage/issues/27 I would expect that a shell e.g. `sh` is baked into the docker image published to Docker Hub. Because `kubectl exec ...` is failing for me I extracted the image, but couldn't find any shell.

I assume this is because `go_image` is called:
https://github.com/buildbarn/bb-storage/blob/81407c8ba4eadaf53e13c21af5203007c38ceba6/cmd/bb_storage/BUILD.bazel#L40

From my understanding the pure flag ensure that an app-only image is built.

EdSchouten:
Closing this, as @ferraith found a way to debug without putting a shell in the container.

PR #156, opened 2022-12-07
Switch to static base image with Busybox shell

ferraith:
This pull-request fixes the issue #153. In case for the Buildbarn developers it's still a goal to provide bb_storage images including a shell-like environment the underlying base image needs to be changed. From a user perspective having a shell available in a bb-storage container simplifies debugging in cloud environments a lot.

### DIFF
--- a/cmd/bb_storage/BUILD.bazel
+++ b/cmd/bb_storage/BUILD.bazel
@@ -36,8 +36,8 @@ go_binary(
 
 go_image(
     name = "bb_storage_container",
+    base = "@go_debug_image_static//image",
     embed = [":bb_storage_lib"],
-    pure = "on",
     visibility = ["//visibility:public"],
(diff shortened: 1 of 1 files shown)


EdSchouten:
Hey there! Are you aware that Kubernetes has in the meantime gained proper debug sidecar support?

https://kubernetes.io/docs/tasks/debug/debug-application/debug-running-pod/

With that in mind, I'm less inclined to accept changes like these.

EdSchouten:
You can also just inspect the config file by looking at the config map, right? Logs also go to stdout and are preserved by kubelet, so there is generally no need to inspect the log files by exec’ing into the container, I guess.

ferraith:
Hi Ed! Accepted, we can just close the pr in case you are fine with it.

PR #167, opened 2023-03-11
fix: create multi-arch container images

thesayyn:


thesayyn:
I will remove `tools/oci/registry` folder once https://github.com/bazel-contrib/rules_oci/issues/68 is fixed.

thesayyn:
for those who want to try out the image built with these changes; 

`podman run  ttl.sh/buildbarn/bb-copy:not-built-with-stamp-99b5ad0`

built with `bazel run --stamp cmd/bb_copy:bb_copy_image_push`

cmd/bb_copy/BUILD.bazel
```diff
@@ -28,14 +29,31 @@ go_binary(
...
+    arch = "amd64",
+    embed = [":bb_copy_lib"],
+    pure = "on",
+    visibility = ["//visibility:public"],
+)
+
+oci_image_index(
+    name = "bb_copy_image",
+    images = [
+        ":bb_copy_image_arm64",
```
EdSchouten, line 50: I was wondering, why do we need to declare the go_image() multiple times? Wouldn’t it be possible to use Bazel’s transitions feature to automatically build for the right platform?

tools/oci/go_image.bzl
```diff
@@ -0,0 +1,33 @@
...
+        srcs = srcs,
+        embed = embed,
+        goarch = arch,
+        goos = os,
+        pure = pure,
+        tags = tags,
+        visibility = visibility,
+    )
+
+    pkg_tar(
```
EdSchouten, line 17: This seems like a lot of manual work. Compared to rules_docker’s approach it likely has the downside that runfiles also don’t work out of the box. (Though we fortunately don’t depend on that.)

Is rules_oci planning on streamlining that?

cmd/bb_copy/BUILD.bazel
thesayyn: that would be possible but it's not implemented yet as bazel's toolchain support has its issues like optional toolchains. ecosystem isn't there yet so we are waiting on that.

tools/oci/go_image.bzl
thesayyn: rules_docker's approach has its downsides as well. it's a huge maintenance burden to us, to take language rules as dependencies and successfully maintain them. rules_docker failed to do so. 

we have two approaches here 

1- design your binaries to not rely on runfiles. statically built languages are less likely to be already dependent on runfiles. 
2- have a rule that outputs bazel-like runfiles layout in a tar file. we have experimented with this in [here](https://github.com/bazel-contrib/rules_oci/blob/122d1b181e3b6d71fed75eb7cf7b61f788b38b82/language/runfiles.bzl). it's a generic helper rule that is capable of transitioning to the target platform and laying out the runfiles tree as a tar layer.

thesayyn: generic `runfiles` helper and `pkg_tar` is the superior choice and we'll be spending our energy fixing bazel and rules_tar to be able to yield a runfiles tree.

cmd/bb_copy/BUILD.bazel
thesayyn: personal opinion: having separate  `*_image` targets make it easier to be understood by entry-level users. I see us supporting both though.

EdSchouten: The downside of having the separate targets is that it doesn't scale if we want to support even more architectures. For example, if someone wants to get the existing 32-bit Intel/ARM binaries or RISC-V added, we need to patch up all BUILD files in all Buildbarn projects to build for that architecture.

Would it be possible to at least move it into a common macro?

thesayyn: for sure. I'll get on that

EdSchouten: With regards to optional toolchains, this is about the lack of a C/C++ toolchain for those platforms, right? We can always declare placeholder toolchains for those architectures. I've always had good enough experiences doing something like this:
```
load("@bazel_tools//tools/cpp:cc_toolchain_config_lib.bzl", "tool_path")

def _empty_cc_toolchain_config_impl(ctx):
    return cc_common.create_cc_toolchain_config_info(
        ctx = ctx,
        toolchain_identifier = "empty-toolchain",
        host_system_name = "local",
        target_system_name = "local",
        target_cpu = "k8",
        target_libc = "unknown",
        compiler = "false",
        abi_version = "unknown",
        abi_libc_version = "unknown",
        tool_paths = [
            tool_path(
                name = "gcc",
                path = "/bin/false",
            ),
        ],
    )

empty_cc_toolchain_config = rule(
    implementation = _empty_cc_toolchain_config_impl,
    attrs = {},
    provides = [CcToolchainConfigInfo],
)
```

Which you can then use as follows:
```
load("//bazel/cpp:cc_toolchain.bzl", "empty_cc_toolchain_config")

filegroup(name = "empty")

empty_cc_toolchain_config(name = "empty_cc_toolchain_config")

cc_toolchain(
    name = "empty_cc_toolchain",
    all_files = ":empty",
    compiler_files = ":empty",
    dwp_files = ":empty",
    linker_files = ":empty",
    objcopy_files = ":empty",
    strip_files = ":empty",
    supports_param_files = 0,
    toolchain_config = ":empty_cc_toolchain_config",
    toolchain_identifier = "empty-toolchain",
)

# This is a bogus C++ toolchain definition for Linux on macOS. We
# declare it here, so that cross builds of Go container images work.
# Because we don't use cgo, we don't actually need a C++ toolchain. It's
# just that Bazel's analysis doesn't know this.
toolchain(
    name = "empty_cc_toolchain_for_linux_on_macos",
    exec_compatible_with = ["@platforms//os:macos"],
    target_compatible_with = ["@platforms//os:linux"],
    toolchain = ":empty_cc_toolchain",
    toolchain_type = "@bazel_tools//tools/cpp:toolchain_type",
)
```

With this in WORKSPACE:

```
register_toolchains("//bazel/cpp:empty_cc_toolchain_for_linux_on_macos")
```

thesayyn: this can be done bb-storage repo specifically, but in order to implement transitions upstream we need a good foundation that doesn't require hacks like this at all.

thesayyn: rules_docker has/had issues with this and people are still trying to work around too many issues.

thesayyn: addressed the separate targets. now there's one image target per image.

tools/oci/go_image.bzl
thesayyn: @EdSchouten I can come back and remove this once we support transitions upstream.

alexeagle: To be clear, pkg_tar does support `include_runfiles` https://github.com/bazelbuild/rules_pkg/pull/398

PR #174, opened 2023-08-03
Add replication for missing blobs in readFallbackBlobAccess

annapst:
The changes are very similar to mirroredBlobAccess

### DIFF
--- a/pkg/blobstore/readfallback/read_fallback_blob_access.go
+++ b/pkg/blobstore/readfallback/read_fallback_blob_access.go
@@ -10,6 +10,7 @@ import (
 	"github.com/buildbarn/bb-storage/pkg/digest"
 	"github.com/buildbarn/bb-storage/pkg/util"
 
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
@@ -89,6 +90,24 @@ func (ba *readFallbackBlobAccess) FindMissing(ctx context.Context, digests diges
 	missingInBoth, err := ba.secondary.FindMissing(ctx, missingInPrimary)
 	if err != nil {
 		return digest.EmptySet, util.StatusWrap(err, "Secondary")
+	}
+
+	// Replicate the blobs that are present only in the secondary backend to primary
+	if ba.replicator != nil {
+		presentOnlyInSecondary, _, _ := digest.GetDifferenceAndIntersection(missingInPrimary, missingInBoth)
+		replicateGroup, replicateCtx := errgroup.WithContext(ctx)
+		replicateGroup.Go(func() error {
+			if err := ba.replicator.ReplicateMultiple(replicateCtx, presentOnlyInSecondary); err != nil {
+				if status.Code(err) == codes.NotFound {
+					return util.StatusWrapWithCode(err, codes.Internal, "Backend secondary returned inconsistent results while synchronizing")
+				}
+				return util.StatusWrap(err, "Failed to synchronize from backend secondary to backend primary")
+			}
+			return nil
+		})
+		if err := replicateGroup.Wait(); err != nil {
+			return digest.EmptySet, err
+		}
 	}
 	return missingInBoth, nil
 }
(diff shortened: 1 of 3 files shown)


pkg/blobstore/readfallback/read_fallback_blob_access.go
```diff
@@ -90,5 +91,23 @@ func (ba *readFallbackBlobAccess) FindMissing(ctx context.Context, digests diges
 	if err != nil {
 		return digest.EmptySet, util.StatusWrap(err, "Secondary")
 	}
+
+	// Replicate the blobs that are present only in the secondary backend to primary
+	if ba.replicator != nil {
```
EdSchouten, line 96: This if-statement can be removed. The Buildbarn codebase tries to follow the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern) where possible. So if you specify a 'noop' blob replicator, it's actually an instance of this type:

https://github.com/buildbarn/bb-storage/blob/master/pkg/blobstore/replication/noop_blob_replicator.go

pkg/blobstore/readfallback/read_fallback_blob_access.go
```diff
@@ -90,5 +91,23 @@ func (ba *readFallbackBlobAccess) FindMissing(ctx context.Context, digests diges
...
+		presentOnlyInSecondary, _, _ := digest.GetDifferenceAndIntersection(missingInPrimary, missingInBoth)
+		replicateGroup, replicateCtx := errgroup.WithContext(ctx)
```
EdSchouten, line 98: We generally only use errgroups in case parallelism is needed. In this specific case you're using the errgroup to only launch a single goroutine, meaning there's no gain here. Could you remove the use of errgroup here? You may just call ReplicateMultiple() directly.

pkg/blobstore/readfallback/read_fallback_blob_access_test.go
```diff
@@ -181,6 +183,8 @@ func TestReadFallbackBlobAccessFindMissing(t *testing.T) {
 			Return(missingFromPrimary, nil)
 		secondary.EXPECT().FindMissing(ctx, missingFromPrimary).
 			Return(missingFromBoth, nil)
+		replicator.EXPECT().ReplicateMultiple(gomock.Any(), presentOnlyInSecondary).
+		    Return(nil)
```
EdSchouten, line 187: If you want, you may omit `.Return(nil)` here. If no return values are provided, gomock will return default initialized values.

pkg/blobstore/readfallback/read_fallback_blob_access.go
```diff
@@ -90,5 +91,23 @@ func (ba *readFallbackBlobAccess) FindMissing(ctx context.Context, digests diges
 	if err != nil {
 		return digest.EmptySet, util.StatusWrap(err, "Secondary")
 	}
+
+	// Replicate the blobs that are present only in the secondary backend to primary
```
EdSchouten, line 95: Nit: Please end all sentences in comments with a full stop (dot), and make sure they are wrapped to at most 80 columns.

EdSchouten (commented):
Hey @annapst,
This change looks pretty good. Thanks for working on this. Be sure to process the comments I've left behind, and I'll merge this in a timely manner.

pkg/blobstore/readfallback/read_fallback_blob_access.go
annapst: I've added this check because of the description of a previous pull request touching this file, mentioning backwards compatibility (https://github.com/buildbarn/bb-storage/commit/47ea975ed072816cb5a2c74f9fe563113d1a9a76) and because there is a nil check in other places in this file.
Is this still safe to remove with this info?

pkg/blobstore/readfallback/read_fallback_blob_access.go
annapst: Good point, let me update!

pkg/blobstore/readfallback/read_fallback_blob_access.go
EdSchouten: It's still safe to remove it. What's in the commit message likely applied to the initial version of the PR. See this comment in the PR, where I requested the same thing:

https://github.com/buildbarn/bb-storage/pull/79#discussion_r468105709

annapst: Thank you for clarifying! Let me update!

EdSchouten (approved):
LGTM! Thanks! Will merge once the CI run completes and passes.

PR #177, opened 2023-09-06
Add mount functionality to 'Directory'

stagnation:
This is needed for 'chroot' runners that must mount the special filesystems '/proc' and '/sys' in the input root.
https://github.com/buildbarn/bb-remote-execution/issues/115

### DIFF
--- a/go_dependencies.bzl
+++ b/go_dependencies.bzl
@@ -1489,6 +1489,7 @@ def go_dependencies():
     go_repository(
         name = "org_golang_x_sys",
         importpath = "golang.org/x/sys",
+        patches = ["//:patches/org_golang_x_sys/golang-issue-59357.diff"],
         sum = "h1:Vz7Qs629MkJkGyHxUlRHizWJRG2j8fbQKjELVSNhy7Q=",
         version = "v0.14.0",
     )
--- a/pkg/filesystem/directory.go
+++ b/pkg/filesystem/directory.go
@@ -112,6 +112,10 @@ type Directory interface {
 	// Function that base types may use to implement calls that
 	// require double dispatching, such as hardlinking and renaming.
 	Apply(arg interface{}) error
+
+	// Mount and Unmount.
+	Mount(mountpoint path.Component, source, fstype string) error
+	Unmount(mountpoint path.Component) error
 }
 
 // DirectoryCloser is a Directory handle that can be released.
--- a/pkg/filesystem/local_directory_linux.go
+++ b/pkg/filesystem/local_directory_linux.go
@@ -8,6 +8,7 @@ import (
 	"runtime"
 
 	"github.com/buildbarn/bb-storage/pkg/filesystem/path"
+	"github.com/buildbarn/bb-storage/pkg/util"
 
 	"golang.org/x/sys/unix"
 	"google.golang.org/grpc/codes"
@@ -33,4 +34,39 @@ func (d *localDirectory) Mknod(name path.Component, perm os.FileMode, deviceNumb
 
 func clonefileImpl(oldFD int, oldName string, newFD int, newName string) error {
 	return status.Error(codes.Unimplemented, "Clonefile is only supported on Darwin")
+}
+
+func (d *localDirectory) Mount(mountpoint path.Component, source, fstype string) error {
+	mountname := mountpoint.String()
+	fd, err := unix.Fsopen(fstype, unix.FSOPEN_CLOEXEC)
+	if err != nil {
+		return util.StatusWrapf(err, "Fsopen '%s'", fstype)
+	}
+	defer unix.Close(fd)
+
+	err = unix.FsconfigSetString(fd, "source", source)
+	if err != nil {
+		return util.StatusWrapf(err, "Fsconfig source '%s'", source)
+	}
+
+	err = unix.FsconfigCreate(fd)
+	if err != nil {
+		return util.StatusWrap(err, "Fsconfig create")
+	}
+
+	mfd, err := unix.Fsmount(fd, unix.FSMOUNT_CLOEXEC, unix.MS_NOEXEC)
+	if err != nil {
+		return util.StatusWrap(err, "Fsmount")
+	}
+	// NB: `Fsmount` creates a file descriptor to the mount object, that can be
+	// used to move it again. But we will not do so, so it is best to close it.
+	// Unmount will fail with `EBUSY` if it is left open.
+	defer unix.Close(mfd)
+
+	err = unix.MoveMount(mfd, "", d.fd, mountname, unix.MOVE_MOUNT_F_EMPTY_PATH)
+	if err != nil {
+		return util.StatusWrapf(err, "Movemount mountname '%s' in file descriptor %d", mountname, d.fd)
+	}
+
+	return nil
 }
--- a/pkg/filesystem/local_directory_unix.go
+++ b/pkg/filesystem/local_directory_unix.go
@@ -268,7 +268,7 @@ func (d *localDirectory) Remove(name path.Component) error {
 
 var workingDirectoryLock sync.Mutex
 
-func (d *localDirectory) unmount(name path.Component) error {
+func (d *localDirectory) Unmount(name path.Component) error {
 	defer runtime.KeepAlive(d)
 
 	// POSIX systems provide no umountat() system call that permits
@@ -302,7 +302,7 @@ func (d *localDirectory) removeAllChildren(parentDeviceNumber rawDeviceNumber) e
 		// unmount until the remaining directory is on the same
 		// file system.
 		for parentDeviceNumber != childDeviceNumber {
-			if err := d.unmount(component); err != nil {
+			if err := d.Unmount(component); err != nil {
 				return err
 			}
 			fileType, childDeviceNumber, _, err = d.lstat(component)
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,6 +1,6 @@
 load("@bazel_gazelle//:def.bzl", "gazelle")
 load("@com_github_bazelbuild_buildtools//buildifier:def.bzl", "buildifier")
-load("@npm//:defs.bzl", "npm_link_all_packages", "npm_link_targets")
+load("@npm//:defs.bzl", "npm_link_all_packages")
 
 # gazelle:prefix github.com/buildbarn/bb-storage
 # gazelle:resolve proto build/bazel/remote/execution/v2/remote_execution.proto @com_github_bazelbuild_remote_apis//build/bazel/remote/execution/v2:remote_execution_proto
--- a/patches/org_golang_x_sys/golang-issue-59357.diff
+++ b/patches/org_golang_x_sys/golang-issue-59357.diff
@@ -1,0 +1,181 @@
+From 9313951415ca77dcf5e84ba5fa41483b5c16d121 Mon Sep 17 00:00:00 2001
+From: Ilya Hanov <ilya.hanov@huawei-partners.com>
+Date: Sat, 30 Sep 2023 12:02:48 +0300
+Subject: [PATCH] unix: add API for fsconfig system call
+
+Fixes golang/go#59537
+
+Change-Id: I8d806ace3adad423c633813455d8f758706cee1d
+---
+
+diff --git unix/linux/types.go unix/linux/types.go
+index 25b9279..b9419f3 100644
+--- unix/linux/types.go
++++ unix/linux/types.go
+@@ -972,6 +972,15 @@
+ 	FSPICK_EMPTY_PATH       = C.FSPICK_EMPTY_PATH
+ 
+ 	FSMOUNT_CLOEXEC = C.FSMOUNT_CLOEXEC
++
++	FSCONFIG_SET_FLAG        = C.FSCONFIG_SET_FLAG
++	FSCONFIG_SET_STRING      = C.FSCONFIG_SET_STRING
++	FSCONFIG_SET_BINARY      = C.FSCONFIG_SET_BINARY
++	FSCONFIG_SET_PATH        = C.FSCONFIG_SET_PATH
++	FSCONFIG_SET_PATH_EMPTY  = C.FSCONFIG_SET_PATH_EMPTY
++	FSCONFIG_SET_FD          = C.FSCONFIG_SET_FD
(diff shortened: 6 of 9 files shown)


pkg/filesystem/local_directory_unix.go
```diff
@@ -460,3 +462,50 @@ func (d *localDirectory) Apply(arg interface{}) error {
...
+		return util.StatusWrap(err, "Mountat")
+	}
+	defer unix.Close(mfd)
+
+	return nil
+}
+
+// Fchdir changes the current working directory to the directory file descriptor.
+// If there is an error, it will be of type *PathError.
+func Fchdir(dfd int) error {
```
stagnation, line 478: This is a copy of `unix.Chdir` without the logging. As `Fchdir` does not have a wrapper function.

pkg/filesystem/mount_unix.go
```diff
@@ -0,0 +1,129 @@
...
+
+// Mounts the `source` filesystem on `mountname` inside a directory
+// given as `dfd` file descriptor,
+// using the `fstype` filesystem type.
+// This returns a file descriptor to the mount object,
+// that can be used to move it again.
+// Remember to close it before unmounting,
+// or unmount will fail with EBUSY.
+func mountat(dfd int, fstype, source, mountname string, options int) (int, error) {
+	_ = options // TODO: how to best send these to the various calls?
```
stagnation, line 102: This is a suboptimal compromise, it should either be handled or removed from the function API.
At the very least assert that no options are set.

:-1:

pkg/filesystem/mount_unix.go
```diff
@@ -0,0 +1,129 @@
...
+// Rudimentary function wrapper for the `fsconfig` syscall.
+//
+// This is implemented as a stop gap solution until real support is merged into the `unix` library.
+// See this patchset: https://go-review.googlesource.com/c/sys/+/399995/ .
+// This only implements the two commands needed for basic `mountat` functionality.
+// And will just exit if any other command is called.
+//
+// TODO(nils): construct proper `syscall.E*` errors.
+//	like `unix.errnoErr`, but the function is not exported.
+func fsconfig(fsfd int, cmd int, key string, value string, flags int) (err error) {
```
stagnation, line 36: Unsure about how to handle this best, these new syscalls are not used very often.
That they are not merged to the `unix` package is an unfortunate consequence for us.
It is also possible to patch `unix`, with the patch from go-review, but I do not know if that is better...

pkg/filesystem/BUILD.bazel
```diff
@@ -13,6 +13,7 @@ go_library(
         "local_directory_linux.go",
         "local_directory_unix.go",
         "local_directory_windows.go",
+        "mount_unix.go",
```
stagnation, line 16: Wanted to move the syscall cruff out of the way. We should not need so much code.

stagnation (commented):
FTR: I can not request changes on my own PR.

pkg/filesystem/directory.go
```diff
@@ -112,6 +113,14 @@ type Directory interface {
...
 	// require double dispatching, such as hardlinking and renaming.
 	Apply(arg interface{}) error
+
+	// Mount and Unmount.
+	// This uses `Mountat` functionality, but not `Unmountat`.
+	// see https://github.com/buildbarn/bb-remote-execution/issues/115 for information.
+	Mount(mountpoint path.Component, source string, fstype string, options int) error
+	// The Unmount call changes the working directory internally
+	// and must use a program-scope mutex to isolate this side effect.
+	Unmount(lock *sync.Mutex, mountpoint path.Component) error
```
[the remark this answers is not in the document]
EdSchouten: I think the mutex can just be process wide. It's perfectly fine to create a `local_directory_linux.go` that contains a global mutex for that.

We'd also need something like this if we ever wanted to support connecting/binding to a UNIX socket through this API. There are also no portable interfaces to do that `*at()` style (except on FreeBSD, which has `bindat` and `connectat`).

It's just an implementation detail that on Linux we require the use of a mutex here.

pkg/filesystem/local_directory_unix.go
```diff
@@ -460,3 +463,53 @@ func (d *localDirectory) Apply(arg interface{}) error {
 		return syscall.EXDEV
 	}
 }
+
+func (d *localDirectory) Mount(mountpoint path.Component, source string, fstype string, options int) error {
+	if options != 0 {
```
EdSchouten, line 468: There is no strict requirement to make the API 'complete'. It needs to be complete enough for our use case. Be sure to just remove the `options` parameter.

pkg/filesystem/local_directory_unix.go
EdSchouten: Do we need this function at all? Why don't we just call `Fchdir()` directly?

pkg/filesystem/local_directory_unix.go
```diff
@@ -460,3 +463,53 @@ func (d *localDirectory) Apply(arg interface{}) error {
...
+// Uses `unmount` on relative pathnames with `fchdir`.
+func fchdir_unmountat(lock *sync.Mutex, dfd int, mountname string) error {
+	lock.Lock()
+	defer lock.Unlock()
+
+	store, err := os.Getwd()
+	if err != nil {
+		return util.StatusWrap(err, "Could not store the working directory %#v.")
+	}
+	defer os.Chdir(store)
```
EdSchouten, line 498: I'd say, don't even bother restoring the working directory. If other pieces of code depend on doing relative path lookups, then they will be broken anyway because they race with this code.

That said, I can imagine it is desirable to leave the FUSE/NFSv4 file system, so that it doesn't remain in use permanently. What are your thoughts on doing an `os.Chdir("/")` instead?

pkg/filesystem/local_directory_unix.go
```diff
@@ -460,3 +463,53 @@ func (d *localDirectory) Apply(arg interface{}) error {
...
+}
+
+// Uses `unmount` on relative pathnames with `fchdir`.
+func fchdir_unmountat(lock *sync.Mutex, dfd int, mountname string) error {
+	lock.Lock()
+	defer lock.Unlock()
+
+	store, err := os.Getwd()
+	if err != nil {
+		return util.StatusWrap(err, "Could not store the working directory %#v.")
```
EdSchouten, line 496: Nit: Don't add a "." to the end of error strings. It's inconsistent with the rest of the codebase.

pkg/filesystem/mount_unix.go
```diff
@@ -0,0 +1,130 @@
...
+	"fmt"
+	"unsafe"
+
+	"github.com/buildbarn/bb-storage/pkg/util"
+
+	"golang.org/x/sys/unix"
+)
+
+const (
+	FSCONFIG_SET_FLAG        = 0
```
EdSchouten, line 17: These are constants from the Linux kernel, right? If so, could you please submit a PR against `golang.org/x/sys/unix`'s bindings generator to auto-extract those?

pkg/filesystem/mount_unix.go
```diff
@@ -0,0 +1,130 @@
...
+
+	_p1, err = unix.BytePtrFromString(value)
+	if err != nil {
+		return
+	}
+	if value == "" {
+		_p1 = nil
+	}
+
+	r0, _, e1 := unix.Syscall6(
```
EdSchouten, line 72: The same holds for `fsconfig()`. We could also let `golang.org/x/sys/unix` ship with a binding for that. My experience is that the folks maintaining that package are willing to accept such changes. They also review them in a timely manner.

pkg/filesystem/mount_unix.go
```diff
@@ -0,0 +1,130 @@
+//go:build darwin || freebsd || linux
```
EdSchouten, line 1: I would call this file `mount_linux.go`, because `fsconfig()` is Linux specific. Other operating systems could have a `mount_unsupported.go` or something.

pkg/filesystem/local_directory_unix.go
```diff
@@ -460,3 +463,53 @@ func (d *localDirectory) Apply(arg interface{}) error {
...
+
+	err = unix.Unmount(mountname, 0)
+	if err != nil {
+		return util.StatusWrap(err, "Unmount")
+	}
+
+	return nil
+}
+
+func (d *localDirectory) Unmount(lock *sync.Mutex, mountpoint path.Component) error {
```
EdSchouten, line 513: I'd say, just keep this function in `mount_linux.go` as well. It's perfectly fine to have `localDirectory`'s `Unmount()` and `Mount()` methods in another file, without having all of the indirection.

stagnation: Okay, split it into `mount_{linux,unsupported}`, with the build tags for Linux and all other operating systems respectively.

pkg/filesystem/local_directory_unix.go
stagnation: I feel a bit uneasy about leaving it in whereverland, but going to the root is fine.
When I debugged this code path it seems to live in /tmp and goes back there after unmounting.

pkg/filesystem/directory.go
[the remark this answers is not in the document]
stagnation: Okay! That's a good way to go about it. It cannot be a struct var, but package/file global is okay. Nice to move it from the API.

pkg/filesystem/local_directory_linux.go
```diff
@@ -17,6 +22,10 @@ import (
 // rawDeviceNumber is the equivalent of POSIX dev_t.
 type rawDeviceNumber = uint64
 
+var (
```
EdSchouten, line 25: Maybe best to call this as follows?
```
var workingDirectoryLock sync.Mutex
```
Be sure to also add a comment on top, explaining why this exists.

pkg/filesystem/local_directory_linux.go
```diff
@@ -34,3 +43,32 @@ func (d *localDirectory) Mknod(name path.Component, perm os.FileMode, deviceNumb
...
+
+	store, err := os.Getwd()
+	if err != nil {
+		return util.StatusWrap(err, "Could not store the working directory %#v.")
+	}
+	defer os.Chdir(store)
+
+	err = syscall.Fchdir(dfd)
+	if err != nil {
+		err = &fs.PathError{Op: "fchdir", Path: fmt.Sprintf("%d", dfd), Err: err}
```
EdSchouten, line 60: Don't bother wrapping the error with a PathError. Also including the file descriptor number in the error message is a bit too low-level and also inconsistent with the other methods. This is sufficient:
```
return util.StatusWrap(err, "Could not change working directory to parent of mount")
```

pkg/filesystem/local_directory_linux.go
```diff
@@ -34,3 +43,32 @@ func (d *localDirectory) Mknod(name path.Component, perm os.FileMode, deviceNumb
...
+	err = unix.Unmount(mountname, 0)
+	if err != nil {
+		return util.StatusWrap(err, "Unmount")
+	}
+
+	return nil
+}
+
+func (d *localDirectory) Unmount(mountpoint path.Component) error {
+	return fchdir_unmountat(d.fd, mountpoint.String())
```
EdSchouten, line 73: We only use `fchdir_unmountat()` in one place. Be sure to just inline `Unmount()` into it.

Shouldn't you use `defer runtime.KeepAlive(d)` here as well, by the way? Otherwise you're at risk of having the directory closed before the `Unmount()` system call has had a chance to run.

pkg/filesystem/mount_unix.go
stagnation: This is now the packaged as a patch for the 'sys' package.
https://go-review.googlesource.com/c/sys/+/399995/6

pkg/filesystem/mount_unix.go
stagnation: I have updated the commit with the patch file, and will wait a couple more days for them to see my comment on the original review before forking it.

pkg/filesystem/local_directory_unix.go
stagnation: Using the old `unmount` now that does not attempt to return.

pkg/filesystem/directory.go
```diff
@@ -112,6 +112,12 @@ type Directory interface {
 	// Function that base types may use to implement calls that
 	// require double dispatching, such as hardlinking and renaming.
 	Apply(arg interface{}) error
+
+	// Mount and Unmount.
+	// This uses `Mountat` functionality, but not `Unmountat`.
+	// see https://github.com/buildbarn/bb-remote-execution/issues/115 for information.
```
EdSchouten, line 118: I'd remove this comment. This is an implementation detail of the Linux/UNIX version of this method, but not a general requirement/property of how this interface is implemented.

go_dependencies.bzl
```diff
@@ -1569,6 +1569,7 @@ def go_dependencies():
         name = "org_golang_x_sys",
         importpath = "golang.org/x/sys",
         sum = "h1:SqMFp9UcQJZa+pmYuAKjd9xq1f0j5rLcDIk0mj4qAsA=",
+        patches = ["//:patches/org_golang_x_sys/unix_add_Fsconfig_syscall_on_linux.patch"],
```
EdSchouten, line 1572: Nits:
- Please name the diff file .diff instead of .patch. That's also how the other files in that directory are called.
- I think your PR doesn't actually include this file.
- Could you please put the number of the upstream PR in the name of the diff file? That way it's easier to find going forward.

Have you also forked the upstream PR as discussed?

pkg/filesystem/mount_unix.go
stagnation: Pushed a new patch, with the new proposed API.

go_dependencies.bzl
stagnation: 1:  done
2:  done
3:  done

4: No, but I found that IlyaHanov started the process: https://github.com/golang/go/issues/59537

pkg/filesystem/mount_unix.go
stagnation: There has not been any progress on the review since we last discussed the patch :/
https://github.com/golang/go/issues/59537#issuecomment-1752217786

EdSchouten (approved):

EdSchouten (approved):
Looks good! Will merge when CI is happy.

PR #178, opened 2023-09-13
Add Buildbarn logo svg

moroten:
Supplying two logos, one with rounded ends on the lines but then also, unfortunately, manually edited after Drawio export.

I like the one with rounded line caps best but feel free to choose the one you prefer. Where shall it be integrated into the docs? Should there be a `bb-docs` for GitHub pages or should it be put in `bb-deployments`?


EdSchouten:
Hey! That's awesome. I think I like the rounded one most as well.

I think that having a `bb-docs` repo only makes sense if we also commit to writing actual documentation. If this is not going to happen in the nearby future, maybe best to just keep it here?

moroten:
I just realised that the [bb-browser favicon](https://github.com/buildbarn/bb-browser/blob/881fd822f75e640f89dcf6b7143259459f106783/cmd/bb_browser/favicon.png) is also a good alternative, maybe even better.

moroten:
I'm happy with this update now. Will be integrated in https://github.com/buildbarn/bb-browser/pull/46.

EdSchouten:
Considering that we likely want to use this all over the place (pkg/otel, bb-scheduler, bb-browser, ...), what are your thoughts on just having a pkg/logo in this repo that uses `//go:embed` to pull it in? That way we can easily reference it from within Go code without further trickery.

moroten:
Nice idea, Ed! Updated the commit accordingly.

pkg/logo/logo.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	_ "embed" // For "go:embed".
+)
+
+//go:embed favicon.svg
+var faviconSvg []byte
+
+// FaviconSvg holds the Buildbarn logo and can be used as favicon in web browsers.
+var FaviconSvg = faviconSvg
```
EdSchouten, line 11: Out of curiosity, why the indirection? We can't make the `go:embed` variable public? I think we should also call this `FaviconSVG`, as in Go they tend to write abbreviations like HTTP, XML, etc. in all caps.

EdSchouten (commented):
Can I persuade you to patch up `pkg/otel/active_spans.html` to make use of this?

moroten:
> Can I persuade you to patch up `pkg/otel/active_spans.html` to make use of this?

I tried but failed to encode the favicon in the otel test. What's the best way forward?

PR #180, opened 2023-10-06
JWT: Enable reading a JSON Web Key Set from a file

mortenmj:
This changes the JWKS configuration to be one of either inline content (as before) or a reference to a file.

To enable this, a new signature validator is added, the ForwardingSignatureValidator, which holds a pointer to some other validator that requests to ValidateSignature are forwarded to. We can replace this validator by calling Replace(). When periodically reading the content at the JWKS file path, this internal validator is updated to one containing the new JWKS content.


pkg/jwt/configuration.go
```diff
@@ -102,3 +115,38 @@ func NewSignatureValidatorFromJSONWebKeySet(jwks *jose.JSONWebKeySet) (Signature
 
 	return NewDemultiplexingSignatureValidator(namedSignatureValidators, allSignatureValidators), nil
 }
+
+func NewSignatureValidatorFromJSONWebKeySetFile(path string, refreshInterval time.Duration) SignatureValidator {
+	// Hmm. I don't want us to first read the file once, to initialize this, and then read the file periodically to update it.
```
EdSchouten, line 120: Why wouldn't you want to do this? Doing a forced read upon startup seems like the most sensible thing to do. This ensures that if you get the pathname wrong in your config, your containers will crashloop and prevent a rolling upgrade to complete.

pkg/jwt/forwarding_signature_validator.go
```diff
@@ -0,0 +1,26 @@
+package jwt
+
+import (
+	"sync/atomic"
+)
+
+type forwardingSignatureValidator struct {
```
EdSchouten, line 7: Considering that this struct's public API is greater than SignatureValidator (i.e., there is a Replace() method), this should just be a public type.

pkg/proto/configuration/jwt/jwt.proto
```diff
@@ -14,17 +15,31 @@ message AuthorizationHeaderParserConfiguration {
...
-  // - https://smallstep.com/docs/step-cli/reference/crypto/key/format/
-  // - https://irrte.ch/jwt-js-decode/pem2jwk.html
-  // - https://russelldavies.github.io/jwk-creator/
-  google.protobuf.Struct jwks_inline = 7;
+  message JwksFile {
+    // Path to JWKS file.
+    string file_path = 1;
+
+    // Interval at which to refresh JWKS file.
+    google.protobuf.Duration refresh_interval = 2;
```
EdSchouten, line 23: Maybe having a refresh interval adds too much flexibility. Is there some kind of constant that we can use here that tends to be sufficient for most use cases? 5 minutes? If so, we should just use that.

pkg/jwt/configuration.go
```diff
@@ -102,3 +115,38 @@ func NewSignatureValidatorFromJSONWebKeySet(jwks *jose.JSONWebKeySet) (Signature
...
+	// However, I also don't want us to initialize this in a state that isn't ready for use.
+	validator := NewForwardingSignatureValidator(nil)
+
+	// TODO: Run this as part of the program.Group, so that it gets
```
EdSchouten, line 124: Just to confirm: indeed, any new code should be written to run as part of program.Group.

pkg/proto/configuration/jwt/jwt.proto
mortenmj: If I we're gonna set a static value, 300 seconds would have been my choice as well

pkg/jwt/configuration.go
mortenmj: It seemed a little inelegant to have two different code paths for the same thing, so my idea was to find some way to ensure that it would never be used before the first call to Replace(). I agree it does make sense to catch errors before we go into our goroutine, though. I've updated the code to do this.

pkg/jwt/forwarding_signature_validator.go
mortenmj: Done. I would normally have preferred to return the struct, and not the interface, but that seemed not to be the common pattern here, at least not in the jwt package. I've updated this to return the ForwardingSignatureValidator struct type.

pkg/jwt/configuration.go
mortenmj: Updated this. Had to touch a few different places to pass the program.Group all the way down.

pkg/proto/configuration/jwt/jwt.proto
mortenmj: Since we're not keeping this field, I've turned `jwks_file` into a string type. The refresh rate is a static 300 seconds.

pkg/global/apply_configuration.go
```diff
@@ -70,16 +70,10 @@ func (ls *LifecycleState) MarkReadyAndWait(group program.Group) {
...
-			if err := bb_http.NewServersFromConfigurationAndServe(
-				ls.config.HttpServers,
-				bb_http.NewMetricsHandler(router, "Diagnostics"),
-				group,
-			); err != nil {
-				return util.StatusWrap(err, "Failed to launch diagnostics HTTP server")
-			}
-			return nil
-		})
+		bb_http.NewServersFromConfigurationAndServe(
```
EdSchouten, line 73: Out of curiosity, what's the reason we're now chaging this function not to return an error? Doesn't this cause error messages to become less specific?

pkg/jwt/configuration.go
```diff
@@ -102,3 +120,52 @@ func NewSignatureValidatorFromJSONWebKeySet(jwks *jose.JSONWebKeySet) (Signature
...
+				return util.StatusFromContext(ctx)
+			}
+		}
+	})
+
+	return forwardingValidator, nil
+}
+
+func getJwksFromFile(path string) (SignatureValidator, error) {
+	r, err := os.Open(path)
```
EdSchouten, line 159: This file isn't getting closed.

pkg/jwt/configuration.go
```diff
@@ -102,3 +120,52 @@ func NewSignatureValidatorFromJSONWebKeySet(jwks *jose.JSONWebKeySet) (Signature
...
+			case <-ctx.Done():
+				return util.StatusFromContext(ctx)
+			}
+		}
+	})
+
+	return forwardingValidator, nil
+}
+
+func getJwksFromFile(path string) (SignatureValidator, error) {
```
EdSchouten, line 158: getJWKSFromFile()

pkg/jwt/configuration.go
```diff
@@ -102,3 +120,52 @@ func NewSignatureValidatorFromJSONWebKeySet(jwks *jose.JSONWebKeySet) (Signature
...
+	r, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+
+	var jwks jose.JSONWebKeySet
+	err = json.NewDecoder(r).Decode(&jwks)
```
EdSchouten, line 165: One interesting fact about `json.NewDecoder().Decode()` I got told about the other day is that it:
- still ends up loading all JSON in memory up front. There is thus no savings in memory usage
- permits trailing garbage, as it simply stops parsing after the top-level object is completed


With that in mind, I guess we should just `io.ReadAll()` it?

pkg/global/apply_configuration.go
EdSchouten: Oh wait. That's obviously because we now run in a separate goroutine entirely.

pkg/jwt/configuration.go
```diff
@@ -102,3 +120,52 @@ func NewSignatureValidatorFromJSONWebKeySet(jwks *jose.JSONWebKeySet) (Signature
...
+	group.Go(func(ctx context.Context, siblingsGroup, dependenciesGroup program.Group) error {
+		t := time.NewTicker(300 * time.Second)
+		defer t.Stop()
+
+		for {
+			select {
+			case <-t.C:
+				internalValidator, err := getJwksFromFile(path)
+				if err != nil {
+					log.Printf("Failed to read JWKS content from file at %s: %s", path, err)
```
EdSchouten, line 144: For the first `%s`, could you please use `%#v`? That way the string ends up getting quoted. Same the `status.Errorf()` above.

pkg/jwt/configuration.go
```diff
@@ -102,3 +120,52 @@ func NewSignatureValidatorFromJSONWebKeySet(jwks *jose.JSONWebKeySet) (Signature
...
 }
+
+// NewSignatureValidatorFromJSONWebKeySetFile creates a new
+// SignatureValidator capable of validating JWTs matching keys contained
+// in a JSON Web Key Set read from a file. The content of the file is
+// periodically refreshed.
+func NewSignatureValidatorFromJSONWebKeySetFile(path string, group program.Group) (SignatureValidator, error) {
+	internalValidator, err := getJwksFromFile(path)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Unable to read JWKS content from file at %s", path)
```
EdSchouten, line 131: `util.StatusWrapf(...)`

EdSchouten (commented):
This is starting to look great. Final round of comments and this should be good to land. Thanks!

pkg/jwt/configuration.go
mortenmj: Should this replace the status.Errorf or wrap it?

EdSchouten: It may replace it.

mortenmj:
@EdSchouten any chance we could get this merged this week?

EdSchouten (commented):
Hey there! Sorry for letting you wait. I overlooked the fact that you made another round of changes. One small remaining request and this is good to merge. Thanks!

mortenmj:
@EdSchouten ping

PR #184, opened 2023-12-20
Feature/upgrade gazelle

stagnation:


patches/bazel_gazelle/dont-flatten-srcs.diff
```diff
@@ -20,16 +20,16 @@ index 0b23cca..c532ad2 100644
  	return goBinary
```
EdSchouten, line 20: Good news is that at some point in the future we may remove this patch:

https://github.com/bazelbuild/rules_go/pull/3785

EdSchouten (approved):
Will merge as soon as CI is happy.

PR #187, opened 2024-02-07
Tidy up

stagnation:
Some small changes to I made during the work to update windows workers.
Keep them as a separate PR to merge them sooner, and reduce the scope of the meaty review of the PR.

### DIFF
--- a/pkg/filesystem/local_directory_windows.go
+++ b/pkg/filesystem/local_directory_windows.go
@@ -873,7 +873,7 @@ func createNTFSHardlink(oldHandle windows.Handle, oldName string, newHandle wind
 func renameHelper(sourceHandle, newHandle windows.Handle, newName string) (areSame bool, err error) {
 	// We want to know a few things before renaming:
 	//  1. Are source and target hard links to the same file? If so, noop.
-	//  2. If target exists and wither source or target is a directory, don't overwrite and report error.
+	//  2. If target exists and whether source or target is a directory, don't overwrite and report error.
 	//  3. If neither is the case, move and, if necessary, replace.
 	var targetHandle windows.Handle
 	err = ntCreateFile(&targetHandle, windows.FILE_READ_ATTRIBUTES, newHandle, newName, windows.FILE_OPEN,
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -33,7 +33,7 @@
             },
             {
                "name": "Protobuf generation",
-               "run": "find . bazel-bin/pkg/proto -name '*.pb.go' -delete || true\nbazel build $(bazel query 'kind(\"go_proto_library\", //...)')\nfind bazel-bin/pkg/proto -name '*.pb.go' | while read f; do\n  cat $f > $(echo $f | sed -e 's|.*/pkg/proto/|pkg/proto/|')\ndone\n"
+               "run": "find . bazel-bin/pkg/proto -name '*.pb.go' -delete || true\nbazel build $(bazel query --output=label 'kind(\"go_proto_library\", //...)')\nfind bazel-bin/pkg/proto -name '*.pb.go' | while read f; do\n  cat $f > $(echo $f | sed -e 's|.*/pkg/proto/|pkg/proto/|')\ndone\n"
             },
             {
                "name": "Embedded asset generation",
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -33,7 +33,7 @@
             },
             {
                "name": "Protobuf generation",
-               "run": "find . bazel-bin/pkg/proto -name '*.pb.go' -delete || true\nbazel build $(bazel query 'kind(\"go_proto_library\", //...)')\nfind bazel-bin/pkg/proto -name '*.pb.go' | while read f; do\n  cat $f > $(echo $f | sed -e 's|.*/pkg/proto/|pkg/proto/|')\ndone\n"
+               "run": "find . bazel-bin/pkg/proto -name '*.pb.go' -delete || true\nbazel build $(bazel query --output=label 'kind(\"go_proto_library\", //...)')\nfind bazel-bin/pkg/proto -name '*.pb.go' | while read f; do\n  cat $f > $(echo $f | sed -e 's|.*/pkg/proto/|pkg/proto/|')\ndone\n"
             },
             {
                "name": "Embedded asset generation",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -163,9 +163,9 @@ http_archive(
 
 http_archive(
     name = "aspect_rules_js",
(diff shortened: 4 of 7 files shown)


pkg/filesystem/local_directory_windows.go
```diff
@@ -606,7 +607,7 @@ func (d *localDirectory) RemoveAllChildren() error {
 			err = subdirectory.RemoveAllChildren()
 			subdirectory.Close()
 			if err != nil {
-				return err
+				return util.StatusWrapf(err, "%s: ", name)
```
EdSchouten, line 610: Using `util.StatusWrap()` recursively might be a bit wacky, because it means you end up getting error messages along these lines:
```
src: myproject: mylibrary: mypackage: Failed to remove bob.c
```
Maybe the better thing to do is to make use of pkg/filesystem/path's Trace type here?
```go
func (d *localDirectory) RemoveAllChildren() error {
	return d.removeAllChildren(nil)
}

func (d *localDirectory) RemoveAllChildren(dPath *path.Trace) error {
	names, err := readdirnames(d.handle)
	if err != nil {
		return util.StatusWrapf(err, "Failed to read contents of directory %#v", dPath.String())
	}
	...
	for _, name := range names {
		childPath := dPath.Append(name)
		...
		err = subdirectory.removeAllChildren(childPath)
		...
```

EdSchouten (commented):
Thanks!

pkg/filesystem/local_directory_windows.go
```diff
@@ -108,10 +108,34 @@ func newLocalDirectory(absPath string, openReparsePoint bool) (DirectoryCloser,
...
 }
 
+// Convert forward-slash drive-letter paths to absolute drive-letter paths.
+// This can come from how build directories are configured for bb-worker on Windows.
+// Where a combination of 'git-bash' and environment variable expansion in jsonnet can create paths
+// that the `filepath` does not detect as absolute.
+//
+// Example: /C:/...
+// Should be C:/...
+func CanonicalizeAbsolutePath(path string) string {
```
EdSchouten, line 118: I don't fully understand this part. `NewLocalDirectory()` be called with whatever is the native representation of a path on that operating system. So I guess that for Windows this means that you should just use drive names and backslashes as part of your config files...?

pkg/filesystem/path/resolve.go
```diff
@@ -38,6 +39,12 @@ func (rs *resolverState) push(scopeWalker ScopeWalker, path string) error {
 		path = stripOneOrMoreSlashes(path)
 		absolute = true
 	}
+	if runtime.GOOS == "windows" {
```
EdSchouten, line 42: Please don't make changes along these lines.

Keep in mind that `pkg/filesystem/path` is supposed to be a platform agnostic path resolution library. Even though nobody should do anything as crazy as this, in theory it should be supported to run bb_worker on a Linux box, expose its build directory over a Samba server to a Windows box, and have bb_runner running there. Or the opposite.

`path.Resolve()` should just implement an abstract path resolver that does the same thing on every platform. If this is insufficient, then we should have discussions how to fix that up. Maybe we should rename it to `path.ResolveUNIX()`, and have a separate `path.ResolveWindows()`?

Taking a step back: are there actually any places where we call `path.Resolve()` against a Windows style path that contains drive letters? As far as I know, the only field in all of REv2 that contains a Windows style path that we need to process is `argv[0]`.

pkg/filesystem/local_directory_windows.go
```diff
@@ -584,44 +585,50 @@ func (d *localDirectory) Remove(name path.Component) error {
...
 	if err != nil {
-		return err
+		return util.StatusWrapf(err, "Failed to read contents of directory %#v", dPath.String())
 	}
 	for _, name := range names {
 		component := path.MustNewComponent(name)
 		fileType, err := d.lstat(component)
 		if err != nil {
-			return err
+			return util.StatusWrapf(err, "Failed to stat component %#v in %#v", name, dPath.String())
```
EdSchouten, line 604: Could you please use `childPath` in all of these error messages?

pkg/filesystem/local_directory_windows.go
```diff
@@ -584,44 +585,50 @@ func (d *localDirectory) Remove(name path.Component) error {
 	return err
 }
 
-// On NTFS mount point is a reparse point, no need to unmount.
 func (d *localDirectory) RemoveAllChildren() error {
+	return d.RemoveAllChildren(nil)
+}
+
+// On NTFS mount point is a reparse point, no need to unmount.
+func (d *localDirectory) RemoveAllChildren(dPath *path.Trace) error {
```
EdSchouten, line 593: Does this even compile? It has the same name as the function above. Please make it lowercase, so that it's private.

pkg/filesystem/local_directory_windows.go
```diff
@@ -584,44 +585,50 @@ func (d *localDirectory) Remove(name path.Component) error {
...
-			return err
+			return util.StatusWrapf(err, "Failed to remove component %#v in %#v", name, dPath.String())
 		}
 	}
 	return nil
 }
 
 func (d *localDirectory) RemoveAll(name path.Component) error {
 	defer runtime.KeepAlive(d)
+	dPath = &path.Trace{}
```
EdSchouten, line 628: Just use `nil`. `nil` indicates the empty path (well, `"."`).

EdSchouten (changes requested):

pkg/filesystem/local_directory_windows.go
stagnation: Check :). I assumed it needed a kernel to start with, but the test looks to work with nil. And your listing seeds it with nil, rather than being a backward-compatible not-tracer.

pkg/filesystem/local_directory_windows.go
stagnation: Yeah, no. I forgot to go to the windows computer.

pkg/filesystem/local_directory_windows.go
stagnation: I think I did this, but it should not be part of this PR. This is just meant to do some simple housekeeping stuff. So I can merge a couple of commits.

pkg/filesystem/path/resolve.go
stagnation: Handled much better in other commits and PRs. It was never meant for this PR.

pkg/filesystem/local_directory_windows.go
stagnation: Done, in subsequent work.

pkg/filesystem/local_directory_windows.go
stagnation: Done, in subsequent work.

pkg/filesystem/local_directory_windows.go
stagnation: Done, in subsequent work.

pkg/filesystem/local_directory_windows.go
stagnation: No longer relevant

PR #190, opened 2024-02-16
Feature/refactor path resolve

stagnation:
Two refactorings to prepare for resolving, parsing and building paths with drive letters on Windows.


pkg/filesystem/path/component_walker.go
```diff
@@ -49,17 +49,16 @@ type ComponentWalker interface {
...
 	//
 	// If the pathname component refers to a symbolic link, this
-	// function will return a GotSymlink containing a ScopeWalker,
-	// which can be used to perform expansion of the symbolic link.
-	// The Resolve() function will call into OnScope() to signal
-	// whether resolution should continue at the root directory or
-	// at the directory that contained the symbolic link.
+	// function will return a GotSymlink containing a ScopeWalker, which
+	// can be used to perform expansion of the symbolic link. The
+	// Resolve() function will call into OnAbsolute() or OnRealtive() to
```
EdSchouten, line 59: s/Realtive/Relative

pkg/filesystem/path/parser.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, status.Error(codes.InvalidArgument, "Path contains a null byte")
+	}
+
+	return &unixAbsoluteParser{path}, nil
+}
+
+func newRelativePath(path string) (unixRelativeParser, error) {
+	if len(path) > 0 && path[0] == '/' {
+		return unixRelativeParser{}, status.Errorf(codes.InvalidArgument, "Invalid relative path %#v", path)
+	}
```
EdSchouten, line 138: I'm not sure this error can actually occur. I mean, we only construct relative paths within this specific source file, and there we always make sure to strip slashes.

pkg/filesystem/path/parser.go
```diff
@@ -0,0 +1,150 @@
...
+
+	return unixRelativeParser{path}, nil
+}
+
+// MustNewParser is identical to NewUNIXParser, except that it panics
+// upon failure.
+func MustNewParser(path string) Parser {
```
EdSchouten, line 145: MustNewUNIXParser()?

pkg/filesystem/path/parser.go
```diff
@@ -0,0 +1,150 @@
...
+	if strings.ContainsRune(path, '\x00') {
+		panic("Invalid path")
+	}
+	return &unixAbsoluteParser{path}
```
EdSchouten, line 149: Please write this as:
```go
parser, err := NewUNIXParser(path)
if err != nil {
    panic(err)
}
return parser
```

pkg/filesystem/path/parser.go
```diff
@@ -0,0 +1,150 @@
...
+type Parser interface {
+	ParseScope(scopeWalker ScopeWalker) (next ComponentWalker, remainder RelativeParser, err error)
+}
+
+// RelativeParser is used by Resolve to parse relative paths in the resolution.
+type RelativeParser interface {
+	ParseFirstComponent(componentWalker ComponentWalker, mustBeDirectory bool) (next GotDirectoryOrSymlink, remainder RelativeParser, err error)
+}
+
+type unixAbsoluteParser struct {
```
EdSchouten, line 20: Maybe just call this unixParser, for consistency with Parser?

Do we want to place all of this code in a `unix_parser.go`, so that the Windows and UNIX implementations aren't all smacked in the same source file?

pkg/filesystem/path/resolve.go
```diff
@@ -1,133 +1,64 @@
...
+		r, err := rs.pop()
+		if err != nil {
+			return err
+		}
+		if r != nil {
+			switch rv := r.(type) {
+			case GotDirectory:
+				rs.componentWalker = rv.Child
+			case GotSymlink:
+				target, err := NewUNIXParser(rv.Target)
```
EdSchouten, line 53: As part of this PR, do we also want to make a followup change to `GotSymlink` to contain a `Target Parser` instead of a `Target string`?

pkg/filesystem/path/resolve.go
```diff
@@ -1,133 +1,64 @@
...
-		componentWalker, err = scopeWalker.OnAbsolute()
-	} else {
-		componentWalker, err = scopeWalker.OnRelative()
-	}
+	componentWalker, remainder, err := parser.ParseScope(scopeWalker)
 	if err != nil {
 		return err
 	}
-	rs.stack = append(rs.stack, path)
+	if remainder != nil {
```
EdSchouten, line 19: Maybe we can require that Parser.ParseScope() always returns a RelativePath? That way this code can just push it on the stack unconditionally.

EdSchouten (commented):
Great stuff!

pkg/filesystem/path/resolve.go
stagnation: Yeah, in the absence of an error there will always be a remainder. But the two nilable return vaules do not reflect this. As the code is written it is safe to drop the check, if that is desireable (I prefer the belt-and-suspenders). 
Or I could refactor the ParseScope function to just return everything with the named output arguments,
and then the error being there is all that matters. 

Something something sum-types

pkg/filesystem/path/parser.go
stagnation: Good point! Replaced with a bare construction.

pkg/filesystem/path/component_walker.go
stagnation: :eyes:

pkg/filesystem/path/parser.go
stagnation: Check!

pkg/filesystem/path/resolve.go
stagnation: Yepp, it is now in a third commit

pkg/filesystem/path/resolve.go
```diff
@@ -1,127 +1,60 @@
...
+	}
 	rs.componentWalker = componentWalker
 	return nil
 }
 
 // Pop a single filename from the stack of paths that need to be
 // processed. This filename is the first one that should be processed.
-func (rs *resolverState) pop() string {
+func (rs *resolverState) pop() (GotDirectoryOrSymlink, error) {
 	p := &rs.stack[len(rs.stack)-1]
```
EdSchouten, line 29: Considering that `rs.stack` is now a stack of interface objects, you can drop the `&`. This will make the call to `ParseFirstComponent()` cleaner.

Alternatively, just drop `p` entirely and inline it?

pkg/filesystem/path/resolve.go
```diff
@@ -1,127 +1,60 @@
...
-					return err
-				}
-				// Observed a symlink at the end of a
-				// path. We should continue to run.
-				if err := rs.push(r.Parent, r.Target); err != nil {
+		r, err := rs.pop()
+		if err != nil {
+			return err
+		}
+		if r != nil {
```
EdSchouten, line 48: Apparently it is permitted to use `case nil:` in `switch x.(type)`. Maybe use that, so that we can drop this if-statement?

pkg/filesystem/path/resolve_test.go
```diff
@@ -16,74 +16,74 @@ func TestResolve(t *testing.T) {
 	ctrl := gomock.NewController(t)
 
 	t.Run("NullByte", func(t *testing.T) {
-		scopeWalker := mock.NewMockScopeWalker(ctrl)
-
+		_, err := path.NewUNIXParser("hello\x00world")
```
EdSchouten, line 19: Technically speaking this test should be moved into unix_parser_test.go or something, but I have to say I don't care strongly.

pkg/filesystem/path/unix_parser.go
```diff
@@ -0,0 +1,129 @@
...
+func stripOneOrMoreSlashes(p string) string {
+	for {
+		p = p[1:]
+		if p == "" || p[0] != '/' {
+			return p
+		}
+	}
+}
+
+func (p *unixParser) ParseScope(scopeWalker ScopeWalker) (next ComponentWalker, remainder RelativeParser, err error) {
```
EdSchouten, line 27: I would change this to not use a pointer receiver. So `p unixParser`. Relatedly, I would advise against mutating `p.path`.

What are your thoughts on just writing this function like this?
```go
if p.path != "" && p.path[0] == '/' {
    next, err := scopeWalker.OnAbsolute()
    if err != nil {
        return nil, nil, err
    }
    return next, unixRelativeParser{stripOneOrMoreSlashes(p.path)}, nil
}

next, err := scopeWalker.OnRelative()
if err != nil {
    return nil, nil, err
}
return next, unixRelativeParser{p.path}, nil

**Review by EdSchouten — COMMENTED** [review]
This is shaping up nicely. I think we can merge this soon.

File: pkg/filesystem/path/unix_parser.go (hunk 11 above)
Comment (stagnation) [reply pkg/filesystem/path/unix_parser.go:27]: Looks good! It is annoying to try to create a variable for the strip function, as a var declaration and then use that at the very end.

File: pkg/filesystem/path/resolve.go (hunk 8 above)
Comment (stagnation) [reply pkg/filesystem/path/resolve.go:29]: I would prefer not to inline it, the next statement would become an eyesore.

File: pkg/filesystem/path/virtual_root_scope_walker_factory.go [hunk 12]
```diff
@@ -22,7 +22,7 @@ type namelessVirtualRootNode struct {
 // that point into the root directory.
 type namedVirtualRootNode struct {
 	nameless namelessVirtualRootNode
-	target   string
+	target   Parser
```
EdSchouten, line 25: By storing `Parser` objects inside these nodes, we end up handing them out repeatedly. I think this is fine right now, because we've implemented `UNIXParser` in such a way that it isn't destructive. That said, should we add some documentation to the `Parser` and `RelativeParser` interfaces to state that they MUST be implemented that way?

pkg/filesystem/path/resolve.go
```diff
@@ -47,18 +47,14 @@ func (rs *resolverState) resolve() error {
...
 				return err
 			}
-			if err := rs.push(rv.Parent, target); err != nil {
-				return err
-			}
-		default:
-			panic("Missing result")
+	default:
+		panic("Missing result")
 		case nil: // Do nothing
```
EdSchouten, line 55: I think `case nil` only happens if we've exhausted path resolution, right? Because we get it when calling `OnTerminal()`. Maybe it's worth making that a bit more explicit?
```go
case nil:
    // Parsed a terminal component. This should only be permitted at the very end of the path resolution process.
    if len(rs.stack) > 0 {
        panic("Parser.ParseFirstComponent() did not respect mustBeDirectory")
    }
```

pkg/filesystem/path/scope_walker.go
```diff
@@ -4,22 +4,25 @@ package path
...
-	// most once. Resolve() will always call into OnScope() for
-	// every ScopeWalker presented, though decorators such as
-	// VirtualRootScopeWalkerFactory may only call it when the path
-	// is known to be valid. Absence of calls to OnScope() are used
-	// to indicate that the provided path does not resolve to a
-	// location inside the file system.
-	OnScope(absolute bool) (ComponentWalker, error)
+	// For every instance of ScopeWalker, One of OnAbsolute() or
+	// OnRelative() may be called at most once, and both must never be
+	// called on the same path. Resolve() will always call into one of
```
EdSchouten, line 19: What does "and both must never be called on the same path" mean? Isn't this redundant with what's stated before that?

pkg/filesystem/path/scope_walker.go
```diff
@@ -4,22 +4,25 @@ package path
...
 	//
-	// For every instance of ScopeWalker, OnScope() may be called at
-	// most once. Resolve() will always call into OnScope() for
-	// every ScopeWalker presented, though decorators such as
-	// VirtualRootScopeWalkerFactory may only call it when the path
-	// is known to be valid. Absence of calls to OnScope() are used
-	// to indicate that the provided path does not resolve to a
-	// location inside the file system.
-	OnScope(absolute bool) (ComponentWalker, error)
+	// For every instance of ScopeWalker, One of OnAbsolute() or
```
EdSchouten, line 17: s/One/one

pkg/filesystem/path/unix_parser.go
```diff
@@ -0,0 +1,128 @@
...
+	// Observed a symlink at the end of a
+	// path. We should continue to run.
+	return GotSymlink{
+		Parent: r.Parent,
+		Target: r.Target,
+	}, remainder, nil
+}
+
+// NewUNIXParser creates a Parser for Unix paths that can be used in Resolve.
+func NewUNIXParser(path string) (Parser, error) {
```
EdSchouten, line 109: Could you maybe reorder the contents of this file to be more logical? For example:

- Utility functions at the top
- `unixParser`, `*NewUNIXParser()`, and its methods
- `unixRelativeParser` and its methods

pkg/filesystem/path/parser.go
```diff
@@ -0,0 +1,16 @@
+package path
+
+// Parser is used by Resolve to parse paths in the resolution.
+// The Parser implementations will be used by value in several bookkeeping
+// structures, they must be implemented so they can safely be given as values.
```
EdSchouten, line 5: I think "safely be given as values" is not a very clear way of phrasing it. I would recommend that we state something like this instead:

> Implementations of ParseScope() should return a new copy of Parser and leave the current instance unmodified. It is permitted to call ParseScope() multiple times.

The same for RelativeParser.

pkg/filesystem/path/scope_walker.go
stagnation: Yes it is redundant, I wanted to make it super clear but happy to remove it if it is clear enough :)

pkg/filesystem/path/parser.go
stagnation: Sounds a lot better!

EdSchouten (approved):

EdSchouten:
LGTM! Will merge after you've addressed that gofmt issue.

PR #194, opened 2024-03-10
add blobaccess for Spanner/GCS

srago:
A blobaccess implementation for Spanner/GCS.  Spanner holds metadata and inlined small objects for blobs.  Larger blobs are written to GCS instead of being inlined.


pkg/blobstore/spanner_blob_access.go
```diff
@@ -0,0 +1,975 @@
...
+// Update reference timestamps at insert time (in Put) and in FindMissing (to ensure blobs live
+// long enough for a bazel build or test to complete).  We also prevent rows scheduled for
+// deletion in GCS from being returned by Get.  Thus, if a blob is stale in GCS, it's metadata
+// should also be stale in Spanner, as long as the GCS TTL is at least as long as the Spanner TTL.
+// To make sure action cache entries are retained on an LRU-like basis, we queue up a list of
+// hashes as they are referenced and periodically update ther reference time.  Doing this one at
+// a time incurs too much overhead in spanner, so we perform bulk operations.  The downside of
+// this approach is that we can lose reference updates if the servers reboot while updates are
+// still queued.  Worst case, these objects will be evicted and need to be rebuilt the next time
+// they are needed.
```
EdSchouten, line 16: I don't think this tradeoff is acceptable.

pkg/blobstore/spanner_blob_access.go
```diff
@@ -0,0 +1,975 @@
...
+	if err != nil {
+		log.Printf("Can't create spanner database admin client: %v", err)
+		return err
+	}
+	defer ac.Close()
+
+	// If daysToLive is zero, use the default.
+	if daysToLive == 0 {
+		daysToLive = defaultDaysToLive
+	}
```
EdSchouten, line 225: Please don't do this. Buildbarn doesn't use default values for configuration options. What might be a good default for you, might be a very poor one for me. Sure, we can give some guidance by adding documentation to the .proto file, but in the end the user should give some thought what's the right value for them.

pkg/blobstore/spanner_blob_access.go
```diff
@@ -0,0 +1,975 @@
...
+
+	s := `CREATE TABLE ` + tableName + ` (
+		Key STRING(MAX),
+		InlineData BYTES(MAX),
+		ReferenceTime TIMESTAMP NOT NULL,
+	) PRIMARY KEY(Key), ROW DELETION POLICY (OLDER_THAN(ReferenceTime, INTERVAL ` + strconv.FormatUint(daysToLive, 10) + ` DAY))`
```
EdSchouten, line 231: Aren't there some libraries we can use to construct queries for us (e.g., sqlc) as opposed to smashing strings together?

pkg/blobstore/spanner_blob_access.go
```diff
@@ -0,0 +1,975 @@
...
+
+type spannerRecord struct {
+	Key           string
+	InlineData    []byte
+	ReferenceTime time.Time
+}
+
+// databaseName is of the form "projects/<project ID>/instances/<instance name>/databases/<database name>".
+// tableName should include the name of the table and the version of the table's schema.
+func createSpannerTable(ctx context.Context, databaseName string, daysToLive uint64) error {
```
EdSchouten, line 214: Do we think it's a good idea to have the table creation living here? Doesn't this require that your processes not only have query/insert/delete rights on the table, but also administrative rights to create new tables? I'd rather see that we provide some SQL file for people to run manually when setting up their cluster. Just like how we also don't auto-create S3 buckets.

pkg/blobstore/spanner_blob_access.go
```diff
@@ -0,0 +1,975 @@
...
+		[]string{"backend_type"})
+	backendOperationsDurationSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "buildbarn",
+			Subsystem: "blobstore",
+			Name:      "backend_operations_duration_seconds",
+			Help:      "Amount of time spent per backend operation in seconds.",
+			Buckets:   util.DecimalExponentialBuckets(-3, 6, 2),
+		},
+		[]string{"storage_type", "backend_type", "operation"})
```
EdSchouten, line 184: Isn't this metric superfluous? We already provide such a metric as part of MetricsBlobAccess.

pkg/blobstore/spanner_blob_access.go
```diff
@@ -0,0 +1,975 @@
...
+
+	BEMap = map[pb.StorageType_Value]string{
+		pb.StorageType_CASTORE:      ST_CAS,
+		pb.StorageType_ACTION_CACHE: ST_AC,
+	}
```
EdSchouten, line 189: This is not exhaustive. There are more data stores than just the CAS and AC in the case of Buildbarn. Please just use BlobAccessCreator.GetStorageTypeName().

pkg/blobstore/spanner_blob_access.go
```diff
@@ -0,0 +1,975 @@
...
+}
+
+func (ba *spannerBlobAccess) keyToBlobSize(key string) (int64, error) {
+	s := strings.Split(key, "-")
+	sz, err := strconv.ParseInt(s[1], 10, 64)
+	if err != nil {
+		return -1, err
+	}
+	return sz, nil
+}
```
EdSchouten, line 385: How does this hold for the Action Cache? In the case of the Action Cache the Action Digest pertains to the size of the Action message. Not the ActionResult.

pkg/blobstore/spanner_blob_access.go
```diff
@@ -0,0 +1,975 @@
...
+			} else {
+				log.Printf("Spanner TTL changed from %d to %d days", days, daysToLive)
+			}
+		}
+	}
+
+	storageClient, err := storage.NewClient(ctx)
+	if err != nil {
+		spannerClient.Close()
+		log.Printf("Can't create GCS client: %v", err)
```
EdSchouten, line 464: Either log errors or propagate them, but not both.

pkg/blobstore/spanner_blob_access.go
```diff
@@ -0,0 +1,975 @@
...
+			CacheCapabilities: &remoteexecution.CacheCapabilities{
+				DigestFunctions: digest.SupportedDigestFunctions,
+			},
+		}, nil
+	} else {
+		// Code in pkg/blobaccess/configuration/new_blob_access.go should prevent this, but
+		// since that lives in a separate place, let's be sure.
+		panic("Invalid Spanner storage Type configured")
+	}
+}
```
EdSchouten, line 552: Any reason this can't leverage `capabilities.Provider`?

pkg/blobstore/spanner_blob_access.go
```diff
@@ -0,0 +1,975 @@
...
+			spannerMalformedKeyCount.Inc()
+			log.Printf("Put Blob: invalid digest size, key %s: %v", key, err)
+			b.Discard()
+			return err
+		}
+		if statedSize != size {
+			log.Printf("Put Blob: MISMATCH between buffer size (%d) and digest size (%d)", size, statedSize)
+			b.Discard()
+			return fmt.Errorf("Invalid contents digest %s size %d", digest, size)
+		}
```
EdSchouten, line 709: Do we really need code like this? `pkg/blobstore/buffer` already has logic for validating data integrity.

pkg/blobstore/spanner_blob_access.go
```diff
@@ -0,0 +1,975 @@
...
+		ReferenceTime: now,
+	}
+	insertMut, err := spanner.ReplaceStruct(tableName, rec)
+	if err != nil {
+		log.Printf("Can't create mutation for Blob %s: %v", digest, err)
+		return err
+	}
+
+	start := time.Now()
+	_, err = ba.spannerClient.Apply(context.Background(), []*spanner.Mutation{insertMut})
```
EdSchouten, line 761: Why use `context.Background()` here?

EdSchouten (changes requested):
It looks like you're using Spanner as a flat key-value data store, where the CAS and AC are represented as two completely disjoint tables. As far as I know, Spanner is supposed to be a relational database. Why aren't we using any of its relational capabilities?

pkg/blobstore/spanner_blob_access.go
srago: Performance in unacceptable without this (databases make poor file systems).  Potentially rebuilding stuff that is evicted earlier than it might be is an acceptable tradeoff, because it still results in correct operation.  However, I'm open to ideas on how we can mitigate this.

srago:
> It looks like you're using Spanner as a flat key-value data store, where the CAS and AC are represented as two completely disjoint tables. As far as I know, Spanner is supposed to be a relational database. Why aren't we using any of its relational capabilities?

We have one table that is shared between the AC and CAS.  The driver will behave differently when accessing the table for AC-based operations versus CAS-based operations.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -194,6 +194,9 @@ message BlobAccessConfiguration {
 
     // Refer to a BlobAccess object declared through 'with_labels'.
     string label = 27;
+
+    // Actually a hybrid of spanner and GCS.
```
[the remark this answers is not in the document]
srago: Will do.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -312,6 +315,34 @@ message MirroredBlobAccessConfiguration {
...
+    // Invalid value.
+    UNKNOWN = 0;
+
+    // Action Cache.
+    ACTION_CACHE = 1;
+
+    // Content-Addressable Store.
+    CASTORE = 2;
+  }
+}
```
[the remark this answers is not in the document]
srago: I'll look into that interface.  It didn't exist in the older code base we started our project with.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -312,6 +315,34 @@ message MirroredBlobAccessConfiguration {
...
+
+  // Name of GCS bucket.
+  string gcs_bucket_name = 2;
+
+  // Distinguish between the action cache and the CAS to know how to
+  // format the keys for storage objects.
+  StorageType.Value storage_type = 3;
+
+  // Expiration time in days.  The default is 7 days.
+  uint64 expiration_days = 4;
```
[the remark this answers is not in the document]
srago: It's a little more general than we need, but I'll look into it.

pkg/blobstore/spanner_blob_access.go
srago: I was getting weird results in the capabilities message, so I rolled my own.  I'll try to get you an example.

pkg/blobstore/spanner_blob_access.go
srago: Oh, nice catch.  It's basically dumb luck that this doesn't cause problems, because action cache entries and action messages are small, so we end up using Spanner.  The only place I validate the buffer size is in the CAS.

EdSchouten:
> We have one table that is shared between the AC and CAS. The driver will behave differently when accessing the table for AC-based operations versus CAS-based operations.

Sure, I get that. But what I'm saying is that the references between ActionResult messages and objects stored in the CAS could also be expressed as a relation between two tables in a database. If we did that, then we could most likely eliminate CompletenessCheckingBlobAccess.

pkg/blobstore/spanner_blob_access.go
srago: That only works if the caller receiving the error logs/prints it.  In this case, it's enabling us to debug start-up problems by simply looking at the k8s logs.

pkg/blobstore/spanner_blob_access.go
srago: This one should be ctx.  I use context.Background when I want to perform an action, but have it unaffected if the context gets cancelled.

srago:
> > We have one table that is shared between the AC and CAS. The driver will behave differently when accessing the table for AC-based operations versus CAS-based operations.
> 
> Sure, I get that. But what I'm saying is that the references between ActionResult messages and objects stored in the CAS could also be expressed as a relation (foreign key) between two tables in a database. If we did that, then we could most likely eliminate CompletenessCheckingBlobAccess for these kinds of setups. AC entries could be dropped automatically when associated CAS objects expire.

That isn't immediately obvious to me.  We should discuss...

pkg/blobstore/spanner_blob_access.go
EdSchouten: You can also install things like middleware for gRPC to do the logging. That way we don't need to plaster our codebase with logging statements.

pkg/proto/configuration/blobstore/blobstore.proto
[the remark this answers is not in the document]
EdSchouten: Sure. I can imagine that nobody is interested in setting expirations in nanoseconds. But the advantage of using types like these is that it's well typed. Furthermore, it's becomes self-documenting, meaning we don't need to use suffixes like `_days`.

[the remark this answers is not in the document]
srago: You do realize that the GCS and Spanner expiration times are measured in days, right?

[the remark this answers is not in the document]
EdSchouten: Yes. And the last time I checked, days are a duration.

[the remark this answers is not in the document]
srago: Of course.  But specifying the number of days as a Duration means that users have to convert the days into seconds to configure the expiration time.  Then the driver needs to convert them back into days and have some policy for handling invalid values.  If the user specifies 3 hours, does the driver fail that or does it round up to 1 day?  I've already started working on these changes, but it feels like the wrong approach to me.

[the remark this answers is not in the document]
EdSchouten: The fact that our existing config file parser/protojson only accepts the '1s' syntax for seconds is a limitation of those systems specifically. It should not be a reason against using `google.protobuf.Duration`. For all we know, the user wants to generate a config file using Python, and there they'd  use `datetime.timedelta` in combination with `google.protobuf.duration_pb2.Duration.FromTimedelta()`.

The [Pkl](https://pkl-lang.org/) configuration language actually has a native data type for durations. There you can just write `14.d` to get a literal that corresponds to 14 days. Converting that to `"1209600s"` to appease the protojson parser would happen automatically as part of rendering.

Rounding it up to full days seems like the most sensible choice to me.

pkg/blobstore/spanner_blob_access.go
srago: The anomaly occurs regardless of whether I use capabilities.Provider or roll my own code:

Capabilities for gs.com received:
cache_capabilities:{digest_functions:MD5  digest_functions:SHA1  digest_functions:SHA256  digest_functions:8  digest_functions:SHA384  digest_functions:SHA512  action_cache_update_capabilities:{}  symlink_absolute_path_strategy:ALLOWED}  deprecated_api_version:{major:2}  low_api_version:{major:2}  high_api_version:{major:2  minor:3}

Capabilities for gs.com/unreviewed received:
cache_capabilities:{digest_functions:MD5  digest_functions:SHA1  digest_functions:SHA256  digest_functions:8  digest_functions:SHA384  digest_functions:SHA512  action_cache_update_capabilities:{update_enabled:true}  symlink_absolute_path_strategy:ALLOWED}  execution_capabilities:{digest_function:SHA256  exec_enabled:true  execution_priority_capabilities:{priorities:{min_priority:-2147483648  max_priority:2147483647}}  5:"\x03\x02\x01\x08\x05\x06"}  deprecated_api_version:{major:2}  low_api_version:{major:2}  high_api_version:{major:2  minor:3}

Notice the 5:"\x03\x02\x01\x08\x05\x06" that occurs in the execution_capabilities in the second case.

EdSchouten: Whatever tool you used to dump these message, make sure it uses an up-to-date version of remote_execution.proto. Once you do, you'll notice that field 5 in ExecutionCapabilities is actually `digest_functions`:

https://github.com/bazelbuild/remote-apis/commit/64cc5e9e422c93e1d7f0545a146fd84fcc0e8b47

And what Buildbarn does here is expected. Namely, it reports the list of digest functions that are supported by the scheduler.

pkg/blobstore/spanner_blob_access.go
srago: I rewrote the code to use the knowledge of where the blobs live.  On the write side, we use the size to determine where the blob should be stored.  On the read side, we use the existence of InlineData in the Spanner record to determine where the blob lives.  I have an MR out to (optionally) write small blobs to GCS also so that they can be accessed from GCS-FUSE.  As soon as I get verification that it works, I can send a PR upstream.

pkg/blobstore/spanner_blob_access.go
srago: Yep, rebuilding the tool prints the hash functions.  I hadn't realized the remote APIs has changed that much.

pkg/blobstore/spanner_blob_access.go
srago: Removed.

pkg/blobstore/spanner_blob_access.go
srago: Each of our services run in separate projects with separate permissions.  The table is BuildBarn's, so it naturally creates it if it is missing.  Our Tech Risk team didn't see anything wrong with this approach.  Humans can, of course, create the table manually, but that is error prone.  That's why we automated it.

pkg/blobstore/spanner_blob_access.go
srago: I removed this.

pkg/blobstore/spanner_blob_access.go
srago: Does that give me enough visibility to distinguish between Spanner and GCS accesses?

Issue #185, opened 2024-01-14
Creating buildbarn storage image doesn't work on bazel 7


ericlchen1:
Running the suggested command to create a buildbarn storage image doesn't work `bazel run --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //cmd/bb_storage:bb_storage_container`

Bazelisk automatically downloads Bazel 7 which was recently released and doesn't seem to be compatible with `bb-storage`.
I suggest adding a `.bazelversion` file with `6.4.0` until compatibility with Bazel 7 is done.

stagnation:
Hi, I took a quick look.

The main errors comes form `rules_docker` and platform transitions/providers.
`rules_docker` is deprecated and the status of bazel7 and bzlmodule support is unclear.
But there is a workaround flag: `--@io_bazel_rules_docker//transitions:enable=false` that removes most of the build errors.
https://github.com/bazelbuild/rules_docker/issues/2275
Then it seems that we need to update `protobuf` and `rules_js`.

The updates are simple enough and we can commit that soon:

```
diff --git a/WORKSPACE b/WORKSPACE
index 8e13aa6..9ee9cb0 100644
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -97,9 +97,8 @@ switched_rules_by_language(

 http_archive(
     name = "com_google_protobuf",
-    sha256 = "a700a49470d301f1190a487a923b5095bf60f08f4ae4cac9f5f7c36883d17971",
-    strip_prefix = "protobuf-23.4",
-    urls = ["https://github.com/protocolbuffers/protobuf/releases/download/v23.4/protobuf-23.4.tar.gz"],
+    strip_prefix = "protobuf-25.2",
+    urls = ["https://github.com/protocolbuffers/protobuf/releases/download/v25.2/protobuf-25.2.tar.gz"],
 )

 load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
@@ -163,9 +162,9 @@ http_archive(

 http_archive(
     name = "aspect_rules_js",
-    sha256 = "00e7b97b696af63812df0ca9e9dbd18579f3edd3ab9a56f227238b8405e4051c",
-    strip_prefix = "rules_js-1.23.0",
-    url = "https://github.com/aspect-build/rules_js/releases/download/v1.23.0/rules_js-v1.23.0.tar.gz",
[...]

PR #199, opened 2024-03-15
feat: publish multi-arch images

### LINKED ISSUE #198: Publish images with arm64
Today, the bb-storage container image is published for x86_64 only.

```
% docker image inspect --format "{{.ID}} {{.Architecture}}" ghcr.io/buildbarn/bb-storage:20240311T070255Z-7b232a7
sha256:e1dcf8ca0b7b17a825ae8853aeb9db555b879302b851193592b239381eb32098 amd64
```

However there are some great instance types like https://instances.vantage.sh/aws/ec2/is4gen.medium which provide double the NVME per instance and 2.5X the I/O bandwidth and $0.144/hr vs $0.172 for `i4i.large`

These instances need arm64 architecture.
Note, I'm happy to work on a PR for this.

alexeagle:
Allows bb-storage users to deploy on arm64 architecture, e.g. AWS Graviton. Fixes #198

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -12,9 +12,21 @@ http_archive(
 )
 
 http_archive(
-    name = "io_bazel_rules_docker",
-    sha256 = "b1e80761a8a8243d03ebca8845e9cc1ba6c82ce7c5179ce2b295cd36f7e394bf",
-    urls = ["https://github.com/bazelbuild/rules_docker/releases/download/v0.25.0/rules_docker-v0.25.0.tar.gz"],
+    name = "aspect_bazel_lib",
+    sha256 = "6c25c59581041ede31e117693047f972cc4700c89acf913658dc89d04c338f8d",
+    strip_prefix = "bazel-lib-2.5.3",
+    url = "https://github.com/aspect-build/bazel-lib/releases/download/v2.5.3/bazel-lib-v2.5.3.tar.gz",
+)
+
+load("@aspect_bazel_lib//lib:repositories.bzl", "register_expand_template_toolchains")
+
+register_expand_template_toolchains()
+
+http_archive(
+    name = "rules_oci",
+    sha256 = "4a276e9566c03491649eef63f27c2816cc222f41ccdebd97d2c5159e84917c3b",
+    strip_prefix = "rules_oci-1.7.4",
+    url = "https://github.com/bazel-contrib/rules_oci/releases/download/v1.7.4/rules_oci-v1.7.4.tar.gz",
 )
 
 http_archive(
@@ -36,7 +48,7 @@ http_archive(
     ],
 )
 
-load("@bazel_gazelle//:deps.bzl", "go_repository")
+load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
 
 # Override the version of gomock to one that includes support for
 # generating mocks for function types. We can't do this through go.mod,
@@ -61,21 +73,31 @@ go_rules_dependencies()
 
 go_register_toolchains(version = "1.21.5")
 
-load("@io_bazel_rules_docker//repositories:repositories.bzl", container_repositories = "repositories")
+load("@rules_oci//oci:pull.bzl", "oci_pull")
+load("@rules_oci//oci:repositories.bzl", "LATEST_CRANE_VERSION", "oci_register_toolchains")
 
-container_repositories()
+oci_register_toolchains(
+    name = "oci",
+    crane_version = LATEST_CRANE_VERSION,
+)
 
-load("@io_bazel_rules_docker//repositories:deps.bzl", container_deps = "deps")
-
-container_deps()
-
-load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
+# NB: this base image is chosen to exactly match what we had when using rules_docker
+# prior to March 2024:
+# $ bazel query --output=build @go_image_static//image:image
+# container_import(
+#   name = "image",
+#   base_image_digest = "sha256:fac888659ca3eb59f7d5dcb0d62540cc5c53615e2671062b36c815d000da8ef4",
+#   base_image_registry = "gcr.io",
+#   base_image_repository = "distroless/static",
+# )
+oci_pull(
+    name = "distroless_static",
+    digest = "sha256:fac888659ca3eb59f7d5dcb0d62540cc5c53615e2671062b36c815d000da8ef4",
+    image = "gcr.io/distroless/static",
+    platforms = ["linux/amd64","linux/arm64"],
+)
 
 gazelle_dependencies()
-
-load("@io_bazel_rules_docker//go:image.bzl", _go_image_repos = "repositories")
-
-_go_image_repos()
 
 http_archive(
     name = "com_github_bazelbuild_buildtools",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -73,6 +73,8 @@ go_rules_dependencies()
 
 go_register_toolchains(version = "1.21.5")
 
+gazelle_dependencies()
+
 load("@rules_oci//oci:pull.bzl", "oci_pull")
 load("@rules_oci//oci:repositories.bzl", "LATEST_CRANE_VERSION", "oci_register_toolchains")
 
@@ -100,8 +102,6 @@ oci_pull(
         "linux/arm64",
     ],
 )
-
-gazelle_dependencies()
 
 http_archive(
     name = "com_github_bazelbuild_buildtools",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -81,7 +81,7 @@ oci_register_toolchains(
     crane_version = LATEST_CRANE_VERSION,
 )
 
-# NB: this base image is chosen to exactly match what we had when using rules_docker
+# NB: this base image is chosen to match what we had when using rules_docker
 # prior to March 2024:
 # $ bazel query --output=build @go_image_static//image:image
 # container_import(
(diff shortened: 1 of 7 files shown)


cmd/bb_copy/BUILD.bazel
```diff
@@ -20,18 +21,43 @@ go_library(
...
+        pure = "on",
+        goarch = goarch,
+        goos = "linux",
+        visibility = ["//visibility:public"],
+    )
+    for goarch in ["amd64", "arm64"]
+]
 
-go_image(
+[
```
alexeagle, line 36: note, there's some code duplication between the three cmd packages. If you'd prefer I can add a macro in container.bzl like `multiarch_go_image`

alexeagle:
I just learned about #167 , sorry this is replacing that I think

cmd/bb_copy/BUILD.bazel
EdSchouten: I'd appreciate that. Otherwise we need to patch up this code every time we want to add support for another architecture.

cmd/bb_copy/BUILD.bazel
```diff
@@ -20,18 +21,43 @@ go_library(
...
+    go_binary(
+        name = "bb_copy-{}".format(goarch),
+        embed = [":bb_copy_lib"],
+        pure = "on",
+        goarch = goarch,
+        goos = "linux",
+        visibility = ["//visibility:public"],
+    )
+    for goarch in ["amd64", "arm64"]
+]
```
EdSchouten, line 34: I don't think this change is acceptable for two reasons:

1. We no longer build Go binaries for the host architecture. We still want to have binaries for macOS, Windows, FreeBSD, etc..
2. `oci_image_index()` should just use transitions. That way there would be no need for us to put loops around `go_binary()`, `pkg_tar()`, and `oci_image()`.

tools/container.bzl
```diff
@@ -1,11 +1,9 @@
...
-    container_push(
+    oci_push(
         name = name,
-        format = "Docker",
         image = image,
-        registry = "ghcr.io",
-        repository = "buildbarn/" + component,
-        tag = "{BUILD_SCM_TIMESTAMP}-{BUILD_SCM_REVISION}",
+        repository = "ghcr.io/buildbarn/" + component,
+        remote_tags = "//tools:stamped_tags",
```
EdSchouten, line 8: i suspect that you want to use `@com_github_buildbarn_bb_storage//tools:stamped_tags` here. Otherwise this won't be usable in bb-remote-execution and other repositories.

EdSchouten (changes requested):

cmd/bb_replicator/BUILD.bazel
```diff
@@ -22,18 +23,43 @@ go_library(
...
+    )
+    for goarch in ["amd64", "arm64"]
+]
 
-go_image(
+[
+    pkg_tar(
+        name = "tar-{}".format(goarch),
+        srcs = ["bb_replicator-{}".format(goarch)],
+        package_dir = "/app/cmd/bb_storage",
```
EdSchouten, line 42: Out of curiosity, why do we want to retain this path prefix?

alexeagle: I only did this to minimize the delta with the images you publish today. That's the path I see the binaries currently reside within the image.

EdSchouten: Feel free to remove it. As long as the resulting image is functionally equivalent, it's fine.

cmd/bb_copy/BUILD.bazel
alexeagle: done, PTAL

tools/container.bzl
```diff
@@ -1,11 +1,48 @@
...
+        name: resulting oci_image_index target
+        binary: label of a go_binary target; it may be transitioned to another architecture
+    """
+    images = []
+    tar_target = "_{}.tar".format(name)
+    image_target = "_{}.image".format(name)
+    pkg_tar(
+        name = tar_target,
+        srcs = [binary],
+        package_dir = "/app/cmd/bb_storage",
```
EdSchouten, line 18: Considering that not all our executables are named "bb_storage", I think we should remove this.

Can't we simply use /app as the package dir?

alexeagle: Sure, as long as the only reference to the path within the container is the entry_point. If users might have some config that has this path, it's obviously a breaking change to put the binary somewhere different.

alexeagle:
Do you think my change is responsible for the gazelle failure on CI?

EdSchouten:
> Do you think my change is responsible for the gazelle failure on CI?

Yes. I can reproduce it by running the commands run by the CI pipeline on my local system. It looks like I can address it by calling `gazell_dependencies()` earlier on:

```diff
diff --git a/WORKSPACE b/WORKSPACE
index 8806a54..9747956 100644
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -73,6 +73,8 @@ go_rules_dependencies()
 
 go_register_toolchains(version = "1.21.5")
 
+gazelle_dependencies()
+
 load("@rules_oci//oci:pull.bzl", "oci_pull")
 load("@rules_oci//oci:repositories.bzl", "LATEST_CRANE_VERSION", "oci_register_toolchains")
 
@@ -101,8 +103,6 @@ oci_pull(
     ],
 )
 
-gazelle_dependencies()
-
 http_archive(
     name = "com_github_bazelbuild_buildtools",
     sha256 = "09a94213ea0d4a844e991374511fb0d44650e9c321799ec5d5dd28b250d82ca3",
```

tools/container.bzl
EdSchouten: It should be. I don't think it's typical for people to override the entrypoint for the Buildbarn containers.

tools/container.bzl
```diff
@@ -1,11 +1,50 @@
...
+        binary: label of a go_binary target; it may be transitioned to another architecture
+    """
+    images = []
+    tar_target = "_{}.tar".format(name)
+    image_target = "_{}.image".format(name)
+
+    pkg_tar(
+        name = tar_target,
+        srcs = [binary],
+        include_runfiles = True,
```
EdSchouten, line 19: Shouldn't this specify /app as a package dir?

tools/container.bzl
alexeagle: okay, changed to just `app`

tools/container.bzl
```diff
@@ -1,11 +1,51 @@
...
+        package_dir = "app",
+    )
+
+    oci_image(
+        name = image_target,
+        base = "@distroless_static",
```
EdSchouten, line 25: Wouldn't this expand to /app/:bb_storage? Notice the colon.

alexeagle:
PTAL - I have not integration-tested this by pushing a container and running a service from it.

tools/container.bzl
alexeagle: yes, just added a removeprefix on that, I noticed as well

EdSchouten: Can't we use Label(binary).name here instead?

tools/container.bzl
alexeagle: yes, I pushed another commit right after this one, sorry you reviewed the intermediate state

tools/container.bzl
alexeagle: not quite - in the macro context the string isn't fully qualified:

`Error in Label: Illegal absolute label syntax: invalid label ':bb_copy': absolute label must begin with '@' or '//'`

EdSchouten: What about https://bazel.build/rules/lib/toplevel/native.html#package_relative_label ?

alexeagle: Yeah I think that's what we want, but it was introduced in Bazel 6.1.0 https://github.com/bazelbuild/bazel/commit/b859571804f2095caaf018b172b59c90f185fd51 and this repo currently uses Bazel 6.0.0

EdSchouten: Then let's try upgrading to 6.1.0 or later.

alexeagle: pushed a commit to upgrade to Bazel 6.1.0, guessing that's what you'd prefer

WORKSPACE
```diff
@@ -12,9 +12,21 @@ http_archive(
 )
 
 http_archive(
-    name = "io_bazel_rules_docker",
```
stagnation, line 15: Speaking of bazel versions. I think you fix the blocker for bazel 7 here :)
https://github.com/buildbarn/bb-storage/issues/185

I never got around to verifying `rules_docker` builds on bazel 7 so thankful for this uplift.
We can look at bazel 7 after this PR :+1:

alexeagle: Yes that sounds correct.

moroten:
I created https://github.com/bazel-contrib/rules_oci/pull/531 to make `oci_image_index` perform the target platform transition instead.

EdSchouten:
I saw that that PR build of yours failed with some protoc code generation issues. I thought that would be caused by the Bazel 6.1.0 upgrade, so that's why I just did an upgrade to Bazel 6.4.0. Be sure to rebase your PR on top of `master` to see if it goes through now.

WORKSPACE
```diff
@@ -61,21 +73,35 @@ go_rules_dependencies()
...
+
+# NB: this base image is chosen to match what we had when using rules_docker
+# prior to March 2024:
+# $ bazel query --output=build @go_image_static//image:image
+# container_import(
+#   name = "image",
+#   base_image_digest = "sha256:fac888659ca3eb59f7d5dcb0d62540cc5c53615e2671062b36c815d000da8ef4",
+#   base_image_registry = "gcr.io",
+#   base_image_repository = "distroless/static",
+# )
```
EdSchouten, line 94: Could you remove this comment and the NB below, please? We don't care strongly about using some exact version of the distroless image. I don't see the point in having some arbitrary version of an image documented.

alexeagle:
@moroten FYI that I tried patching your rules_oci PR but it still fails on `linux/i386` in the same way as this PR does.

 https://github.com/buildbarn/bb-storage/actions/runs/8397776269/job/23006173303?pr=201

alexeagle:
@EdSchouten I'm hoping my last commit made it green - do you have something like a label you could apply so the workflow doesn't require a manual approval when I push commits? 🙏🏻

EdSchouten:
Github auto approves workflows if you have made at least 1 commit to this repo. So after I merge this change, any followup PR made by you should automatically run the build.

alexeagle:
Hooray it's green, anything else you'd like me to change?

moroten:
Thank you for merging this. The ideas with https://github.com/bazel-contrib/rules_oci/pull/531 for multi arch images will take some time to get ready.

PR #206, opened 2024-06-04
Feature/implement windows parser and on drive letter

stagnation:


pkg/filesystem/path/scope_walker.go
```diff
@@ -25,4 +27,5 @@ type ScopeWalker interface {
 	// system.
 	OnAbsolute() (ComponentWalker, error)
 	OnRelative() (ComponentWalker, error)
+	OnDriveLetter() (ComponentWalker, error)
```
EdSchouten, line 30: Shouldn't this be:
```go
OnDriveLetter(drive rune) (ComponentWalker, error)
```
?

pkg/filesystem/local_directory_windows.go
```diff
@@ -108,11 +109,7 @@ func newLocalDirectory(absPath string, openReparsePoint bool) (DirectoryCloser,
 }
 
 func NewLocalDirectory(path string) (DirectoryCloser, error) {
```
EdSchouten, line 111: Could we please change `NewLocalDirectory()` now to take a `path.Parser` instead of a string now?

pkg/filesystem/path/builder.go
```diff
@@ -20,6 +20,7 @@ import (
 // system.
 type Builder struct {
 	absolute             bool
+	driveletter          rune
```
EdSchouten, line 23: `driveLetter`.

pkg/filesystem/path/builder.go
```diff
@@ -53,6 +64,12 @@ func (b *Builder) GetUNIXString() string {
 	if b.absolute {
 		prefix = "/"
 	}
+	if b.driveletter != rune(0) {
+		// Drive letter is not meaningful in the REAPI. We just drop it.
+		// Symlinks may point to the same drive but not other drives,
+		// so it can be implicit.
+		prefix = "/"
+	}
```
EdSchouten, line 72: Instead of doing this, can't we just let `OnDriveLetter()` set `absolute = true`?

pkg/filesystem/path/builder.go
```diff
@@ -62,15 +79,57 @@ func (b *Builder) GetUNIXString() string {
...
 	// or a dot or slash if the path is empty.
-	out.WriteString(b.suffix)
+	suffix := b.suffix
+	// The suffix may have been constructed by platform-specific code that
+	// uses backslashes. To construct a UNIX path we must use a forward slash.
+	// We can construct UNIX paths from Windows-native paths, where the On*
+	// scope functions set the suffix to backslash.
+	if suffix == "\\" {
+		suffix = "/"
+	}
```
EdSchouten, line 89: Can't we just let `suffix` always store the UNIX format?

pkg/filesystem/path/relative_scope_walker.go
```diff
@@ -21,6 +21,10 @@ func (pw *relativeScopeWalker) OnAbsolute() (ComponentWalker, error) {
 	return nil, status.Error(codes.InvalidArgument, "Path is absolute, while a relative path was expected")
 }
 
+func (pw *relativeScopeWalker) OnDriveLetter(drive rune) (ComponentWalker, error) {
+	return nil, status.Error(codes.InvalidArgument, "Path is absolute with drive letter, while a relative path was expected")
```
EdSchouten, line 25: s/is absolute with/has a/

pkg/filesystem/path/stringer.go
```diff
@@ -4,4 +4,5 @@ package path
 // converted to string representations.
 type Stringer interface {
 	GetUNIXString() string
+	GetWindowsString() string
```
EdSchouten, line 7: I have a feeling this should return `(string, error)`. Windows is a lot more restrictive than UNIX when it comes to representing paths. There are a whole bunch of reserved characters in filenames. There are also certain names that are forbidden (e.g., `LPT0`).

There may be cases in Buildbarn where we construct paths only using UNIX-style paths as inputs. But at the end we call `GetWindowsString()` to obtain a Windows-style path. That call should fail if the path is not representable.

Though I'm not requiring that you get this 100% watertight (that it rejects all paths that are not valid on Windows), the API should be prepared for it.

pkg/filesystem/path/builder_test.go
```diff
@@ -31,13 +32,77 @@ func TestBuilder(t *testing.T) {
...
+			"/C/hello/..",
+			"/C/hello/../world",
+			"/C/hello/../world/",
+			"/C/hello/../world/foo",
+			"\\C\\",
+			"\\C\\hello\\",
+			"\\C\\hello\\..",
+			"\\C\\hello\\..\\world",
+			"\\C\\hello\\..\\world\\",
+			"\\C\\hello\\..\\world\\foo",
```
EdSchouten, line 70: I don't think we should support these constructs. On Windows, `\C\Hello` would refer to `X:\C\Hello`, assuming `X:` is the current drive.

pkg/filesystem/path/component_walker.go
```diff
@@ -56,9 +56,10 @@ type ComponentWalker interface {
 	// If the pathname component refers to a symbolic link, this
 	// function will return a GotSymlink containing a ScopeWalker, which
 	// can be used to perform expansion of the symbolic link. The
-	// Resolve() function will call into OnAbsolute() or OnRelative() to
-	// signal whether resolution should continue at the root directory
-	// or at the directory that contained the symbolic link.
+	// Resolve() function will call into OnAbsolute(), OnRealtive() or
```
EdSchouten, line 59: s/OnRealtive/OnRelative/

pkg/filesystem/local_directory_windows.go
stagnation: Yes! I have the corresponding changes to bb-remote-execution on the way, and it improved the build working directory code I had outstanding before, which is nice.

stagnation: https://github.com/buildbarn/bb-remote-execution/pull/138

pkg/filesystem/local_directory_unix.go
```diff
@@ -33,8 +33,17 @@ func newLocalDirectoryFromFileDescriptor(fd int) (*localDirectory, error) {
 
 // NewLocalDirectory creates a directory handle that corresponds to a
 // local path on the system.
-func NewLocalDirectory(path string) (DirectoryCloser, error) {
-	fd, err := unix.Openat(unix.AT_FDCWD, path, unix.O_DIRECTORY|unix.O_NOFOLLOW|unix.O_RDONLY, 0)
+func NewLocalDirectory(directoryParser path.Parser) (DirectoryCloser, error) {
+	directoryPath, scopeWalker := path.EmptyBuilder.Join(path.NewAbsoluteScopeWalker(path.VoidComponentWalker))
```
EdSchouten, line 37: Maybe better to use VoidScopeWalker here. Right now we don't require that this function takes an absolute path.

stagnation: :+1:

pkg/filesystem/path/builder_test.go
```diff
@@ -31,13 +37,65 @@ func TestBuilder(t *testing.T) {
...
+			"C:\\hello\\..\\world\\",
+			"C:\\hello\\..\\world\\foo",
+		} {
+			t.Run(p, func(t *testing.T) {
+				// Drive letter paths. These are printed as qualified absolute
+				// paths by the Windows string formatter, and as bare absolute
+				// UNIX paths without a drive letter.
+				noDriveletter := p[2:]
+				unixStyle := strings.Replace(noDriveletter, "\\", "/", -1)
+				driveletter := "C:" + noDriveletter
```
EdSchouten, line 77: https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

I would prefer it if we could refrain from string concatenation and `strings.Replace()` as part of these tests. Maybe it's best if this test purely focused on the interaction between `NewWindowsParser()` and `GetWindowsString()`, because as the name implies this is about the identity relationship between them.

If we want to test the interaction between `NewWindowsString()` and `GetUNIXString()`, that should likely be provided as part of a separate test? (Maybe also rename "Identity" to "UNIXIdentity", so that this can become a separate "WindowsIdentity"?)

pkg/filesystem/path/builder.go
```diff
@@ -66,11 +83,44 @@ func (b *Builder) GetUNIXString() string {
...
+
+	var out strings.Builder
+	if b.driveLetter != rune(0) {
+		out.WriteString(string(b.driveLetter) + ":")
+	}
+
+	for _, component := range b.components {
+		out.WriteString(prefix)
+		out.WriteString(component)
```
EdSchouten, line 102: This does not take into account whether the component is valid on Windows. Could we please extend it to check for control characters, backslashes, colons, etc. etc. etc.?

pkg/filesystem/path/windows_parser.go
```diff
@@ -0,0 +1,78 @@
...
+	path = strings.Replace(path, "/", "\\", -1)
+	// C:\
+	if len(path) >= 3 && unicode.IsLetter(rune(path[0])) && path[1] == ':' && path[2] == '\\' {
+		driveLetter = rune(path[0])
+		path = path[3:]
+	}
+
+	if len(path) >= 1 && path[0] == '\\' {
+		absolute = true
+	}
```
EdSchouten, line 38: My recommendation is to do this as part of ParseScope(). That prevents the need for storing all of these things separately in `windowsParser`.

pkg/filesystem/path/unix_parser.go
```diff
@@ -49,33 +49,34 @@ func (p unixParser) ParseScope(scopeWalker ScopeWalker) (next ComponentWalker, r
...
 
-	return next, unixRelativeParser{p.path}, nil
+	return next, relativeParser{p.path, '/'}, nil
 }
 
-type unixRelativeParser struct {
-	path string
+type relativeParser struct {
+	path          string
+	pathSeparator byte
```
EdSchouten, line 65: I thought that Windows always accepted paths that either contain backslashes or forward slashes, or a mixture thereof. My recommendation would be to give the Windows parser its own `relativeParser` that checks both.

pkg/filesystem/path/windows_parser.go
```diff
@@ -0,0 +1,78 @@
...
+	driveLetter := rune(0)
+	absolute := false
+
+	if strings.ContainsRune(path, '\x00') {
+		return nil, status.Error(codes.InvalidArgument, "Path contains a null byte")
+	}
+
+	path = strings.Replace(path, "/", "\\", -1)
+	// C:\
+	if len(path) >= 3 && unicode.IsLetter(rune(path[0])) && path[1] == ':' && path[2] == '\\' {
```
EdSchouten, line 31: Instead of using `unicode.IsLetter()`, I'd recommend using `path[0] >= 'A' && path[0] <= 'Z'`. The reason being that Windows does not permit any non-ASCII characters. The conversion you're doing there (`rune(path[0])`) also means you're interpreting the first byte as if it's ISO-8859-1 instead of UTF-8.

EdSchouten: Furthermore, you likely also want to put some code here that converts lowercase letters to uppercase.

pkg/filesystem/path/unix_parser.go
stagnation: Aha, I did not think of a mixture of the two. Do we want to fully support that now? With a new relative parser we improve the state of affairs, but I think more changes are required for it to work entirely.

pkg/filesystem/path/windows_parser.go
stagnation: Good idea, that makes it much neater! Less state to hang on to and the ParseScope() function is improved too.

pkg/filesystem/path/builder.go
stagnation: :+1: Do we want them in-line here? And should I create test-cases for them through the Parse code path, or should I create a component-verification function that can be tested directly?

pkg/filesystem/path/builder_test.go
stagnation: Sure thing!

pkg/filesystem/path/builder.go
```diff
@@ -66,11 +83,44 @@ func (b *Builder) GetUNIXString() string {
...
+	// Emit pathname components.
+	prefix := ""
+	if b.absolute {
+		prefix = "\\"
+	}
+
+	var out strings.Builder
+	if b.driveLetter != rune(0) {
+		out.WriteString(string(b.driveLetter) + ":")
+	}
```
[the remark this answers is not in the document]
stagnation: Yes, that looks neater. Though we must write out the colon for an empty component list.
So I guess it is better to write the colon directly to `out`.

pkg/filesystem/path/builder.go
```diff
@@ -66,11 +83,44 @@ func (b *Builder) GetUNIXString() string {
 	return out.String()
 }
 
+// GetWindowsString returns a string representation of the path for use on
+// UNIX-like operating systems. We do not create a windows-style path.
```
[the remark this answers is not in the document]
stagnation: :eyes: This must have been a remnant from when I misunderstood the task and operated on the forward-slash unixy paths. And wanted to indicate that this had confusing behavior. But it doesn't anymore.

pkg/filesystem/local_directory_test.go
```diff
@@ -13,13 +13,17 @@ import (
...
+	parser, err := path.NewLocalParser(t.TempDir())
+	require.NoError(t, err)
+	d, err := filesystem.NewLocalDirectory(parser)
 	require.NoError(t, err)
 	return d
 }
 
 func TestLocalDirectoryCreationFailure(t *testing.T) {
-	_, err := filesystem.NewLocalDirectory("/nonexistent")
+	parser, err := path.NewLocalParser("/nonexistent")
```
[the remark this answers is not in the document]
stagnation: I can change that if you want, but the name indicates it should be "local". Just making sure. 
I should also note that this PR does not successfully run the tests on Windows for a host of Bazel issues. I would prefer to handle that separately.

pkg/filesystem/path/windows_parser.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		panic(err)
+	}
+	return parser
+}
+
+func (p windowsParser) ParseScope(scopeWalker ScopeWalker) (next ComponentWalker, remainder RelativeParser, err error) {
+	path := strings.Replace(p.path, "/", "\\", -1)
+
+	if len(path) >= 3 {
```
stagnation, line 42: Not too happy with this nested if. 
Maybe we should check for the empty string and let that exit early instead.
Then the is-absolute-branch can access the first character as well.

pkg/filesystem/path/trace.go
```diff
@@ -44,3 +44,13 @@ func (t *Trace) GetUNIXString() string {
...
+
+// GetWindowsString returns a string representation of the path for use on Windows.
+func (t *Trace) GetWindowsString() (string, error) {
+	if t == nil {
+		return ".", nil
+	}
+	var sb strings.Builder
+	t.writeToStringBuilder(&sb)
+	return sb.String(), nil
+}
```
EdSchouten, line 56: This still emits a UNIX style path containing potentially invalid filenames, etc. Probably best to rename `writeToStringBuilder` to `writeUNIXToStringBuilder` and add a `writeWindowsToStringBuilder`?

pkg/filesystem/path/windows_parser.go
EdSchouten, line 42: This ends up interpreting paths like `C:` and `C:Hello.txt` as relative paths.

Even though `C:` and `C:Hello.txt` technically speaking have a different meaning than `C:\` and `C:\Hello.txt`, I think we should simply interpret them as such.

I suspect the right thing to do is to only consume the first two characters, and then call something similar to `stripOneOrMoreSlashes()`.

pkg/filesystem/path/windows_parser.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// NewWindowsParser creates a Parser for Windows paths that can be used in Resolve.
+func NewWindowsParser(path string) (Parser, error) {
+	// Windows paths are generally passed to system calls that
+	// accept C strings. There is no way these can accept null
+	// bytes.
+	if strings.ContainsRune(path, '\x00') {
+		return nil, status.Error(codes.InvalidArgument, "Path contains a null byte")
+	}
```
EdSchouten, line 25: What are your thoughts on using
```
if strings.ContainsFunc(component, unicode.IsControl) || strings.ContainsAny(component, "<>:\"/\\|?*") {
```
here as well? Probably best if we put that in a helper function. Though for paths `:` *are* allowed. It's just that they are not allowed for filenames. Maybe it's best to remove this check entirely from `NewWindowsParser()` entirely, and only perform it as part of `Parse*()`?

pkg/filesystem/path/windows_parser.go
```diff
@@ -0,0 +1,152 @@
...
+func MustNewWindowsParser(path string) Parser {
+	parser, err := NewWindowsParser(path)
+	if err != nil {
+		panic(err)
+	}
+	return parser
+}
+
+func (p windowsParser) ParseScope(scopeWalker ScopeWalker) (next ComponentWalker, remainder RelativeParser, err error) {
+	path := strings.Replace(p.path, "/", "\\", -1)
```
EdSchouten, line 40: Any way we can implement this without copying the incoming string?

pkg/filesystem/path/windows_parser.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func (rp windowsRelativeParser) ParseFirstComponent(componentWalker ComponentWalker, mustBeDirectory bool) (next GotDirectoryOrSymlink, remainder RelativeParser, err error) {
+	var name string
+	terminal := false
+
+	separator := -1
+	{
+		slash := strings.IndexByte(rp.path, '/')
+		backslash := strings.IndexByte(rp.path, '\\')
```
EdSchouten, line 82: `strings.IndexAny()`.

pkg/filesystem/path/windows_parser.go
EdSchouten: What about this?

```go
if len(path) >= 2 && (
    (path[0] >= 'A' && path[0] <= 'Z') ||
    (path[0] >= 'a' && path[0] <= 'z')
) && path[1] == ':' {
    next, err = scopeWalker.OnDriveLetter(rune(path[0]&0x1f + 'A' - 1))
    if err != nil {
        return nil, nil, err
    }
    return next, windowsRelativeParser{stripOneOrMoreSlashes(path[2:])}, nil
}
```

pkg/filesystem/path/builder_test.go
```diff
@@ -42,6 +48,104 @@ func TestBuilder(t *testing.T) {
...
+				require.NoError(t, path.Resolve(path.MustNewWindowsParser(p), scopewalker1))
+				require.Equal(t, expected, getWindowsString(t, builder1))
+
+				builder2, scopewalker2 := path.EmptyBuilder.Join(path.VoidScopeWalker)
+				require.NoError(t, path.Resolve(builder1, scopewalker2))
+				require.Equal(t, expected, getWindowsString(t, builder2))
+			})
+		}
+	})
+
```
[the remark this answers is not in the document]
stagnation: Yes. That would be good. 
Though with the current setup some of the tests need to be changed,
and maybe we should have a windows-normalization test for those in particular.

```
t.Run("WindowsIdentity", func(t *testing.T) {
...
				require.Equal(t, p, getWindowsString(t, builder1))
```

The `p` here is not conducive for normalization tests.

```
t.Run("WindowsParseUNIXPaths", func(t *testing.T) {
...
				require.Equal(t, p, builder1.GetUNIXString())
```

Likewise here. But I could introduce a third column "input, expectedLinux, expectedWindows".

pkg/filesystem/path/trace.go
stagnation: Good. I am less familiar with the code paths for traces, I guess this should have an identity parsing test. Notably I do not think this tracks the drive letter as the code stands.

pkg/filesystem/path/windows_parser.go
stagnation: Looks good. The pathological cases do contain invalid characters, so we will error out if they are used rather than attempting to use files that it should not use. So I think this is fine.

pkg/filesystem/path/windows_parser.go
stagnation: Sure. I can try that, more and more code need to have work on "separator" rather than "slash" so I can give that a stab here too.

pkg/filesystem/path/windows_parser.go
stagnation: Resolved in another thread.

pkg/filesystem/local_directory_test.go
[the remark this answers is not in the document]
stagnation: Thanks, that makes sense.

pkg/filesystem/path/unix_parser.go
```diff
@@ -49,22 +61,22 @@ func (p unixParser) ParseScope(scopeWalker ScopeWalker) (next ComponentWalker, r
 			return nil, nil, err
 		}
 
-		return next, unixRelativeParser{stripOneOrMoreSlashes(p.path)}, nil
+		return next, relativeParser{stripOneOrMoreSlashes(p.path)}, nil
```
EdSchouten, line 64: Please rename this back to unixRelativeParser.

pkg/filesystem/path/unix_parser.go
```diff
@@ -16,6 +16,18 @@ func stripOneOrMoreSlashes(p string) string {
...
 
+func stripZeroOrMoreSeparators(p string) string {
+	if p[0] == '/' || p[0] == '\\' {
+		for {
+			p = p[1:]
+			if p == "" || (p[0] != '/' && p[0] != '\\') {
+				return p
+			}
+		}
+	}
```
EdSchouten, line 27:
```go
for p != "" && (p[0] == '/' || p[0] == '\\') {
    p = p[1:]
}
```

This function should live in windows_parser.go.

pkg/filesystem/path/windows_parser.go
stagnation: I am fine with the parse then validate and use split. And then only basic verification is needed here, just like in the UNIX case. That strategy is clear for me now that I have worked with it. But it may need to be articulated further in the parser documentation.

pkg/filesystem/path/windows_parser.go
stagnation: We practically had everything in place already, this was a small change following the other refactorings for path separators :+1:.

pkg/filesystem/path/trace.go
stagnation: Offline discussion: the Trace is only used for relative paths. So we do not need to worry about drive letters.


I have now moved the validation into the meat of the implementation. Will add a test.

pkg/filesystem/path/trace.go
```diff
@@ -26,21 +26,50 @@ func (t *Trace) Append(component Component) *Trace {
...
+func (t *Trace) writeToUNIXStringBuilder(sb *strings.Builder) {
 	if t.parent != nil {
-		t.parent.writeToStringBuilder(sb)
+		t.parent.writeToUNIXStringBuilder(sb)
 		sb.WriteByte('/')
 	}
 	sb.WriteString(t.component.String())
 }
 
+func (t *Trace) writeToWindowsStringBuilder(sb *strings.Builder) error {
```
EdSchouten, line 37: Sorry for getting back to this, but giving it some more thought maybe the cleanest thing is to stick to a single `writeToStringBuilder()`, but extend it to take a `separator rune` argument.

With regards to validating the path for Windows, we can add an extra loop to `GetWindowsString()`:
```go
for tValidate := t; tValidate != nil; tValidate = tValidate.parent {
    if err := validateWindowsPathComponent(tValidate.component.String()); err != nil {
        return err
    }
}
```

pkg/filesystem/path/trace_test.go
```diff
@@ -4,19 +4,58 @@ import (
 	"testing"
 
 	"github.com/buildbarn/bb-storage/pkg/filesystem/path"
+	"github.com/buildbarn/bb-storage/pkg/testutil"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/codes"
```
EdSchouten, line 9: Nit: Add empty line between github.com and golang.org imports, for consistency with the rest of the code base.

pkg/filesystem/path/builder.go
```diff
@@ -268,3 +336,10 @@ func (rp builderRelativeParser) ParseFirstComponent(componentWalker ComponentWal
 	}
 	return r, nil, nil
 }
+
+func validateWindowsPathComponent(component string) error {
+	if strings.ContainsFunc(component, unicode.IsControl) || strings.ContainsAny(component, "<>:\"/\\|?*") {
+		return status.Errorf(codes.InvalidArgument, "Path component contains a control character or invalid character %#v.", component)
```
EdSchouten, line 342: Please remove the trailing dot. Also move the `%#v` to the right place in the sentence.
```go
return status.Errorf(codes.InvalidArgument, "Path component %#v contains a control character or invalid character %#v", component)
```
That said, the general pattern in the Buildbarn codebase is to never format errors with values that also happen to be arguments of the function. Probably better to do this?
```go
if strings.ContainsFunc(component, unicode.IsControl) {
    return status.Error(codes.InvalidArgument, "Path component contains control characters")
}
if strings.ContainsAny(component, "<>:\"/\\|?*") {
   return status.Error(codes.InvalidArgument, "Path component contains reserved characters")
}
```
And then let the caller do a `util.StatusWrapf(err, "Invalid path component %#v", component.String())`.

pkg/filesystem/path/windows_parser.go
```diff
@@ -0,0 +1,138 @@
...
+		return GotDirectory{Child: componentWalker}, remainder, nil
+	case "..":
+		// Traverse to the parent directory.
+		parent, err := componentWalker.OnUp()
+		if err != nil {
+			return nil, nil, err
+		}
+		return GotDirectory{Child: parent}, remainder, nil
+	}
+
```
EdSchouten, line 106: Could we add a call to `validateWindowsPathComponent()` here? This ensures that we also don't parse malformed paths.

EdSchouten (commented):
This is looking pretty good. Thanks for working on this!

pkg/filesystem/path/builder.go
EdSchouten: Note that there are also some additional restrictions on Windows, such as filenames not being allowed to end with dots or whitespace. And there are obviously reserved names such as LPT1. But whatever. I think we should only add checks for that if it ever becomes problematic. This should do.

pkg/filesystem/path/windows_parser.go
EdSchouten: Adding such a check would also allow you to get rid of
```go
	if strings.ContainsRune(path, '\x00') {
		return nil, status.Error(codes.InvalidArgument, "Path contains a null byte")
	}
```
Because `validateWindowsPathComponent()` also does that for you.

EdSchouten: Thinking slightly ahead: if the `strings.ContainsRune(path, '\x00')` check is removed from `NewWindowsParser()`, then it can no longer return any errors. That also removes the need for having `MustNewWindowsParser()`.

EdSchouten:
Hey @stagnation ,

I just processed the remaining comments myself and merged this change. Thanks a lot for working on this!

pkg/filesystem/path/trace.go
stagnation: Sounds good!

pkg/filesystem/path/builder.go
stagnation: Good point on the former! I did not make that connection, just thought about the error myopically. That is a great rule.

Argh windows... "But wait, there is more!"

PR #210, opened 2024-07-01
Feature/improvements to the windows filesystem implementation

stagnation:
This fixes some issues we found when developing windows runners.

### DIFF
--- a/pkg/filesystem/local_directory_windows.go
+++ b/pkg/filesystem/local_directory_windows.go
@@ -859,7 +859,7 @@ func buildFileLinkInfo(root windows.Handle, name []uint16) ([]byte, uint32) {
 
 func createNTFSHardlink(oldHandle windows.Handle, oldName string, newHandle windows.Handle, newName string) error {
 	var handle windows.Handle
-	err := ntCreateFile(&handle, windows.FILE_GENERIC_READ|windows.FILE_GENERIC_WRITE, oldHandle, oldName, windows.FILE_OPEN, 0)
+	err := ntCreateFile(&handle, windows.FILE_GENERIC_READ, oldHandle, oldName, windows.FILE_OPEN, 0)
 	if err != nil {
 		return err
 	}
--- a/pkg/filesystem/local_directory_windows.go
+++ b/pkg/filesystem/local_directory_windows.go
@@ -395,6 +395,7 @@ func (d *localDirectory) lstat(name path.Component) (FileType, error) {
 	if err != nil {
 		return FileTypeOther, err
 	}
+	defer windows.CloseHandle(handle)
 	var fileInfo windows.ByHandleFileInformation
 	err = windows.GetFileInformationByHandle(handle, &fileInfo)
 	if err != nil {
@@ -515,6 +516,7 @@ func (d *localDirectory) Readlink(name path.Component) (path.Parser, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer windows.CloseHandle(handle)
 	outBufferSize := uint32(512)
 	outBuffer := make([]byte, outBufferSize)
 	var returned uint32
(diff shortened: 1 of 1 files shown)


pkg/filesystem/local_directory_windows.go
```diff
@@ -439,50 +439,60 @@ func (d *localDirectory) Mknod(name path.Component, perm os.FileMode, deviceNumb
...
+		outBufferSize = clamp(outBufferSize*2, 62000)
+		outBuffer := make([]byte, outBufferSize)
+		err := windows.GetFileInformationByHandleEx(handle, windows.FileFullDirectoryInfo, &outBuffer[0], outBufferSize)
+		if err != nil {
+			if err.(syscall.Errno) == windows.ERROR_NO_MORE_FILES {
+				break
+			}
+			// NB: We have never seen `ERROR_MORE_DATA` during development,
+			// it seems it is never raised. But according to the docs is should be set
+			// and we should proceed with the happy path.
```
EdSchouten, line 469: That's because you've never attempted to read directories containing filenames that are more than 256 bytes long. Please remove this comment, as it is a valid case that we should handle.

pkg/filesystem/local_directory_windows.go
```diff
@@ -439,50 +439,60 @@ func (d *localDirectory) Mknod(name path.Component, perm os.FileMode, deviceNumb
...
-			break
-		}
-		if err.(syscall.Errno) == windows.ERROR_NO_MORE_FILES {
-			if firstIteration {
-				return []string{}, nil
+		// NB: In testing this was found to be about the largest buffer the API
+		// would fill. Any larger allocation would be unused and lead to
+		// resource exhaustion. We have not found good documentation for how to
+		// use this.
+		outBufferSize = clamp(outBufferSize*2, 62000)
```
EdSchouten, line 460: Instead of doing this, why don't we simply use a buffer that is twice as big as what the last call to `GetFileInformationByHandleEx()` **filled**? So if we give it a 4 KB buffer and the kernel fills the entire 4 KB, then the next time we call it with 8 KB. Then we don't need to make guesses about what the kernel's limit actually is.

pkg/filesystem/local_directory_windows.go
```diff
@@ -439,50 +439,60 @@ func (d *localDirectory) Mknod(name path.Component, perm os.FileMode, deviceNumb
...
+			dirInfoPtr = (*windowsext.FILE_FULL_DIR_INFO)(unsafe.Pointer(uintptr(unsafe.Pointer(dirInfoPtr)) + uintptr(offset)))
+
+			fileName := windows.UTF16ToString(fileNameUTF16)
+			if fileName == "." || fileName == ".." {
+				continue
+			}
+			names = append(names, fileName)
 		}
-		names = append(names, fileName)
+		continue
```
EdSchouten, line 494: This statement isn't necessary, right? A loop always continues on its own.

pkg/filesystem/local_directory_windows.go
stagnation: Aha, that makes sense. I will setup a test to verify that this works. Have you used this API before, or know of good documentation? I have mostly been tinkering with it manually to see what pops back out.


https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getfileinformationbyhandleex
https://github.com/microsoft/Windows-classic-samples/blob/main/Samples/Win7Samples/winbase/io/extendedfileapis/ExtendedFileAPIs.cpp#L468

EdSchouten: I don't know anything about it. I don't use Windows myself. ;-)

stagnation: I'll need some more time to tinker with this.
I now have a manual test that hits the `ERROR_MORE_DATA` for a 250-character filename and a small initial buffer.
Will wire that up.

Have not been able to test against the long-file path API as it is hard to actually create such files. I'm not very familiar with the special APIs either.

PR #213, opened 2024-08-15
Add new 'quorum' blob access

anguslees:
Add a new 'quorum' blob access, that provides high availability on top of already-durable storage.

Quorum blob access only requires that a subset of backends are available in order to function successfully.  The exact quorum size is configurable, but almost all cases should use "smallest integer greater than half" for both read and write quorum size.  ie: 2 out of 3 backends for single-failure tolerance, or 3 out of 5 backends for double-failure tolerance.

Writes (Put) must succeed on at least write_quorum number of backends, and occur in parallel.  Reads (Get) must see at least read_quorum number of not-found responses before concluding the blob does not exist, and occur sequentially. FindMissing reads are performed in parallel on read_quorum number of backends, and results are merged.

Note: blobs are not replicated again after the initial Put, so the underlying storage should be durable.

### DIFF
--- a/pkg/blobstore/configuration/new_blob_access.go
+++ b/pkg/blobstore/configuration/new_blob_access.go
@@ -10,6 +10,7 @@ import (
 	"github.com/buildbarn/bb-storage/pkg/blobstore"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/local"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/mirrored"
+	"github.com/buildbarn/bb-storage/pkg/blobstore/quorum"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/readcaching"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/readfallback"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/sharding"
@@ -143,6 +144,35 @@ func (nc *simpleNestedBlobAccessCreator) newNestedBlobAccessBare(configuration *
 			BlobAccess:      mirrored.NewMirroredBlobAccess(backendA.BlobAccess, backendB.BlobAccess, replicatorAToB, replicatorBToA),
 			DigestKeyFormat: backendA.DigestKeyFormat.Combine(backendB.DigestKeyFormat),
 		}, "mirrored", nil
+    case *pb.BlobAccessConfiguration_Quorum:
+        backends := make([]blobstore.BlobAccess, 0, len(backend.Quorum.Backends))
+		var combinedDigestKeyFormat *digest.KeyFormat
+
+        for _, b := range backend.Quorum.Backends {
+            backend, err := nc.NewNestedBlobAccess(b, creator)
+            if err != nil {
+                return BlobAccessInfo{}, "", err
+            }
+            backends = append(backends, backend.BlobAccess)
+            if combinedDigestKeyFormat == nil {
+                combinedDigestKeyFormat = &backend.DigestKeyFormat
+            } else {
+                newDigestKeyFormat := combinedDigestKeyFormat.Combine(backend.DigestKeyFormat)
+                combinedDigestKeyFormat = &newDigestKeyFormat
+            }
+        }
+		if len(backends) == 0 {
+			return BlobAccessInfo{}, "", status.Errorf(codes.InvalidArgument, "Cannot create quorum blob access without any backends")
+		}
+        readQuorum := int(backend.Quorum.ReadQuorum)
+        writeQuorum := int(backend.Quorum.WriteQuorum)
+        if readQuorum + writeQuorum <= len(backends) {
+			return BlobAccessInfo{}, "", status.Errorf(codes.InvalidArgument, "Quorum blob access requires read_quorum + write_quorum > number of backends")
+        }
+        return BlobAccessInfo{
+            BlobAccess: quorum.NewQuorumBlobAccess(backends, readQuorum, writeQuorum),
+            DigestKeyFormat: *combinedDigestKeyFormat,
+        }, "quorum", nil
 	case *pb.BlobAccessConfiguration_Local:
 		digestKeyFormat := digest.KeyWithInstance
 		if !backend.Local.HierarchicalInstanceNames {
--- a/pkg/blobstore/quorum/quorum_blob_access.go
+++ b/pkg/blobstore/quorum/quorum_blob_access.go
@@ -1,0 +1,218 @@
+package quorum
+
+import (
+	"context"
+	"sync"
+
+	remoteexecution "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
+	"github.com/buildbarn/bb-storage/pkg/blobstore"
+	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
+	"github.com/buildbarn/bb-storage/pkg/blobstore/slicing"
+	"github.com/buildbarn/bb-storage/pkg/digest"
+	"github.com/buildbarn/bb-storage/pkg/random"
+	"github.com/buildbarn/bb-storage/pkg/util"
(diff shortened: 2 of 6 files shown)


pkg/blobstore/configuration/new_blob_access.go
```diff
@@ -143,6 +144,35 @@ func (nc *simpleNestedBlobAccessCreator) newNestedBlobAccessBare(configuration *
...
+            if err != nil {
+                return BlobAccessInfo{}, "", err
+            }
+            backends = append(backends, backend.BlobAccess)
+            if combinedDigestKeyFormat == nil {
+                combinedDigestKeyFormat = &backend.DigestKeyFormat
+            } else {
+                newDigestKeyFormat := combinedDigestKeyFormat.Combine(backend.DigestKeyFormat)
+                combinedDigestKeyFormat = &newDigestKeyFormat
+            }
```
EdSchouten, line 162: This was probably copy-pasted from the ShardingBlobAccess code. But that code only needs to use a pointer to keep track of whether the value is actually set, because not all backends may be undrained.

My recommendation would be to not let `combinedDigestKeyFormat` be a pointer here. Then do this:
```go
for i, b := range ... {
     if i == 0 {
         combinedDigestKeyFormat = backend.DigestKeyFormat
     } else {
         combinedDigestKeyFormat = combinedDigestKeyFormat.Combine(backend.DigestKeyFormat)
     }
} 
```

pkg/blobstore/quorum/quorum_blob_access.go
```diff
@@ -0,0 +1,218 @@
...
+		backends:    backends,
+		readQuorum:  readQuorum,
+		writeQuorum: writeQuorum,
+		generator:   random.FastThreadSafeGenerator,
+	}
+}
+
+func (ba *quorumBlobAccess) shuffledBackends() []blobstore.BlobAccess {
+	backends := make([]blobstore.BlobAccess, len(ba.backends))
+	copy(backends, ba.backends)
```
EdSchouten, line 61:
```go
backends := append([]blobstore.BlobAccess(nil), ba.backends...)
```

pkg/blobstore/quorum/quorum_blob_access.go
```diff
@@ -0,0 +1,218 @@
...
+		if i == ba.writeQuorum-1 {
+			// Last writer, no need to clone buffer
+			b1 = b
+		} else {
+			b, b1 = b.CloneStream()
+		}
+
+		group.Go(func() error {
+			var err error
+			for backend := range backendQueue {
```
EdSchouten, line 167: Huh? So we launch n coroutines, and each of those attempt to send n requests?

pkg/blobstore/quorum/quorum_blob_access.go
```diff
@@ -0,0 +1,218 @@
...
+				err = backend.Put(ctx, digest, b1)
```
EdSchouten, line 168: BlobAccess.Put() takes ownership of `b1`. You can't call it again with the same buffer. The reason being that it could have converted the buffer to a reader and already consumed some of the data inside. We can't rewind, as we may be streaming data coming from the client/server directly.

pkg/blobstore/quorum/quorum_blob_access.go
```diff
@@ -0,0 +1,218 @@
...
+	if err := group.Wait(); err != nil {
+		return digest.EmptySet, err
+	}
+
+	// Find intersection of all results
+	missingFromAll := results[0]
+	for _, result := range results[1:] {
+		_, missingFromAll, _ = digest.GetDifferenceAndIntersection(missingFromAll, result)
+	}
+	return missingFromAll, nil
```
EdSchouten, line 211: Imagine I have 5 storage backends, named 0 to 4. Read quorum is set to 4.

- I first issue a Put(). Due to network flakiness, the blob only ends up getting written to backend 0.-
- Then I issue FindMissingBlobs(), which gets run against backends 0, 1, 2, and 3. Backend 0 reports it as present, while backends 1, 2 and 3 report it as absent. As we compute the intersection, we report it as being present.
- Then I issue Get(), which gets run against backends 1, 2, 3 and 4. All of those return NOT_FOUND.

This is in violation of the protocol.

pkg/blobstore/quorum/quorum_blob_access.go
```diff
@@ -0,0 +1,218 @@
...
+		&getQuorumErrorHandler{
+			remainingBackends: remainingBackends,
+			remainingReads:    ba.readQuorum,
+			retry:             getter,
+		})
+}
+
+func (ba *quorumBlobAccess) Get(ctx context.Context, digest digest.Digest) buffer.Buffer {
+	return ba.get(func(b blobstore.BlobAccess) buffer.Buffer {
+		return b.Get(ctx, digest)
```
EdSchouten, line 125: How do you deal with servers being down in such a way that calls to Get() spend a lot of time hanging, e.g. waiting for a TCP connection to establish?

EdSchouten (commented):
To me it's unclear why one would want to use a backend like this. The way MirroredBlobAccess are designed/implemented right now, you get fairly decent storage density. You "only" lose 50% of space. Well, slightly more if you want to keep some nodes on hot standby. It's possible to lose less than 50% if you resort to Reed-Solomon encoding, but for the small objects we often have to work with that's likely too much overhead. This backend only lets you achieve 100%/n = 33%, 25%, 20%, .... I fail to see how that's economic.

The reason MirroredBlobAccess requires both backends to be online is because we rely on battle tested tools like Kubernetes to do the health checking/backend migration/service resolution for us. The advantage of that is that Buildbarn can stay simple, and that you as a cluster administrator can use tools like `kubectl`, `kube-state-metrics`, etc. to get decent insight in what's going on.

Now you may think: "I use bare metal hardware. Having a dependency on Kubernetes is unreasonable." Well, my response to that is that you can keep your cluster as simple as you want. It doesn't need to be as complete as the stock clusters offered by cloud providers. You can get pretty far by running etcd, kube-apiserver, kube-controller-manager and kube-scheduler on a single server, and kubelet on all of your storage nodes. There is no need to set up a [CNI](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/), nor do you need kube-proxy/coredns, as [Buildbarn binaries are nowadays capable of resolving endpoints using kuberesolver](https://github.com/buildbarn/bb-storage/commit/a4267fc3c5c3a916004c5021fb13bc2bcf214e05).

EdSchouten:
> Note: blobs are not replicated again after the initial Put, so the underlying storage should be durable.

I don't think that's a very desirable property. What should users do if the underlying storage nodes do get out of sync (e.g., due to disks in a server getting replaced)? Do we provide any tools to do the copying?

pkg/blobstore/configuration/new_blob_access.go
```diff
@@ -143,6 +144,35 @@ func (nc *simpleNestedBlobAccessCreator) newNestedBlobAccessBare(configuration *
...
+        }
+		if len(backends) == 0 {
+			return BlobAccessInfo{}, "", status.Errorf(codes.InvalidArgument, "Cannot create quorum blob access without any backends")
+		}
+        readQuorum := int(backend.Quorum.ReadQuorum)
+        writeQuorum := int(backend.Quorum.WriteQuorum)
+        if readQuorum + writeQuorum <= len(backends) {
+			return BlobAccessInfo{}, "", status.Errorf(codes.InvalidArgument, "Quorum blob access requires read_quorum + write_quorum > number of backends")
+        }
```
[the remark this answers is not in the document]
anguslees: Yeah, I felt it was far more likely for someone to mistakenly put a small negative number in here rather than a super large positive number - so I went with a uint in the proto.  I agree I don't like casting int types and I'm happy to change this to a int throughout. (Or whatever other solution you would prefer)

PR #217, opened 2024-09-10
LocalBlobAccess: Double the recommended number of attempts

EdSchouten:
Right now we set the maximum get attempts to just eight. This might be a bit too tight, requiring users to create KeyLocationMaps that are unnecessarily big. Given that nobody has really complained about these values being too large, let's bump them.


moroten (changes requested):
Thank you for a good discussion. Please also update `doc/zh_CN/README.md`.

doc/zh_CN/README.md
```diff
@@ -31,8 +31,8 @@ $ cat config/bb_storage.jsonnet
             sizeBytes: 16 * 1024 * 1024,
           },
         },
-        keyLocationMapMaximumGetAttempts: 8,
-        keyLocationMapMaximumPutAttempts: 32,
+        keyLocationMapMaximumGetAttempts: 16,
+        keyLocationMapMaximumPutAttempts: 64,
```
moroten, line 35: Line 63-64 also need an update.

moroten (changes requested):

PR #226, opened 2024-11-18
Emit additional metrics on replicate blobs and refresh blobs

kevinye202:
This pr adds several more metrics to replicate blobs and refresh blobs, specifically batch size and blob size in bytes as well as latency, which would provide more insight in the event of slowness in findMissing and replication.


pkg/blobstore/local/flat_blob_access.go
```diff
@@ -28,6 +29,26 @@ var (
 			Buckets:   append([]float64{0}, prometheus.ExponentialBuckets(1.0, 2.0, 16)...),
 		},
 		[]string{"storage_type", "operation"})
+
+	flatBlobAccessRefreshLatencySeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "buildbarn",
+			Subsystem: "blobstore",
+			Name:      "flat_blob_access_refresh_latency_seconds",
```
EdSchouten, line 37: Here we use `_latency_` as a midfix, whereas the rest of our codebase uses `_duration_`. Could we please make this more consistent?

Also on the MetricsBlobAccess side, we use `_operations_` (plural). Probably good to stick to `_refreshes_` for consistency?

pkg/blobstore/local/flat_blob_access.go
```diff
@@ -28,6 +29,26 @@ var (
 			Buckets:   append([]float64{0}, prometheus.ExponentialBuckets(1.0, 2.0, 16)...),
```
EdSchouten, line 29: Could we please rename this to `flat_blob_access_refreshes_blobs`?

pkg/blobstore/local/flat_blob_access.go
```diff
@@ -128,6 +167,8 @@ func (ba *flatBlobAccess) Get(ctx context.Context, blobDigest digest.Digest) buf
 		return b
 	}
 
+	ba.refreshSizeGet.Observe(float64(location.SizeBytes))
```
EdSchouten, line 170: Any reason we don't do this next to the other `.Observe()` calls?

pkg/blobstore/local/flat_blob_access.go
```diff
@@ -361,6 +409,7 @@ func (ba *flatBlobAccess) FindMissing(ctx context.Context, digests digest.Set) (
 				// Blob is present and still needs to be
 				// refreshed. Allocate space for a copy.
 				b := getter(blobToRefresh.digest)
+				ba.refreshSizeFindMissing.Observe(float64(location.SizeBytes))
```
EdSchouten, line 412: I think it would be more valuable to aggregate the sizes and report it for the refresh as a whole, because then these metrics can be used together with the other metrics we have for refreshing. Now the number of samples within will differ.

pkg/blobstore/replication/metrics_blob_replicator.go
```diff
@@ -0,0 +1,165 @@
...
+)
+
+var (
+	replicatorOperationsPrometheusMetrics sync.Once
+
+	replicatorOperationsDurationSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "buildbarn",
+			Subsystem: "blobstore",
+			Name:      "replicator_operations_duration_seconds",
```
EdSchouten, line 25: `blob_replicator_operations*` for consistency with `blob_access_operations_*` ?

pkg/blobstore/replication/metrics_blob_replicator.go
```diff
@@ -0,0 +1,165 @@
...
+	compositeDurationSeconds prometheus.ObserverVec
+	compositeBlobSizeBytes   prometheus.Observer
+	multipleDurationSeconds  prometheus.ObserverVec
+	multipleBatchSize        prometheus.Observer
+	multipleBlobSizeBytes    prometheus.Observer
+}
+
+// NewMetricsBlobReplicator creates a wrapper around BlobReplicator that adds
+// Prometheus metrics for monitoring replication operations.
+func NewMetricsBlobReplicator(replicator BlobReplicator, clock clock.Clock) BlobReplicator {
```
EdSchouten, line 69: Right now these metrics aren't broken down by backend/storage type. Is that intentional? If not, maybe we can simply use the backend type of the BlobAccess that they are used in conjunction with?

EdSchouten (commented):
Thanks for working on this. Really appreciated!

pkg/blobstore/replication/metrics_blob_replicator.go
kevinye202: done.

pkg/blobstore/replication/metrics_blob_replicator.go
kevinye202: added storage type, I didn't see a good way to extract backendType from blobAccess though.

pkg/blobstore/local/flat_blob_access.go
kevinye202: done.

cmd/bb_replicator/main.go
```diff
@@ -53,7 +53,9 @@ func main() {
 			configuration.Replicator,
 			source.BlobAccess,
 			sink,
-			blobstore_configuration.NewCASBlobReplicatorCreator(grpcClientFactory))
+			blobstore_configuration.NewCASBlobReplicatorCreator(grpcClientFactory),
+			"cas",
```
EdSchouten, line 57: Probably better to call `blobAccessCreator.GetStorageTypeName()` here, to make sure it doesn't go out of sync?

pkg/blobstore/configuration/cas_blob_replicator_creator.go
```diff
@@ -27,7 +27,7 @@ func NewCASBlobReplicatorCreator(grpcClientFactory grpc.ClientFactory) BlobRepli
 func (brc *casBlobReplicatorCreator) NewCustomBlobReplicator(configuration *pb.BlobReplicatorConfiguration, source blobstore.BlobAccess, sink BlobAccessInfo) (replication.BlobReplicator, error) {
 	switch mode := configuration.Mode.(type) {
 	case *pb.BlobReplicatorConfiguration_Deduplicating:
-		base, err := NewBlobReplicatorFromConfiguration(mode.Deduplicating, source, sink, brc)
+		base, err := NewBlobReplicatorFromConfiguration(mode.Deduplicating, source, sink, brc, "cas")
```
EdSchouten, line 30: Probably better if we move `GetStorageTypeName()` to implementations of `BlobReplicatorCreator`? Because then we can call that function from here as well.

pkg/blobstore/local/flat_blob_access.go
EdSchouten: Please also rename the variables/struct fields for consistency.

pkg/blobstore/local/flat_blob_access.go
```diff
@@ -351,8 +397,10 @@ func (ba *flatBlobAccess) FindMissing(ctx context.Context, digests digest.Set) (
 	// one thread.
 	ba.refreshLock.Lock()
 	defer ba.refreshLock.Unlock()
-
+	// Add refresh start time before the refresh loop
+	refreshStart := time.Now()
 	blobsRefreshedSuccessfully := 0
+	blobRefreshSizeBytes := 0
```
EdSchouten, line 403: Please change this to `int64`, so that the `int()` conversion isn't needed. It's fairly easy to overflow `int` on 32-bit architectures.

pkg/blobstore/configuration/cas_blob_replicator_creator.go
kevinye202: Good call, moved it there.

pkg/blobstore/configuration/proto_blob_replicator_creator.go
```diff
@@ -14,3 +14,7 @@ type protoBlobReplicatorCreator struct{}
 func (brc protoBlobReplicatorCreator) NewCustomBlobReplicator(configuration *pb.BlobReplicatorConfiguration, source blobstore.BlobAccess, sink BlobAccessInfo) (replication.BlobReplicator, error) {
 	return nil, status.Error(codes.InvalidArgument, "Configuration did not contain a supported replicator")
 }
+
+func (brc protoBlobReplicatorCreator) GetStorageTypeName() string {
```
EdSchouten, line 18: Please remove this, as it is not valid to return an empty string.  Any type that is based on `protoBlobReplicatorCreator` should bring its own implementation.

pkg/blobstore/configuration/blob_replicator_creator.go
```diff
@@ -16,4 +16,8 @@ type BlobReplicatorCreator interface {
 	// type. For example, sending replication requests over gRPC is
 	// only supported for the Content Addressable Storage.
 	NewCustomBlobReplicator(configuration *pb.BlobReplicatorConfiguration, source blobstore.BlobAccess, sink BlobAccessInfo) (replication.BlobReplicator, error)
+
+	// GetStorageTypeName() returns the name of the storage type that
+	// this BlobReplicatorCreator is able to create BlobReplicators for.
+	GetStorageTypeName() string
```
EdSchouten, line 22: Note that BlobAccessCreator inherits BlobReplicatorCreator. Please remove GetStorageTypeName() from the latter, and also remove the implementations from {cas,ac,...}BlobAccessCreator.

pkg/blobstore/local/flat_blob_access.go
```diff
@@ -19,15 +20,35 @@ import (
...
 			Namespace: "buildbarn",
 			Subsystem: "blobstore",
-			Name:      "flat_blob_access_refreshes",
+			Name:      "flat_blob_access_refreshes_blobs",
 			Help:      "The number of blobs that were refreshed when requested",
 			Buckets:   append([]float64{0}, prometheus.ExponentialBuckets(1.0, 2.0, 16)...),
 		},
 		[]string{"storage_type", "operation"})
+
+	flatBlobAccessRefreshesLatencySeconds = prometheus.NewHistogramVec(
```
EdSchouten, line 33: s/Latency/Duration/

pkg/blobstore/local/flat_blob_access.go
```diff
@@ -43,6 +64,13 @@ type flatBlobAccess struct {
 	refreshesGet              prometheus.Observer
 	refreshesGetFromComposite prometheus.Observer
 	refreshesFindMissing      prometheus.Observer
```
EdSchouten, line 66: s/refreshes/refreshesBlobs/

pkg/blobstore/local/flat_blob_access.go
```diff
@@ -43,6 +64,13 @@ type flatBlobAccess struct {
...
+
+	refreshesLatencyGet              prometheus.Observer
```
EdSchouten, line 68: s/Latency/Duration/

pkg/blobstore/replication/metrics_blob_replicator.go
```diff
@@ -0,0 +1,162 @@
...
+	"sync"
+	"time"
+
+	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
+	"github.com/buildbarn/bb-storage/pkg/blobstore/slicing"
+	"github.com/buildbarn/bb-storage/pkg/clock"
+	"github.com/buildbarn/bb-storage/pkg/digest"
+	"github.com/buildbarn/bb-storage/pkg/util"
+	"github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/grpc/codes"
```
EdSchouten, line 14: Please group imports by 2nd level domain, for consistency with the rest of the codebase. In other words, add an empty above this one.

pkg/blobstore/configuration/blob_replicator_creator.go
kevinye202: I see, I removed it from BlobAccessCreator, but I don't think I can remove it from the {cas,ac,...}BlobAccessCreators? since they don't inherit from the respective replicators, and it still needs the implementation to make it a legit BlobAccessCreators?

kevinye202: nvm, i did see that it inherits the replicator, removing it.

PR #229, opened 2024-12-09
add missing label on replicator metrics and unit test

kevinye202:
The storage type label is missing from metrics blob replicator, causing the initialization to fail on: 
panic: 1 unknown label(s) found during currying

goroutine 1 [running]:
github.com/prometheus/client_golang/prometheus.(*HistogramVec).MustCurryWith(...)
	external/gazelle~~go_deps~com_github_prometheus_client_golang/prometheus/histogram.go:1278

Also, add a unit test to catch this case.


EdSchouten:
Thanks a lot, @kevinye202!

PR #234, opened 2024-12-20
Remote authenticator and authorizer

moroten:
The authenticate and authorize tasks can now be sent remotely over gRPC to an external service. This way, custom authentication and authorization does not require a modified builds of the Buildbarn components.

To avoid spamming the remote service with calls for every REv2 request and keep the latency low, the verdicts, both allow and deny, are cached for a duration specified in the response from the remote service.


moroten:
@EdSchouten Welcome back from Christmas holiday. I forgot to bring this up for discussion yesterday. Any comments? No hurry, take your time.

pkg/proto/auth/auth.proto
```diff
@@ -27,3 +29,73 @@ message AuthenticationMetadata {
...
+  oneof verdict {
+    // Allow the request.
+    //
+    // The value of this field is used as authentication metadata in
+    // literal form.
+    AuthenticationMetadata allow = 1;
+
+    // Deny the request, returning a fixed error message back
+    // to the client.
+    string deny = 2;
```
EdSchouten, line 63: Maybe good to document what kind of status code should be sent back to the client? Should it always be the equivalent of `Unauthenticated`? Same with `AuthorizeResponse.deny`.

pkg/proto/auth/auth.proto
```diff
@@ -27,3 +29,73 @@ message AuthenticationMetadata {
...
+
+message AuthenticateRequest {
+  message ValueList {
+    repeated string value = 1;
+  }
+
+  // The metadata is forwarded as part of the request as accessing the
+  // authentication service might have its own authentication and authorizing
+  // system.
+  map<string, ValueList> request_metadata = 1;
```
EdSchouten, line 46: So this is the equivalent of HTTP or gRPC request headers? Maybe good to explicitly state that?

pkg/grpc/remote_authenticator.go
```diff
@@ -0,0 +1,181 @@
...
+	// keeping credentials in the memory.
+	requestKey := sha256.Sum256(requestBytes)
+
+	a.lock.Lock()
+	now := a.clock.Now()
+	entry := a.getAndTouchCacheEntry(requestKey)
+	if entry != nil && entry.HasExpired(now) {
+		entry = nil
+	}
+	if entry == nil {
```
EdSchouten, line 104:
```go
if entry == nil || entry.HasExpired(now) {
```
?

pkg/grpc/remote_authenticator.go
```diff
@@ -0,0 +1,181 @@
...
+		// Perform the remote authentication request.
+		entry.response = a.authenticateRemotely(ctx, request)
+		close(responseReady)
+	} else {
+		a.lock.Unlock()
+
+		// Wait for the remote request to finish.
+		select {
+		case <-ctx.Done():
+			return nil, util.StatusWrapWithCode(ctx.Err(), codes.Unauthenticated, "Context cancelled")
```
EdSchouten, line 123:
```
return nil, util.StatusFromContext(ctx)
```

pkg/grpc/remote_authenticator.go
```diff
@@ -0,0 +1,181 @@
...
+		case <-entry.ready:
+			// Noop
+		}
+	}
+	return entry.response.authMetadata, entry.response.err
```
EdSchouten, line 128: I don't think this is sound. What happens if `entry.response.err` is of type `codes.Canceled`? We end up returning a cancelation error belonging to a different client.

pkg/grpc/remote_grpc_request_authenticator.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"context"
+
+	"github.com/buildbarn/bb-storage/pkg/auth"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+type remoteGrpcRequestAuthenticator struct {
```
EdSchouten, line 12: I'd recommend removing Grpc from the name. That it's for gRPC is already implied based on the package name.

pkg/grpc/remote_grpc_request_authenticator_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+func TestRemoteGrpcRequestAuthenticator(t *testing.T) {
+	ctrl, ctx := gomock.WithContext(context.Background(), t)
+	md := metadata.New(
+		map[string]string{
+			"Authorization": "token",
```
EdSchouten, line 24: Doesn't gRPC metadata use lowercase header names?

pkg/grpc/remote_authorizer.go
```diff
@@ -0,0 +1,171 @@
+package grpc
```
EdSchouten, line 1: I understand why you put this code here, though I would argue that `pkg/auth/remote_authorizer.go` would be a better spot for this. Or if this ends up introducing cycles, add a `pkg/auth/remote` or something.

The idea is that `pkg/auth` contains all authorization related code that can be used irrespective of the kind of transport that was used to make the incoming connection. In our case we can use RemoteAuthorizer both with incoming gRPC and HTTP requests.

What we have in `pkg/http` and `pkg/grpc` is the authentication related code, which is quite protocol specific (due to it receiving metadata differently and returning different styles of errors).

pkg/grpc/request_headers_authenticator.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"context"
+
+	"github.com/buildbarn/bb-storage/pkg/auth"
+)
+
+// RequestHeadersAuthenticator can be used to grant or deny access to a server
+// based on request headers, typically from an HTTP or gRPC request.
+type RequestHeadersAuthenticator interface {
```
EdSchouten, line 11: I think this should live in pkg/auth, as it's not specific to any specific transport protocol, right? Just ones that "provide string key/value headers".

pkg/grpc/remote_authenticator.go
moroten: Would it be correct to use `contect.Background()` in the request?

EdSchouten: That could work, but has the downside that cancelation isn't propagated immediately. My recommendation would be to stick to `ctx`, but don't coalesce error results of in-flight RPCs.

pkg/grpc/remote_grpc_request_authenticator_test.go
moroten: The Go gRPC library is case insensitive and translates them to lower case.

EdSchouten: Ah, gotcha. It looks like `metadata.New()` does the normalization to lowercase for you, so it's fine.

pkg/grpc/remote_authorizer.go
moroten: I did have a cycle in the beginning of the development, but apparently not any more.

moroten:
I've now rebased and fixed all the review comments. The changes were done in `Add remote auth .proto definitions` and `Implement remote authenticator and authorizer`.

moroten:
@EdSchouten friendly ping. Would you like me to click "Resolve conversation"?

pkg/http/remote_request_authenticator.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"net/http"
+
+	"github.com/buildbarn/bb-storage/pkg/auth"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type remoteRequestAuthenticator struct {
```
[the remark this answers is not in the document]
moroten: > This also makes me wonder: should we reimplement our JWT auth to be built on top of this interface as well?

Is that just to extract a single header into the JWT handler? Maybe I didn't get how you meant.

pkg/auth/remote_authenticator.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	// Hash the request to use as a cache key to both save memory and avoid
+	// keeping credentials in the memory.
+	requestKey := sha256.Sum256(requestBytes)
+
+	now := a.clock.Now()
+	for {
+		a.lock.Lock()
+		entry := a.getAndTouchCacheEntry(requestKey)
+		if entry == nil || (entry.IsReady() && !entry.IsValid(now)) {
```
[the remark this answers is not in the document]
moroten: Created two maps, much better. Thank you.

pkg/auth/remote_authenticator.go
```diff
@@ -0,0 +1,196 @@
...
+		// Error response on the remote request, make a new request.
+		return false
+	}
+	return now.Before(ce.response.expirationTime)
+}
+
+// NewRemoteAuthenticator creates a new RemoteAuthenticator for incoming
+// requests that forwards headers to a remote service for authentication. The
+// result from the remote service is cached.
+func NewRemoteAuthenticator(
```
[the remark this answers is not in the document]
moroten: Done

pkg/auth/remote_authenticator.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// IsValid returns false if a new remote request should be made.
+func (ce *remoteAuthCacheEntry) IsValid(now time.Time) bool {
+	if ce.response == nil {
+		// Error response on the remote request, make a new request.
+		return false
+	}
+	return now.Before(ce.response.expirationTime)
+}
```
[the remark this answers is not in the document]
moroten: Removed and created two maps.

pkg/http/remote_request_authenticator.go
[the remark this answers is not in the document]
EdSchouten: Right now we have these:

```
-rw-r--r--   1 ed  staff   1240 Dec 15 10:38 pkg/grpc/jwt_authenticator.go
-rw-r--r--   1 ed  staff   1048 Dec 15 10:38 pkg/http/jwt_authenticator.go
```

But considering that we now have pkg/{grpc,http}/request_header_authenticator.go, we no longer need these, right? We could throw them away and provide a single implementation of `auth.RequestHeadersAuthenticator` that does the same thing.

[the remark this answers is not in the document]
moroten: Now, I get it. I've pushed an update which I've verified in bb-deployments docker-compose for the gRPC header handling.

pkg/auth/remote_authorizer.go
```diff
@@ -0,0 +1,187 @@
...
+	requestBytes, err := proto.Marshal(request)
+	if err != nil {
+		return util.StatusWrapWithCode(err, codes.PermissionDenied, "Failed to marshal authorize request")
+	}
+	// Hash the request to use as a cache key to save memory.
+	requestKey := sha256.Sum256(requestBytes)
+
+	for {
+		a.lock.Lock()
+		now := a.clock.Now()
```
EdSchouten, line 92: Probably worth calling `a.clock.Now()` before acquiring the lock.

pkg/auth/remote_authorizer.go
```diff
@@ -0,0 +1,187 @@
...
+	return remoteAuthorizerResponse{}, false
+}
+
+func (a *remoteAuthorizer) insertCacheEntry(requestKey RemoteAuthorizerCacheKey, response remoteAuthorizerResponse) {
+	if _, ok := a.cachedResponses[requestKey]; ok {
+		a.evictionSet.Touch(requestKey)
+	} else {
+		// Cache contains no matching entry. Free up space, so that the
+		// caller may insert a new entry.
+		for len(a.cachedResponses) >= a.maximumCacheSize {
```
EdSchouten, line 152: Nit:
```go
for len(a.cachedResponses) > 0 && ....
```
To prevent crashes if someone sets `maximumCacheSize` to zero.

pkg/auth/remote_authorizer.go
```diff
@@ -0,0 +1,187 @@
...
+		expirationTime: time.Time{},
+	}
+
+	response, err := a.remoteAuthClient.Authorize(ctx, request)
+	if err != nil {
+		return ret, util.StatusWrapWithCode(err, codes.PermissionDenied, "Remote authorization failed")
+	}
+
+	// An invalid expiration time indicates that the response should not be cached.
+	if response.GetCacheExpirationTime().IsValid() {
```
EdSchouten, line 173: Please only use `.Get*()` in case the containing message may be nil. In this specific case `response` is guaranteed to be non-nil, so use `.CacheExpirationTime` instead.

Also instead of using `.IsValid()`, I'd recommend comparing against `nil` and having a separate `.CheckValid()` to return an error. The downside of the current approach is that in case some incompatibility arises in the way timestamps are sent over, the client will silently discard it.

pkg/proto/configuration/http/http.proto
```diff
@@ -177,3 +185,12 @@ message AcceptHeaderAuthenticationPolicy {
...
   AuthenticationPolicy policy = 2;
 }
+
+message RemoteHttpRequestAuthenticationPolicy {
+  // The remote authenticator to grant or deny access the HTTP request.
+  buildbarn.configuration.grpc.RemoteAuthenticationPolicy backend = 1;
+
+  // Headers from the incoming HTTP request to forward to the remote
+  // authenticator.
+  repeated string headers = 2;
```
EdSchouten, line 195: Why isn't this field part of `RemoteAuthenticationPolicy`?

pkg/proto/configuration/http/http.proto
```diff
@@ -177,3 +185,12 @@ message AcceptHeaderAuthenticationPolicy {
   // provided.
   AuthenticationPolicy policy = 2;
 }
+
+message RemoteHttpRequestAuthenticationPolicy {
+  // The remote authenticator to grant or deny access the HTTP request.
+  buildbarn.configuration.grpc.RemoteAuthenticationPolicy backend = 1;
```
EdSchouten, line 191: Shouldn't this message live in buildbarn.configuration.auth, because it's independent of gRPC/HTTP?

pkg/proto/configuration/grpc/grpc.proto
```diff
@@ -365,6 +366,13 @@ message AuthenticationPolicy {
     // LOCAL_PEERCRED is used, which is typically capable of returning
     // up to 16 group memberships.
     string peer_credentials_jmespath_expression = 6;
+
+    // Forwards the authentication task to a remote server. This can be a
+    // convenient way to integrate custom authentication processes, for example
+    // if a proxy has already performed authentication where the resulting HTTP
+    // request's headers need to be verified and processed in a custom way to
+    // produce buildbarn.auth.AuthenticationMetadata.
+    RemoteGrpcRequestAuthenticationPolicy remote = 8;
```
EdSchouten, line 375: Would it make sense to replace this field and the `jwt` with one called `request_headers` or something, containing a `oneof` for `remote` and `jwt` inside? That way people can add new request header based authenticators without requiring further code changes to HTTP and gRPC.

pkg/auth/remote_authorizer.go
```diff
@@ -0,0 +1,187 @@
...
+			case <-responseReady:
+				a.lock.Lock()
+				// Check whether the remote authentication call succeeded.
+				response, ok := a.cachedResponses[requestKey]
+				a.lock.Unlock()
+				if ok {
+					// Note that the expiration time is not checked, as the
+					// response is as fresh as it can be.
+					return response.err
+				}
```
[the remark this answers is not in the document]
moroten: If the remote doesn't set any expiry, all requests will basically be queued and processed one at a time. I think it's better to either keep the logic or remove the deduplication altogether.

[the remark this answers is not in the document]
EdSchouten: Aaah, gotcha. The caching is optional. We don't fill in a default value in that case. We should indeed leave the code as is then.

One slight concern I have though: for setups that don't do caching we will still attempt to do deduplication of requests with the same request key. Doesn't that lead to excessive serialization/contention?

pkg/proto/configuration/grpc/grpc.proto
moroten: Looking at the factory code, it makes sense.

pkg/proto/configuration/http/http.proto
moroten: This creates a circular dependency between `grpc.proto` and `auth.proto`: `grpc.ServerConfiguration -> grpc.AuthenticationPolicy -> auth.RemoteAuthenticator -> grpc.ClientConfiguration`. If we split into `grpc_client.proto` and `grpc_server.proto`, it will work. Shall I do that?

pkg/proto/configuration/http/http.proto
moroten: In practive, it is the same. In the implementation, `headers` is used in a different layer, extracting the headers before going into the remote authorization. I suppose it's fine for the configuration code to split the structure and pass some variables to one constructor and other variables to another constructor. Better to have simplified configuration.

pkg/auth/remote_authorizer.go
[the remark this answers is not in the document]
moroten: Do you prefer keeping the logic or removing the deduplication of requests?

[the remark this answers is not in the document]
moroten: (I don't know why my UI did not show your last comment.)

I agree. I'll add an `if a.maximumCacheSize == 0` to skip the deduplication code.

moroten:
It would be nice to put `RequestHeadersAuthenticator` and `RemoteAuthenticator` in `auth.proto`, but that creates the following dependency cycle between `grpc.proto` and `auth.proto`:
`grpc.ServerConfiguration -> grpc.AuthenticationPolicy -> auth.RequestHeadersAuthenticator -> auth.RemoteAuthenticator -> grpc.ClientConfiguration`. This can be solved by splitting `grpc.proto` into `grpc_client.proto` and `grpc_server.proto`.

Both `grpc.ServerConfiguration` and `grpc.ClientConfiguration` use `grpc.TracingMethodConfiguration`. Where shall that one be moved to `grpc_client.proto` as well or to an independent `grpc_tracing.proto`?

It starts to become a bit of a refactoring, but I'm up for doing it if you think this is the way to do it.

Just a detail, shall the name be `pkg/proto/configuration/grpc_client/grpc_client.proto` with the go package name with underscore? Protobuf suggests with underscore but Go suggests without in package names.

pkg/auth/remote_authorizer.go
```diff
@@ -0,0 +1,199 @@
...
+
+	for {
+		now := a.clock.Now()
+		a.lock.Lock()
+		if response, ok := a.getAndTouchCacheEntry(requestKey); ok && response.expirationTime.After(now) {
+			a.lock.Unlock()
+			return response.err
+		}
+		// No valid cache entry available. Deduplicate requests.
+		if responseReady, ok := a.pendingRequests[requestKey]; ok {
```
EdSchouten, line 107: Tiny nit: for readability, I'd appreciate it that we wrote code like this:

```go
if foo {
    A
    // no return here
} else {
   B
   return
}
```
As this instead:
```go
if !foo {
   B
   return
}
A
```

This makes it more explicit that the `select { case <-responseReady: ... }` falls through and causes the loop to run again, as it'll now be closer to the bottom of the loop.

pkg/http/request_headers_authenticator_test.go
```diff
@@ -0,0 +1,95 @@
...
+	auth_pb "github.com/buildbarn/bb-storage/pkg/proto/auth"
+	"github.com/buildbarn/bb-storage/pkg/testutil"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+func TestRemoteHttpRequestAuthenticator(t *testing.T) {
```
EdSchouten, line 20: s/Http// to match the type?

EdSchouten:
> It would be nice to put `RequestHeadersAuthenticator` and `RemoteAuthenticator` in `auth.proto`, but that creates the following dependency cycle between `grpc.proto` and `auth.proto`: `grpc.ServerConfiguration -> grpc.AuthenticationPolicy -> auth.RequestHeadersAuthenticator -> auth.RemoteAuthenticator -> grpc.ClientConfiguration`. This can be solved by splitting `grpc.proto` into `grpc_client.proto` and `grpc_server.proto`.

Ugh! Well, let's not go that far for now. Let's just accept that there's a minimal amount of overlap between the two then. Can you at least add a TODO/comment/... something to the code to document that this overlap exists?

EdSchouten (approved):
Thanks a lot for working on this, @moroten. Really appreciated.

I have some final (small) remarks. Be sure to address those, and I'll merge this swiftly.

pkg/auth/remote_authorizer.go
moroten: Also changed
```
select {
case <-ctx.Done():
    return ctx.Err()
case <-wakeupChan:
    A
}
```
into
```
select {
case <-ctx.Done():
    return ctx.Err()
case <-wakeupChan:
}
A
```

PR #238, opened 2025-02-03
Implement sharding based on Rendezvous Hashing

meroton-benjamin:
Improves the ability to perform resharding by switching the sharding
struct from array to map. Each map entry has a key which is used in
rendezvous hashing to deterministically select which shard to use from
the collection of keys. When a shard is removed it is guaranteed that
only blobs which belonged to the removed shard will resolve to a new
shard.

In combination with ReadFallbackConfigurations this allows adding and
removing shards with minimal need to rebalance the blobs between the
shards.

For more details and migration instructions, see:
https://github.com/buildbarn/bb-adrs/blob/bf2066633e1712a3ef7c295d37cd52e65867391d/0011-rendezvous-hashing.md


pkg/blobstore/sharding/rendezvous_shard_selector.go
```diff
@@ -0,0 +1,64 @@
...
+		keyMap:	keyMap,
+		weightMap: weightMap,
+	}
+}
+
+func score(x uint64) float64 {
+	// branchless clamp to [1,MAX_UINT64-1]
+	x = x - ((x|-x)>>63) + (((^x)|-(^x)) >> 63)
+	frac := float64(x)/float64(^uint64(0))
+	return 1.0/-math.Log(frac)
```
EdSchouten, line 38: I'm trying to understand the math behind this.

- Why do we need the clamping? I understand math.Log(0) doesn't work as expected, but what's wrong with MaxUint64?
- Why do we need the division by MaxUint64? Considering that we're applying a log() afterwards, wouldn't that be the same as computing the log() first, but then subtracting a constant?
- Why do we need the natural logarithm? Would the algorithm work equally well with log2()?
- If log2() works, would there be any way to simply approximate the result without relying on any floating point math?

pkg/blobstore/sharding/rendezvous_shard_selector.go
```diff
@@ -0,0 +1,64 @@
...
+	var bestKey string
+
+	for keyHash, weight := range s.weightMap {
+		mixed := splitmix64(hash^keyHash)
+		current := float64(weight) * score(mixed)
+		if current > best {
+			best = current
+			bestKey = s.keyMap[keyHash]
+		}
+	}
```
EdSchouten, line 62: Because we are simply iterating over `weightMap`, I don't think we actually need to use maps here. Have you considered using a simple slice?
```go
type rendezvousShard struct {
    keyHash uint64
    weight uint32
    key string
}

type rendezvousShardSelector struct {
    shards []rendezvousShard
}
```

Assuming this works, is it then still necessary to change ShardingBlobAccess to take a map? This function could return an index of the shard like it did before.

pkg/blobstore/sharding/rendezvous_shard_selector.go
meroton-benjamin: 1. MaxUint64 would give log(1)=0 which would be  a division by zero.

However this is ieee floating point math and not real math, weight/-log(1) = -Inf which would be infinitely undesired, while while we would expect it to be infinitely desired. Thinking more about it log(0) seems to be defined as -Inf which would make the quota weight/-(-Inf) = +0 in IEEE math which is the result we actually want. If my math is correct then this means the code actually handles the extreme values in a somewhat decent manner with the MaxUint64 essentially wrapping around. 

2. The original proof uses log of a random uniform ]0,1[ distribution, so approximating the uniform distribution with random_uint64 / max_uint64 we could do the equivalent instead of 1.0/(log(max_uint64)-log(x))

3. The original proof uses the natural logarithm, log2 is just a constant away from log and we're only interested in their relative ordering rather than the score itself so we can use log2 without any difference.

4. There are some possible log2 bit trickery hacks, such approximating the logarithm by its integer part (i.e. number of set bits) + a polynomial that approximates its fractional part, technically I think this is the actual implementation of the the standard log algorithm just that the chosen  polynomial is of high enough order to give the 15 or so bits of precision possible in float64.

Since we don't have to support the entire range of positive float64s and we don't need the full precision we could probably get something slightly faster, maybe by abusing the float64 mantissa in it's binary representation. I'll have a look if I can get something which produces sufficient precision faster.

pkg/blobstore/sharding/rendezvous_shard_selector.go
meroton-benjamin: The implementation can iterate over a slice of shards, that would make the tiebreaker in case of a collision result into a stable value (biased towards earlier in the slice) and significantly reduce the changes in sharding_blob_access.go.

I still think it's better to replace the interface of ShardingBlobAccessConfiguration.Shards with ShardingBlobAccessConfiguration.ShardMap since a gives a natural interface for removing and inserting new shards compared to an array.

We could also error out at startup when there is a hash collision between two shards, it should never happen but maybe erroring out is preferable to ignoring an entire shard (which would be the effect of two shards sharing a hash).

meroton-benjamin:
The newest implementation integer math with fixed point calculations. Log2 is calculated with the integer part from the number of bits + a combination of a lookup table with linear interpolation for values.

I have evaluated the error by running a billion entries over 5 shards of weight 1, 2, 4, 7, 1 respectively. I have also experimented with skipping the linear interpolation and/or the lookup table.

The current solution takes about 45 seconds on my laptop to calculate a billion entries. It has a distribution error of <5e-3 using a lookup table of 130 bytes.

It is possible to get higher precision, using a lookup table of 1 kilobytes the error can go down to 5e-4 which is similar to what I got with the math.Log2 solution

It is also possible to speed up the implementation by using a larger lookup table and foregoing the interpolation, at a lookup table of 260 kilobytes it also gets a relative error of 5e-4 but can instead do the billion entries in about 30 seconds.

Skipping the lookup table and doing straight interpolation is the fastest since but gives a significant errors (~15%).

There are also some tests added to verify that the distribution is unchanged to prevent accidental changes. It verifies that the exact distribution of the first 10 million entries is unchanged that the exact first 20 values are unchanged. It takes about half a second to run on my machine.

The lookup table is calculated at package initialization, this could potentially give different answers in case the lookup table is calculated on a platform which produces a different value when rounding log2 to 20 bits of precision. To guard against this I have added a checksum for verifying the lookup table at creation of a ShardedBlobAccess Another more stable solution would be to calculate it outside of the process and have Bazel inject the lookup table.

Since changing the constants gives small perturbations to the distribution we should probably be pretty set on the parameters we want to use for interpolation.

EdSchouten:
I think the math we do can be simplified a bit further. What if we created a LUT of $-1/\log(x)$, using an algorithm like the one below?
```go
        // Create a lookup table for f(x) = -1/log(x) between 0<=x<=1,
        // normalized to the range of uint16, which can be used to
        // perform linear interpolation.
        const inBits = 6
        const outBits = 16
        const maxOut = math.MaxUint16
        var lut [1<<inBits + 1]uint16

        // Scale the lookup table so that the final entry to be used as
        // a base (i.e. the penultimate entry of the lookup table) is
        // set to 2^16-1.
        scale := maxOut * math.Log(float64(1<<inBits-1)/float64(1<<inBits))
        for i := 0; i < 1<<inBits; i++ {
                lut[i] = uint16(scale / math.Log(float64(i)/float64(1<<inBits)))
        }

        // -1/log(x) converges to infinity at x=1. Work around this by
        // ensuring the delta between the two final entries is 2^16-1.
        lut[1<<inBits] = lut[1<<inBits-1] + maxOut
```
That will yield the following LUT:
```go
var lut = [...]uint16{
        0x0000, 0x00f8, 0x0129, 0x0151, 0x0174, 0x0194, 0x01b4, 0x01d2,
        0x01f0, 0x020e, 0x022b, 0x024a, 0x0268, 0x0287, 0x02a7, 0x02c7,
        0x02e8, 0x030a, 0x032d, 0x0351, 0x0377, 0x039e, 0x03c6, 0x03f0,
        0x041c, 0x0449, 0x0479, 0x04ab, 0x04e0, 0x0517, 0x0552, 0x058f,
        0x05d0, 0x0616, 0x065f, 0x06ae, 0x0701, 0x075b, 0x07bb, 0x0823,
        0x0893, 0x090d, 0x0992, 0x0a23, 0x0ac2, 0x0b72, 0x0c35, 0x0d0e,
        0x0e03, 0x0f18, 0x1054, 0x11c1, 0x136a, 0x1560, 0x17ba, 0x1a9a,
        0x1e31, 0x22ce, 0x28f4, 0x318f, 0x3e77, 0x53f9, 0x7efb, 0xffff, // <-- Note how this is 2^16-1.
        0xfffe, // <-- And this is 2^16-1 more than the entry before.
}
```
We can then compute the following:
```go
hash := uint32(....)

// Destructure the hash into top/bottom bits.
top := hash >> trailingBits
bottom := hash & (1<<trailingBits - 1)

// Obtain the base and the slope from the lookup table.
base := lut[top]
slope := lut[top+1] - base

// Compute 1/log(x) normalized to the range of uint32.
oneDividedByLog := uint64(base)<<trailingBits + uint64(bottom)*uint64(slope)
oneDividedByLogTruncated := oneDividedByLog >>(outBits-inBits+1)
```
Because weights are also 32-bit integers, we can then compute the score (without any risk of overflows) as follows:
```go
score := uint64(weight) * oneDividedByLogTruncated
```

Would it make sense to use an algorithm like this instead?

EdSchouten:
My concern is that this new algorithm may be (significantly) slower if the number of shards is high. Could we please offer the existing algorithm as a secondary implementation?

pkg/blobstore/sharding/rendezvous_shard_selector.go
```diff
@@ -0,0 +1,174 @@
...
+	"encoding/binary"
+	"fmt"
+	"hash/fnv"
+	"math"
+	"math/bits"
+	"sort"
+)
+
+type rendezvousShard struct {
+	key string
```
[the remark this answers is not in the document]
meroton-benjamin: Ack

pkg/blobstore/sharding/rendezvous_shard_selector.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	h.Write(buf)
+	return h.Sum64()
+}
+
+func NewRendezvousShardSelector(shards []Shard) (*rendezvousShardSelector, error) {
+	// verify integrity of the look up table
+	if lut := hashLut(); lut != 0xb61c4caeda464106 {
+		return nil, fmt.Errorf("Can not shard correctly, log lookup table is broken on this platform got 0x%x expected 0x%x", lut, uint64(0xb61c4caeda464106))
+	}
```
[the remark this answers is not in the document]
meroton-benjamin: Ack, I'll do that as we figure out which precision we're aiming for.

meroton-benjamin:
I have an implementation of your suggested modification. It is effectively almost the same implementation, see below for difference

 * F suffix = estimate the fraction -1/log(x)
 * L suffix = estimate Log2(x)
 * FP{F,L} = number of bits used for fixed point representation of fractional values
 * LS{F,L} = number of bits used for lookup table
 * IP{F,L} = number of bits used for interpolation

Score function:
```golang
func scoreF(x uint64, weight uint32) uint64 {
	return uint64(weight)*LookupF(x)
}
func scoreL(x uint64, weight uint32) uint64 {
	neglogx := (uint64(64)<<FPL) - LookupL(x)
	return (uint64(weight) << 32) / neglogx
}
```

lut init:
```golang
func init() {
	// lookup table for fraction -1/log(x) where x in [0,1]
	lutF[0] = 0
	for i := 1; i < (1<<LSF); i++ {
		target := float64(i) / (1<<LSF)
		val := uint32(math.Round(-1.0/math.Log(target)*math.Pow(2.0, float64(FPF))))
		lutF[i] = val
	}
	lutF[1<<LSF] = lutF[1<<LSF-1]+^uint32(0)
	// lookup table for fraction log2(x) where x in [1,2]
	for i := 0; i < (1<<LSL); i++ {
		target := 1.0+float64(i)/(1<<LSL)
		val := uint32(math.Round(math.Log2(target)*math.Pow(2.0, float64(FPL))))
		lutL[i] = val
	}
	lutL[1<<LSL] = 1<<FPL
}
```

Lookup implementation:
```golang
func LookupF(x uint64) uint64 {
	index := x>>(64-LSF)
	interp := (x >> (64 - IPF - LSF)) & ((1<<IPF)-1)
	current := uint64(lutF[index])
	next := uint64(lutF[index+1])
	delta := next-current
	return (current+(delta*interp)>>IPF)
}

func LookupL(x uint64) uint64 {
	msb := bits.Len64(x)-1
	var bitfield uint64
	if msb >= FPL {
		bitfield = (x >> (msb-FPL)) & ((1<<(FPL))-1)
	} else {
		bitfield = (x << (FPL-msb)) & ((1<<(FPL))-1)
	}
	index := bitfield>>(FPL-LSL)
	interp := bitfield&((1<<(FPL-LSL)-1))
	current := uint64(lutL[index])
	next := uint64(lutL[index+1])
	delta := next-current
	return (uint64(msb)<<FPL) | (current+(delta*interp)>>(FPL-LSL))
}
```

There are some quantitative differences when it comes down to the size of the lookup table required for sufficient precision and speed of implementations.  The F implementation is slightly faster requiring ~5ns per entry per server, the L implementation requires ~9ns per entry per server. This is coincidentally about the same difference as when the L implementation with only a lookup table (no of values in the table)

The L implementation requires a smaller lut for sufficient precision, se comparision here using a lookup table of fixed point uint32 values for both with linear interpolation between both values.

```
  Size of lookup table (in bits)
 ---------- ------------------ ------------------ 
  Existing   F-implementation   L-implementation  
  1e-2       7                  3                 
  1e-3       9                  5                 
  1e-4       12                 8                 
  ```

**Regarding performance**

The implementation does indeed get slower as the number of shards grow large (results from running GetShard in a loop)

```
 Cost of a GetShard call
  Shard Count   Shard Permuter   F-implementation   L-implementation  
 ------------- ---------------- ------------------ ------------------ 
            1   20ns             5ns                9ns               
            2   20ns             9ns                18ns              
            3   20ns             14ns               26ns              
            4   20ns             18ns               35ns              
            5   20ns             23ns               44ns              
```

While I don't have the data to support this, my gut feeling is that the amount of compute spent figuring out which blob belongs to which shard is unimportant, if you have an FMB for 10k blobs destined for n shards you will have to send n FMB requests and perform N*10k GetShard calls. The FMB requests themselves will likely be far more costly than the extra cost of 50-90 microseconds per shard. If performance is the only reason to support both sharding methods I don't think it's worth the effort. We could however want to support both sharding methods for other reasons.

**Regarding reclaiming bits of precision back from the weight**

We could reduce the maximum number of bits we allow the weight to be, however allowing all 32 bits to still allows us ot have very high precision, even if we improve the balance of addresses between the  shards, each individual blob will also have a µ and a σ which will dominate the equation for some value. Looking at my clusters, todays algorithm which gives a perfect distribution of address space still has about a 10% variance between shards of the same size due to individual blob size variance. Adding an addressing error of 1e-5 or 1e-6 would simply be lost in the noise.

**Comment by EdSchouten** [comment]
> The L implementation requires a smaller lut for sufficient precision

Ah, yeah. That's what the core difference is between the two.

- The nice thing about log2(x) where 1<=x<=2 is that it's very smooth, meaning the LUT of LookupL will be fairly accurate. This differs from -1/log2(x) where 0<=x<=1, which is very steep close to both 0 and 1.
- LookupL can also reuse the same LUT for each powers of two, whereas for the F implementation we only have a single LUT for the entire spectrum, meaning that we need log2(64) = 6 more bits to get the same granularity.

So let's stick with algorithm L, but let's see if we can make some minor cleanups to it. I'll post some more comments in bit. Could you also please move it out of draft when you think it's appropriate?

File: pkg/blobstore/sharding/rendezvous_shard_selector.go [hunk 5]
```diff
@@ -0,0 +1,174 @@
...
+	// Replace weight with fixed point representation of weight. We're not using
+	// floating point math so we relative size of the weight to be as big as
+	// possible compared to the log. Since weight is 32 bit it is safe to shift
+	// it by an additional 32 bits. The maximum number of bits of the log is
+	// 5+LOG_BITS.
+	weightFixed := uint64(weight)<<(64-MAX_WEIGHT_BITS)
+	return weightFixed/logFixed
+}
+
+const LOG_BITS = 15
```
EdSchouten, line 97: Maybe it's me, but I would find this algorithm slightly more logical if `LOG_BITS` was set to 16, so that the uint16's effectively correspond to the top part of the IEEE floating point significand of these values. This would cause the final entry in the table to be set to `0x10000`. But like the function I provided, setting it to 0 would likely also work fine, assuming the delta is computed using 16 bit math.

pkg/blobstore/sharding/rendezvous_shard_selector.go
```diff
@@ -0,0 +1,174 @@
...
+	x *= 0x94d049bb133111eb
+	x ^= x >> 31
+	return x
+}
+
+func (s *rendezvousShardSelector) GetShard(hash uint64) int {
+	var best uint64
+	var bestIndex int
+	for _, shard := range s.shards {
+		mixed := splitmix64(shard.hash^hash)
```
EdSchouten, line 166: I wonder whether simply XORing these two values together is good enough at mixing them. This means that every bit of the digest's hash is only affected by exactly one bit of the server hash.

Is this a common way of combining hash values?

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -285,13 +285,21 @@ message ShardingBlobAccessConfiguration {
   // the data.
   uint64 hash_initialization = 1;
```
EdSchouten, line 286: Do we think this field is still needed? Now that shards are identified with a string, the potential concerns arising from layering multiple copies of ShardingBlobAccess don't really apply anymore, assuming each level uses a distinct set of keys (which it likely will).

pkg/blobstore/sharding/sharding_blob_access.go
```diff
@@ -43,36 +45,28 @@ func (ba *shardingBlobAccess) getBackendIndexByDigest(blobDigest digest.Digest)
 }
 
 func (ba *shardingBlobAccess) getBackendIndexByHash(h uint64) int {
```
EdSchouten, line 47: This function no longer really carries its weight. Probably best to remove it and inline the calls to GetShard()?

pkg/blobstore/sharding/rendezvous_shard_selector.go
```diff
@@ -0,0 +1,174 @@
...
+	if len(shards) == 0 {
+		return nil, fmt.Errorf("RendezvousShardSelector must have shards to be defined")
+	}
+	internalShards := make([]rendezvousShard, 0, len(shards))
+	keyMap := make(map[uint64]string, len(shards))
+	for index, shard := range shards {
+		hash := hashServer(shard.Key)
+		if collision, exists := keyMap[hash]; exists {
+			return nil, fmt.Errorf("Hash collision between shards: %s and %s", shard.Key, collision)
+		}
```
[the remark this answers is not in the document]
meroton-benjamin: In case of a hash collision between two different keys the latter of the backends would become unused, and since the iteration order of the map is not well defined which backend is the latter would vary from instance to instance.

Since we're using 64-bits from sha256 hash collisions are unlikely, we could remove the check if you prefer but it's only checked at creation of the shard selector anyway.

pkg/blobstore/sharding/rendezvous_shard_selector.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+type rendezvousShardSelector struct {
+	shards []rendezvousShard
+}
+
+func hashServer(key string) uint64 {
+	h := fnv.New64a()
+	h.Write([]byte(key))
+	return h.Sum64()
```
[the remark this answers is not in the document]
meroton-benjamin: ack

pkg/blobstore/sharding/rendezvous_shard_selector.go
```diff
@@ -0,0 +1,174 @@
...
+	// math. Calculates the logarithm as the sum of three pieces:
+	// 1. The integer value, which is calculated by counting number of bits.
+	// 2. A value calculated by a lookup table of LUT_ENTRY_BITS
+	// 3. The linearly interpolated value between the lookup table and the next value.
+	//
+	// log2(x) = N+log2(x/2^N) we pick the N integer part of the logaritm which
+	// we can calculate exactly by counting the number of bits in the number.
+	// x/2^N will then be a number between 1 and 2 for which we can use a lookup
+	// table to get precomputed values.
+	msb := bits.Len64(x)-1
```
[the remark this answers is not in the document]
meroton-benjamin: ack

pkg/blobstore/sharding/rendezvous_shard_selector.go
```diff
@@ -0,0 +1,174 @@
...
+	if msb >= LOG_BITS {
+		bitfield = (x >> (msb-LOG_BITS)) & ((1<<LOG_BITS)-1)
+	} else {
+		bitfield = (x << (LOG_BITS-msb)) & ((1<<LOG_BITS)-1)
+	}
+	// Use the first LUT_ENTRY_BITS to look up the index, use the remaining
+	// fraction to partially interpolate to the next entry.
+	INTERP_BITS := LOG_BITS-LUT_ENTRY_BITS
+	index  := bitfield>>INTERP_BITS
+	interp := bitfield&((1<<INTERP_BITS)-1)
```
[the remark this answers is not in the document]
meroton-benjamin: That works and removing the branch made a significant performance difference on my machine (down to 5ns per server).

pkg/blobstore/sharding/rendezvous_shard_selector.go
meroton-benjamin: I've set this to a hardcoded 16 and used uint16s, I've also removed the constant in favor of the literal 16 making the code a bit easier on the eyes.

pkg/proto/configuration/blobstore/blobstore.proto
meroton-benjamin: I have removed hash_initialization and the fnv algorithm in the latest version in favor of using the first 64 bits of the digest hash.

pkg/blobstore/sharding/shard_selector.go
```diff
@@ -0,0 +1,20 @@
...
+	GetShard(hash uint64) int
+}
+
+// A description of shard. The shard selector will resolve to the same shard
+// independent of the order of shards, but the returned index will correspond
+// to the index sent to the ShardSelectors constructor.
+type Shard struct {
+	Key string
+	Weight uint32
+}
```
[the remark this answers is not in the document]
meroton-benjamin: The Shard type is used by the ShardSelector as well as the ShardingBlobAccess but I modified the ShardingBlobAccess to use a ShardBackend type instead.

pkg/blobstore/sharding/sharding_blob_access.go
meroton-benjamin: ack

pkg/blobstore/sharding/rendezvous_shard_selector.go
meroton-benjamin: xoring is a common way of preserving entropy, as long as the sources of randomness is uncorrelated (which the hash of the digest and the hash of the server key should be) the operation is guaranteed to only increase entropy. The mixing to the keyspace comes from splitmix64 which has very good randomness properties at a super slim runtime cost.

EdSchouten (approved):

EdSchouten:
Well done, Benjamin!

PR #246, opened 2025-05-06
Add a cassandra-backed storage option

shs96c:


pkg/blobstore/cassandra/README.md
```diff
@@ -0,0 +1,104 @@
...
+## Required setup
+
+`bb-storage` should not be configured to use the admin user by default in
+production. Instead, you should be using a regular user which cannot create
+tables or keyspaces.
+
+The tables that are required can be set up using command similar to the
+below in a `cqlsh` session:
+
+```shell
```
EdSchouten, line 12: s/shell/sql/ or something?

pkg/blobstore/cassandra/cassandra_blob_access.go
```diff
@@ -0,0 +1,673 @@
...
+	"github.com/buildbarn/bb-storage/pkg/capabilities"
+	bbdigest "github.com/buildbarn/bb-storage/pkg/digest"
+	"github.com/buildbarn/bb-storage/pkg/util"
+
+	remoteexecution "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
+	"github.com/gocql/gocql"
+	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
EdSchouten, line 30: For consistency with the rest of the code base, please group imports by second level domain. So first comes github.com, then an empty line, and then golang.org.

pkg/blobstore/cassandra/cassandra_blob_access.go
```diff
@@ -0,0 +1,673 @@
...
+	lastAccessUpdateWorkerCount = 50
+)
+
+var (
+	cassandraBlobStorePrometheusMetrics sync.Once
+
+	segmentWriteHist = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: "buildbarn",
+		Subsystem: "storage",
+		Name:      "cassandra_segment_write_duration_in_seconds",
```
EdSchouten, line 50: Please following the Prometheus best practices on naming: https://prometheus.io/docs/practices/naming/

In particular, remove the `_in_` part of the metric's name.

pkg/blobstore/cassandra/cassandra_blob_access.go
```diff
@@ -0,0 +1,673 @@
...
+	if err != nil {
+		log.Printf("Unable to create session: %v", err)
+		return blobstore.NewErrorBlobAccess(err)
+	}
+
+	// Create the worker pool for updating the last access times
+	jobs := make(chan job, lastAccessUpdateWorkerCount)
+	for i := 0; i < lastAccessUpdateWorkerCount; i++ {
+		go lastUpdateWorker(jobs)
+	}
```
EdSchouten, line 151: As discussed, please move this into a separate function, as it prevents cassandraBlobAccess from being unit tested properly. Also remove the `log.Printf()` calls, as there is no need to repeat whatever is specified in config. Because once we start to tolerate this, the question becomes: "Why do we print the value of this one configuration option, but not of this other one?" Also get rid of the calls to NewErrorBlobAccess. Any misconfigurations should lead to a fatal startup error.

pkg/blobstore/cassandra/cassandra_blob_access.go
```diff
@@ -0,0 +1,673 @@
...
+		lastAccessUpdateInterval: lastAccessUpdateInterval,
+		lastAccessUpdate:         jobs,
+		universalInstanceName:    universalInstanceName,
+		tables:                   newTables(tablePrefix, session),
+	}
+}
+
+func (a *cassandraBlobAccess) GetCapabilities(ctx context.Context, instanceName bbdigest.InstanceName) (*remoteexecution.ServerCapabilities, error) {
+	return a.capabilitiesProvider.GetCapabilities(ctx, instanceName)
+}
```
EdSchouten, line 169: Probably nicer to embed CapabilitiesProvider directly into cassandraBlobAccess, so we don't need this wrapper.

pkg/blobstore/cassandra/cassandra_blob_access.go
```diff
@@ -0,0 +1,673 @@
...
+		Namespace: "buildbarn",
+		Subsystem: "storage",
+		Name:      "cassandra_last_access_failure_count",
+		Help:      "Count of the number of times the `last_access` field could not be updated",
+	})
+)
+
+type job struct {
+	task func()
+}
```
EdSchouten, line 81: Do we expect that `job` ever gets additional members? If not, probably remove it and use the callback type directly?

pkg/blobstore/cassandra/cassandra_blob_access.go
```diff
@@ -0,0 +1,673 @@
...
+		Subsystem: "storage",
+		Name:      "cassandra_segment_read_duration_in_seconds",
+		Help:      "Amount of time in seconds it takes to read each segment from Cassandra",
+		Buckets:   util.DecimalExponentialBuckets(-3, 6, 2),
+	})
+
+	totalWritesHist = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: "buildbarn",
+		Subsystem: "storage",
+		Name:      "cassandra_total_write_duration_in_seconds",
```
EdSchouten, line 66: Instead of using "total", please specify to what it is "total". `cassandra_blob_write_duration_seconds`?

Again, to me it's also not clear why this metric is needed. We're essentially measuring the duration of the entire call to `Put()`. This is something that MetricsBlobAccess can already do. MetricsBlobAccess already breaks down the metrics by backend type. If that's not sufficient, we should improve the way we set up MetricsBlobAccess. Not add additional metrics on the Cassandra side to compensate.

pkg/blobstore/cassandra/cassandra_blob_access.go
```diff
@@ -0,0 +1,673 @@
...
+func (a *cassandraBlobAccess) GetCapabilities(ctx context.Context, instanceName bbdigest.InstanceName) (*remoteexecution.ServerCapabilities, error) {
+	return a.capabilitiesProvider.GetCapabilities(ctx, instanceName)
+}
+
+func logErrorIfNotCancelledContext(err error, msg string, args ...interface{}) {
+	if err != nil && !errors.Is(err, context.Canceled) {
+		initialMessage := fmt.Sprintf(msg, args...)
+		log.Printf("%s: %v", initialMessage, err)
+	}
+}
```
EdSchouten, line 176: Please remove logging like this. If we just bubble up errors properly, then all of the existing tracing middleware should provide sufficient insight. If not, we should look into setting up https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/github.com/gocql/gocql.

pkg/blobstore/cassandra/cassandra_blob_access.go
```diff
@@ -0,0 +1,673 @@
...
+		Name:      "cassandra_segment_read_duration_in_seconds",
+		Help:      "Amount of time in seconds it takes to read each segment from Cassandra",
+		Buckets:   util.DecimalExponentialBuckets(-3, 6, 2),
+	})
+
+	totalWritesHist = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: "buildbarn",
+		Subsystem: "storage",
+		Name:      "cassandra_total_write_duration_in_seconds",
+		Help:      "Total amount of time in seconds it takes to write an entire bytestream to Cassandra",
```
EdSchouten, line 67: I would not consider using the term "bytestream" here. REv2 only uses the Bytestream API for interfacing with the CAS, even though this backend can also be used for the AC and such.

pkg/blobstore/cassandra/cassandra_blob_access.go
```diff
@@ -0,0 +1,673 @@
...
+func (a *cassandraBlobAccess) Get(ctx context.Context, digest bbdigest.Digest) buffer.Buffer {
+	metadata, err := a.tables.metadata.read(ctx, digest)
+	if err != nil {
+		if errors.Is(err, gocql.ErrNotFound) {
+			return buffer.NewBufferFromError(status.Errorf(codes.NotFound, "No metadata found for digest %s", digest.String()))
+		}
+		return buffer.NewBufferFromError(err)
+	}
+
+	go a.tables.metadata.updateLastAccessTime(context.Background(), digest, metadata.instanceName, metadata.lastAccess, a.lastAccessUpdateInterval, a.lastAccessUpdate)
```
EdSchouten, line 187: Again, even though having no concurrency limits in place might at this point never have been an issue for you, I'd like to see this addressed before this change lands.

pkg/blobstore/cassandra/cassandra_blob_access.go
```diff
@@ -0,0 +1,673 @@
...
+	var mu sync.Mutex
+
+	g, errCtx := errgroup.WithContext(ctx)
+	// Try not to be a terrible, terrible citizen
+	g.SetLimit(simultaneousWorkers)
+	for _, digest := range digests.Items() {
+		g.Go(func() error {
+			return a.isMissing(errCtx, digest, &builder, &mu)
+		})
+	}
```
EdSchouten, line 317: A downside of using .SetLimit() in combination with a loop like this is that there is no cancelation at all during loop iteration. If a single `.isMissing()` fails, this code will fail to break immediately, meaning all remaining entries MUST be checked, even if we know they will fail.

Please consider getting rid of the `.SetLimit()` call and using a separate semaphore, like we do in most other places of the Buildbarn codebase.

pkg/blobstore/cassandra/cassandra_blob_access.go
```diff
@@ -0,0 +1,673 @@
...
+
+func lastUpdateWorker(jobs chan job) {
+	for j := range jobs {
+		j.task()
+	}
+}
+
+func (a *cassandraBlobAccess) getInstanceName(digest bbdigest.Digest) string {
+	instanceName := digest.GetInstanceName().String()
+	if instanceName == a.universalInstanceName {
```
EdSchouten, line 355: Please remove support for "universal instance names". I don't think there's any need to do this as part of this backend specifically. Can't it be implemented as a separate decorator for BlobAccess?

pkg/blobstore/cassandra/format_test.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/gocql/gocql"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_formatQuery(t *testing.T) {
+	session := gocql.Session{}
+	longBlob := bytes.Repeat([]byte("This is a very long string!"), 100)
+	query := session.Query("INSERT INTO some_table (id, first_name, long_blob) VALUES (?, ?, ?)", 1234, "Jon Snow", longBlob)
+	expectedOutput := `[query ANY "INSERT INTO some_table (id, first_name, long_blob) VALUES (?, ?, ?)" ` +
+		`[1234 Jon Snow [84 104 105 115 32 105 115 32 97 32 118 101 114 121 32 108 111 110 103 32 115 116 114 105 110 103 33...]]`
```
EdSchouten, line 16: Does this mean that if failures occur, this storage backend will potentially dump megabytes of decimal numbers to stderr?

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -935,3 +941,52 @@ message DeadlineEnforcingBlobAccess {
...
+  // The hosts to use, assuming no discovery is being used.
+  repeated string hosts = 1;
+  // Override the port to connect on from the default.
+  int32 port = 2;
+
+  buildbarn.configuration.tls.ClientConfiguration tls = 3;
+
+  // The Cassandra protocol version. This can normally be omitted from
+  // your config.
+  int32 protocolVersion = 4;
```
EdSchouten, line 955: protocol_version, preferred_dc.

pkg/blobstore/cassandra/cassandra_blob_access.go
shs96c: Removed. I'll do a follow up PR to set the instance name to a default value if necessary.

shs96c (commented):
I've landed changes for the simple things. Still need to update to handle the go routines piling up, and I'll get to that ASAP.

pkg/blobstore/cassandra/cassandra_blob_access.go
shs96c: How's that?

PR #247, opened 2025-05-19
Get user claims from ID token

isakstenstrom:
## What?

This makes it possible to configure a OIDC authenticator to use the ID token for user info claims instead of the user info endpoint.

## Why?

Some OIDC providers (such as Entra ID) only provide some user info claims in the ID token and not in the user info endpoint.

## How?

This adds a `oneof` configuration where you can either use a `userInfoEndpointUrl` to fetch user info claims from the endpoint (same as before), or 
```
useIdTokenClaims: {
  defaultTokenValidityDuration: <some duration>,
}
```
to fetch user info claims from the ID token.

If the ID token does not contain a `exp` field, we can't know how long the token will be valid. The strategy when having a access token without an expiry does not work, since we cannot check with the server if the ID token is still valid. Therefore, a default validity duration is configurable.


pkg/http/oidc_authenticator.go
```diff
@@ -136,18 +144,65 @@ func (a *oidcAuthenticator) getClaimsAndSetCookie(ctx context.Context, token *oa
...
+		return nil, time.Time{}, errors.New("Failed to unmarshal ID token claims")
 	}
 
+	if claimsMap, ok := claims.(map[string]interface{}); ok {
+		if expStr, ok := claimsMap["exp"].(string); ok {
+			if exp, err := strconv.ParseInt(expStr, 10, 64); err == nil {
+				return claims, time.Unix(exp, 0), nil
+			}
+		}
+	}
```
EdSchouten, line 189: I don't think we should extract `exp` here. The reason being that this will cause the expiration time of the authenticated session to be tied to that value. Looking at the OIDC spec:

https://openid.net/specs/openid-connect-core-1_0.html#IDToken

> NOTE: The ID Token expiration time is unrelated the lifetime of the authenticated session between the RP and the OP.

I think that the code as it is (using the expiration returned by the token request) should do the right thing?

pkg/http/oidc_authenticator.go
```diff
@@ -136,18 +144,65 @@ func (a *oidcAuthenticator) getClaimsAndSetCookie(ctx context.Context, token *oa
...
+
+	claimsDecoded, err := base64.RawURLEncoding.DecodeString(parts[1])
+	if err != nil {
+		return nil, time.Time{}, errors.New("Failed to decode ID token claims")
+	}
+
+	var claims interface{}
+	if err := json.Unmarshal(claimsDecoded, &claims); err != nil {
+		return nil, time.Time{}, errors.New("Failed to unmarshal ID token claims")
 	}
```
EdSchouten, line 181: This does not perform any validation of `id_token`. Do we think this is acceptable? Any way we can leverage pkg/jwt?

pkg/http/oidc_authenticator.go
```diff
@@ -136,18 +144,65 @@ func (a *oidcAuthenticator) getClaimsAndSetCookie(ctx context.Context, token *oa
...
+
+	return claims, a.clock.Now().Add(a.defaultIDTokenValidityDuration), nil
+}
+
+func (a *oidcAuthenticator) getClaimsAndTokenExpiry(ctx context.Context, token *oauth2.Token) (interface{}, time.Time, error) {
+	if a.useIDTokenClaims {
+		return a.getClaimsAndExpiryFromIDToken(token)
+	}
+	return a.getClaimsAndExpiryFromEndpoint(ctx, token)
+}
```
EdSchouten, line 199: I'd rather see that we decompose `oidcAuthenticator` to facilitate this. Could you please add something like this?

```go
type OIDCClaimsFetcher interface {
        FetchClaimsForToken(ctx context.Context, token *oauth2.Token) (any, error)
}

type userInfoOIDCClaimsFetcher struct {
        oauth2Config *oauth2.Config
 	userInfoURL  string
}

func (cf *userInfoOIDCClaimsFetcher) FetchClaimsForToken(ctx context.Context, token *oauth2.Token) (any, error){
        claimsRequest, err := http.NewRequestWithContext(ctx, http.MethodGet, cf.userInfoURL, nil)
        ...
}

type idTokenOIDCClaimsFetcher struct {}

func (cf *idTokenOIDCClaimsFetcher) FetchClaimsForToken(ctx context.Context, token *oauth2.Token) (any, error){
        idToken, ok := token.Extra("id_token").(string)
        ...
}
```

Optionally, create a pkg/http/oidc and move it in there. Then we can remove the "OIDC" part from the name of both OIDCAuthenticator and the types proposed above.

pkg/http/oidc_authenticator.go
```diff
@@ -136,18 +144,65 @@ func (a *oidcAuthenticator) getClaimsAndSetCookie(ctx context.Context, token *oa
...
+		return nil, time.Time{}, util.StatusWrap(err, "Failed to unmarshal claims")
+	}
+
+	return claims, token.Expiry, nil
+}
+
+func (a *oidcAuthenticator) getClaimsAndExpiryFromIDToken(token *oauth2.Token) (interface{}, time.Time, error) {
+	idToken, ok := token.Extra("id_token").(string)
+	if !ok || idToken == "" {
+		return nil, time.Time{}, errors.New("No ID token present in OIDC server response")
```
EdSchouten, line 165: Please use `status.Error(codes.*, "...")` for error messages. The reason being that consumers of Authenticator depend on gRPC status codes to distinguish between transient/non-transient errors.

EdSchouten (commented):
Thanks for your PR, Isak. Looking forward to see this land!

pkg/http/oidc_authenticator.go
isakstenstrom: Good point! I have removed all expiration handling from the PR, so now it is back to how the expiry was before, including the exponential increase.

pkg/http/oidc_authenticator.go
isakstenstrom: Yes, that is a much better structure, thanks!

pkg/http/oidc_authenticator.go
isakstenstrom: Done! I didn't find a error code that felt like it fit perfectly. The best i found was `codes.NotFound`. Let me know if you think it should be something else.

pkg/http/oidc_authenticator.go
isakstenstrom: I discussed this with Benjamin Ingberg and we think that this is fine. We already trust that the access token and refresh token we have gotten are correct, and since we get the ID token from the same API call, we should be able to trust it as well. But if you think that we should validate the ID token, I can try to implement it.

pkg/http/oidc_authenticator_test.go
```diff
@@ -55,26 +55,47 @@ func TestOIDCAuthenticator(t *testing.T) {
...
+	userInfoAuthenticator, err := bb_http.NewOIDCAuthenticator(
+		oauth2Config,
+		userInfoClaimsFetcher,
+		jmespath.MustCompile("{\"public\": @}"),
+		&http.Client{Transport: roundTripper},
+		randomNumberGenerator,
+		"CookieName",
+		cookieAEAD,
+		clock,
+	)
```
EdSchouten, line 84: This is missing a `require.NoError(t, err)`.

pkg/http/oidc_authenticator_test.go
```diff
@@ -55,26 +55,47 @@ func TestOIDCAuthenticator(t *testing.T) {
...
+		oauth2Config,
+		idTokenClaimsFetcher,
 		jmespath.MustCompile("{\"public\": @}"),
 		&http.Client{Transport: roundTripper},
 		randomNumberGenerator,
 		"CookieName",
 		cookieAEAD,
-		clock)
+		clock,
+	)
```
EdSchouten, line 97: Maybe readability of the tests would increase if you added additional `t.Run()` blocks here?

```go
func TestOIDCAuthenticator(t *testing.T) {
    t.Run("UserInfo", func(t *testing.T) { /* all of the existing tests go here */ })
    t.Run("IDToken", func(t *testing.T) { /* some tests go here */ })
}
```

Then you can also keep the name `authenticator` instead of renaming it to `userInfoAuthenticator`.

pkg/http/oidc_authenticator.go
EdSchouten: I'd say, use `codes.Internal` here and in other places in this function. Yes, you're correct that the token was "not found". But think about it from the user's perspective. The user is trying to load the page. Why does the page not appear? Because of an issue that is internal to the underlying services. Namely, a bad interaction between the web service and the OAuth2 server.

Returning `codes.NotFound` should only be done if the user actively tries to fetch something that does not exist (in this case, requesting a page that does not exist).

pkg/http/oidc_authenticator.go
EdSchouten: Agreed. Leveraging pkg/jwt might also not be desirable, because it means we'll need to add a `NoopSignatureValidator`. And there are so many ways in which that may be used accidentally. Keeping pkg/jwt "strong" and having a bare bones JWT parser here that skips the header and signature sounds fine to me.

EdSchouten (commented):
Great work, Isak, and thanks for following through on my comments. Some final remarks and this will be good to land.

PR #248, opened 2025-06-04
refactor: use pre-built protoc binary

alexeagle:
Saves a bunch of time in analysis and execution when there is a cache miss on protoc, as well as avoiding a bunch of gcc logspam.


.bazelrc
```diff
@@ -1 +1,6 @@
 run --workspace_status_command="bash tools/workspace-status.sh"
+
+# Don't build protoc from the cc_binary, it's slow and spammy when cache miss
+common --per_file_copt=external/.*protobuf.*@--PROTOBUF_WAS_NOT_SUPPOSED_TO_BE_BUILT
+common --host_per_file_copt=external/.*protobuf.*@--PROTOBUF_WAS_NOT_SUPPOSED_TO_BE_BUILT
```
EdSchouten: So I guess you're adding these lines to ensure that the build fails if protoc does end up getting built from source, right?

If so, please remove this. It ends up making this repository unbuildable on platforms for which toolchains_protoc provides no prebuilt binary.

MODULE.bazel
```diff
@@ -31,6 +32,14 @@ single_version_override(
     patches = ["//:patches/jsonnet/bazel-8.diff"],
 )
 
+protoc = use_extension("@toolchains_protoc//protoc:extensions.bzl", "protoc")
+
+# Match protobuf above, until it ships a pre-built binary toolchain itself
+protoc.toolchain(version = "v29.3")
+use_repo(protoc, "toolchains_protoc_hub")
+
+register_toolchains("@toolchains_protoc_hub//:all")
```
EdSchouten: I don't understand why this needs to be done manually. We don't need to do this for rules_go to get a Go toolchain. Is this something that should have been part of toolchains_protoc's module extension?

.bazelrc
alexeagle: That's right. I think it's desirable to prevent accidental new dependencies on the protoc `cc_binary` target, which can creep in via dependencies. We could add this flag only to the CI configuration for that purpose.

alexeagle: Moved to `.github/workflows/ci.bazelrc` which gets copied to the home directory's bazelrc during bazel install. This way we are sure all CI actions use that flag.

MODULE.bazel
alexeagle: I verified that it's required for the root module to register the toolchain.
In fact in bb-deployments mentioned earlier by @stagnation I checked some recent CI logs and see that we're waiting on protoc compile in the critical path.

This is also what we decided for bazel-lib 2.0, having users explicitly register toolchains was less surprising. @kormide can probably recall more detail on that.

stagnation: Thanks! I created a mirror PR for the deployments: https://github.com/buildbarn/bb-deployments/pull/170
:-)

EdSchouten: > I verified that it's required for the root module to register the toolchain. [...]
> 
> This is also what we decided for bazel-lib 2.0, having users explicitly register toolchains was less surprising. @kormide can probably recall more detail on that.

But how does that scale? The last time I checked, even a somewhat simple project like bb-storage depends on 1015 toolchains belonging to 39 toolchain types. It's clearly not sustainable to have 39 `register_toolchains()` calls in MODULE.bazel. If this ends up becoming the norm, then the MODULE.bazel transition is arguably a failure, because [we wouldn't gain anything over WORKSPACE](https://github.com/bazelbuild/bazel/discussions/22024).

Furthermore, we have other repos that depend on bb-storage, like bb-remote-execution, bb-clientd, etc.. Those now don't need to call `register_toolchains()`, because these do in its current form inherit these toolchains. That's kind of arbitrary, right? If toolchains_protoc doesn't register the toolchains, then it clearly doesn't make sense for bb-storage to be the point of decision of which Protobuf toolchains are used. This can be prevented by adding `dev_dependency=False`.

.github/workflows/ci.bazelrc
```diff
@@ -0,0 +1,2 @@
+common --per_file_copt=external/.*protobuf.*@--PROTOBUF_WAS_NOT_SUPPOSED_TO_BE_BUILT
```
EdSchouten: Now each Buildbarn repo needs to manually add a file like this to get these flags picked up, which I don't think is desirable.

I'd rather see that we don't make any changes along these lines. They aren't strictly necessary, and also prone to breakage. For example, this would break if our build starts to depend on another C/C++ tool that is stored at a location having "protobuf" in its name:

```
external/rocket_boots_calibration_data+/protobuf_test_data_ingester.cc
```

alexeagle: They're not strictly necessary, however I do think we should defend against what happened in bb-deployments where we thought we stopped compiling protoc from source, but it was either never working that way, or regressed. Do you have any alternative suggestion how we can make sure this change has the intended effect? Simply a README that reminds us to manually verify this change occasionally?

EdSchouten: I don't think it's really necessary to either document this, or have additional checks in place. I expect that if we have a line in MODULE.bazel that says that we want to use a prebuilt version of protoc, that it ends up getting used.

MODULE.bazel
alexeagle: I'd note that toolchains_llvm also instructs users to register the toolchain: https://github.com/bazel-contrib/toolchains_llvm/releases/tag/v1.4.0

However I agree that it's not desirable to force users to be explicit about this. I misremembered the conversation with @kormide about bazel-lib, found here: https://github.com/bazel-contrib/bazel-lib/pull/597#issuecomment-1754122447

The answer should be as you say, proven by https://github.com/aspect-build/toolchains_protoc/pull/75

So next I'm debugging why we still see protoc being compiled in this repo without this line...

alexeagle: Fixed: I didn't realize `protobuf` stole the toolchain symbol https://github.com/bazelbuild/rules_proto/blob/main/proto/BUILD#L59 and then register the from-source protoc at that toolchain

https://github.com/protocolbuffers/protobuf/blob/main/MODULE.bazel#L119 -> https://github.com/protocolbuffers/protobuf/blob/main/bazel/private/toolchains/BUILD.bazel#L11 

So the order of `bazel_dep` calls is what matters for getting the one we want.

.github/workflows/ci.bazelrc
alexeagle: Okay, following your lead :)

EdSchouten:
Thanks!

MODULE.bazel
kormide: I like the set of guidelines that matts1 outlined [here](https://github.com/bazelbuild/bazel/discussions/22024#discussioncomment-9137180). I agree that modules defining the toolchain type ought to register a default toolchain. It adds a convenience and doesn't hurt extensibility since anyone can override it in the root module.

PR #255, opened 2025-06-24
WIP feat(blobstore): create resilliant mirrored storage

tyler-french:
We have situations where we need to restart either `storage-a` or `storage-b`, and we can't have downtime.

To support this use case, we should allow for only one of the storage shards to be available.


tyler-french:
> > Additionally, we've seen unexpected storage failures (from memory pressure, nodes being cycled by GCP, etc.) result in complete build system outages rather than graceful degradation.
> 
> Please figure out what can be done to limit the impact of that. 

Thanks, the references are helpful. I think playing around with [`remote_retry_max_delay`](https://bazel.build/reference/command-line-reference#common_options-flag--remote_retry_max_delay) might help here. I'll look around to see what more client-side and infra-side we can do to alleviate these issues.

tyler-french:
> Think of what would happen if frontends can only reach replica A, while workers can only reach replica B: actions would fail due to incomplete input roots, or Bazel would be unable to download output files.

I understand the split-brain counterargument, but wouldn't this only affect the `Put` and FMB path? My intuition is that the `Read` path should be able to fallback, similar with `GetCapabilities`, since we know that `Write` and `FindMissing` will synchronize mirrors.

EdSchouten:
Reads could fall back. But if the second backend returns NotFound, then it should be suppressed.

PR #258, opened 2025-07-10
feat(blobstore/grpcservers): support over-the-wire compression

tyler-french:


pkg/blobstore/configuration/ac_blob_access_creator.go
```diff
@@ -23,6 +23,10 @@ var acCapabilitiesProvider = capabilities.NewStaticProvider(&remoteexecution.Ser
 		},
 		// CachePriorityCapabilities: Priorities not supported.
 		SymlinkAbsolutePathStrategy: remoteexecution.SymlinkAbsolutePathStrategy_ALLOWED,
+		SupportedCompressors: []remoteexecution.Compressor_Value{
```
EdSchouten: Compressors only apply to CAS traffic.

pkg/blobstore/grpcclients/cas_blob_access.go
```diff
@@ -187,6 +187,10 @@ func (ba *casBlobAccess) GetCapabilities(ctx context.Context, instanceName diges
 	return &remoteexecution.ServerCapabilities{
 		CacheCapabilities: &remoteexecution.CacheCapabilities{
 			DigestFunctions: digest.RemoveUnsupportedDigestFunctions(cacheCapabilities.DigestFunctions),
+			SupportedCompressors: []remoteexecution.Compressor_Value{
```
EdSchouten: I'm not really convinced that it's a good idea to announce the supported compressors here. Whereas the digest functions that are supported is something that applies end to end (i.e., Bazel -> bb_clientd -> frontends -> storage), the compressors only apply to a single hop.

I'd say that the compressors supported by the server should only be interpreted by this BlobAccess implementation and not leak out of it. Though it looks like you're not adding support for compression to the ByteStream client, so I don't think anything should change here.

Instead of changing any of this code, you likely just want to add a global config option of which compressors to announce?

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -2,13 +2,14 @@ package grpcservers
...
 	"context"
+	"errors"
 	"io"
 
 	remoteexecution "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
 	"github.com/buildbarn/bb-storage/pkg/blobstore"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
 	"github.com/buildbarn/bb-storage/pkg/digest"
-
+	"github.com/klauspost/compress/zstd"
```
EdSchouten, line 13: Please leave the empty line. In the Buildbarn source tree, all imports are grouped by second level domain.

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -29,109 +30,177 @@ func NewByteStreamServer(blobAccess blobstore.BlobAccess, readChunkSize int) byt
 	}
 }
 
-func (s *byteStreamServer) Read(in *bytestream.ReadRequest, out bytestream.ByteStream_ReadServer) error {
-	if in.ReadLimit != 0 {
+// Read handles streaming a blob from the blobstore to a client, when a user is downloading a file
+// from the CAS.
+func (s *byteStreamServer) Read(req *bytestream.ReadRequest, stream bytestream.ByteStream_ReadServer) error {
```
EdSchouten: Out of curiosity, what's with the in/out -> req/stream rename? Is there some kind of naming collision we need to work around?

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -29,109 +30,177 @@ func NewByteStreamServer(blobAccess blobstore.BlobAccess, readChunkSize int) byt
...
-			return readErr
-		}
-		if writeErr := out.Send(&bytestream.ReadResponse{Data: readBuf}); writeErr != nil {
-			return writeErr
+	var writer io.Writer
+	switch compressor {
+	case remoteexecution.Compressor_IDENTITY:
+		writer = &readStreamWriter{stream: stream}
+	case remoteexecution.Compressor_ZSTD:
+		zstdWriter, err := zstd.NewWriter(&readStreamWriter{stream: stream})
```
EdSchouten: Considering that this is an io.Writer, write this instead:
```go
b := s.blobAccess.Get(stream.Context(), blobDigest)
switch compressor {
case remoteexecution.Compressor_ZSTD:
    err := b.IntoWriter(zstdWriter)
    ....
}
```

The `.ToChunkReader()` is only useful if you're interested in receiving fixed size chunks that can be placed in ByteStream message bodies.

pkg/blobstore/grpcservers/byte_stream_server_test.go
```diff
@@ -274,7 +582,7 @@ func TestByteStreamServer(t *testing.T) {
 			gomock.Any(),
 		).DoAndReturn(func(ctx context.Context, digest digest.Digest, b buffer.Buffer) error {
 			_, err := b.ToByteSlice(100)
-			testutil.RequireEqualStatus(t, status.Error(codes.InvalidArgument, "Client closed stream twice"), err)
+			require.Error(t, err)
```
EdSchouten: What's the reason behind weakening these checks?

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -29,109 +30,177 @@ func NewByteStreamServer(blobAccess blobstore.BlobAccess, readChunkSize int) byt
...
+	default:
+		return status.Errorf(codes.Unimplemented, "This service does not support writing compression type: %s", compressor)
 	}
+
 	if err := s.blobAccess.Put(
 		stream.Context(),
-		digest,
-		buffer.NewCASBufferFromChunkReader(digest, r, buffer.UserProvided)); err != nil {
+		blobDigest,
+		buffer.NewCASBufferFromReader(blobDigest, blobReader, buffer.UserProvided),
```
EdSchouten: By always using a Reader instead of a ChunkReader, doesn't this mean that we now perform an extra copy of all of the data, even in the uncompressed case?

pkg/blobstore/grpcservers/byte_stream_server_test.go
tyler-french: The combined abstraction made it difficult to catch this case, but by separating out the previous implementation of identity, we can keep this more aligned with the previous setup.

pkg/blobstore/grpcservers/byte_stream_server.go
tyler-french: I got a bit confused initially, but I guess `in` and `out` is clear enough, so I will revert this change

pkg/blobstore/grpcservers/byte_stream_server.go
tyler-french: Sounds good :), just a habit

pkg/blobstore/grpcservers/byte_stream_server.go
tyler-french: Yeah, looking at it again I think it's best to retain the original logic exactly as it was, and separate out the compression implementation. That way we can make future adjustments and modifications to the ZSTD without impacting performance of the identity direction.

pkg/blobstore/grpcclients/cas_blob_access.go
tyler-french: Updated, but the plumbing isn't the most ideal. I think keeping CAS to default support compression isn't the worst idea, and it's cleaner. But I guess server owners would probably want to be able to toggle this.

pkg/blobstore/grpcservers/byte_stream_server.go
tyler-french: Updated, and kept identity as a chunker

pkg/proto/configuration/global/global.proto
```diff
@@ -294,6 +295,16 @@ message Configuration {
...
   // constants (e.g., "NOFILE" to limit the number of file descriptors).
   map<string, SetResourceLimitConfiguration> set_resource_limits = 15;
+
+  // List of compression algorithms supported by the Content Addressable
+  // Storage. When not set, defaults to IDENTITY compression only, which
+  // means that no compression is supported.
+  //
+  // Valid values include:
+  // - IDENTITY: No compression
+  // - ZSTD: Zstandard compression
```
EdSchouten: REv2 states that IDENTITY is implied. Maybe it's cleaner to stick to that here as well?

> List of compression algorithms supported by the Content Addressable Storage. Valid values include:
>
> - ZSTD: Zstandard compression
>
> Support for IDENTITY (i.e., no compression) is implied.

That also allows us to get rid of `global.GetSupportedCompressors()` as well.

pkg/proto/configuration/global/global.proto
```diff
@@ -8,6 +8,7 @@ import "google/protobuf/wrappers.proto";
 import "opentelemetry/proto/common/v1/common.proto";
 import "pkg/proto/configuration/grpc/grpc.proto";
 import "pkg/proto/configuration/http/http.proto";
+import "build/bazel/remote/execution/v2/remote_execution.proto";
```
EdSchouten: Please keep imports sorted by pathname.

pkg/blobstore/grpcservers/byte_stream_server_test.go
```diff
@@ -15,13 +16,13 @@ import (
 	"github.com/buildbarn/bb-storage/pkg/testutil"
 	"github.com/stretchr/testify/require"
 
+	"github.com/klauspost/compress/zstd"
+	"go.uber.org/mock/gomock"
```
EdSchouten: Please keep imports grouped by second level domain:

- github.com
- golang.org / google.golang.org
- go.uber.org

pkg/proto/configuration/global/global.proto
```diff
@@ -294,6 +295,16 @@ message Configuration {
   // Keys of this map correspond to the suffixes of the RLIMIT_*
   // constants (e.g., "NOFILE" to limit the number of file descriptors).
   map<string, SetResourceLimitConfiguration> set_resource_limits = 15;
+
+  // List of compression algorithms supported by the Content Addressable
```
EdSchouten: This has no business being part of pkg/global. That package is used by all Buildbarn binaries. Announcing compressors is specific to bb_storage.

cmd/bb_storage/main.go
```diff
@@ -52,12 +52,16 @@ func main() {
...
+				int(configuration.MaximumMessageSizeBytes),
+				supportedCompressors)
+
 			info, authorizedBackend, allAuthorizers, err := newScannableBlobAccess(
 				dependenciesGroup,
 				configuration.ContentAddressableStorage,
-				blobstore_configuration.NewCASBlobAccessCreator(
-					grpcClientFactory,
-					int(configuration.MaximumMessageSizeBytes)),
+				casCreator,
```
EdSchouten: Why can't we get rid of all of these changes, but change this line further below:
```go
cacheCapabilitiesProviders = append(cacheCapabilitiesProviders, info.BlobAccess)
```
To:
```go
cacheCapabilitiesProviders = append(
        cacheCapabilitiesProviders,
        info.BlobAccess,
        capabilities.NewStaticProvider(&remoteexecution.ServerCapabilities{
		CacheCapabilities: &remoteexecution.CacheCapabilities{
 			SupportedCompressors: configuration.SupportedCompressors,
 		},
        }),
)
```
?

tyler-french: Nice, this merged provider is awesome

pkg/blobstore/grpcservers/byte_stream_server_test.go
tyler-french: ack. What about the internal ones? Do those stay grouped? I notice stretcher testify is grouped with bb-storage

We should probably come up with some auto-formatter to do this for us. I've also seen stdlib, external, generated

EdSchouten: Internal ones are grouped together with the other GitHub.com imports.

Yeah, it would be nice if we had some more automation here. Maybe we should set up goimports?

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -132,6 +163,95 @@ func (s *byteStreamServer) Write(stream bytestream.ByteStream_WriteServer) error
...
+}
+
+func (z *zstdReadCloser) Read(p []byte) (n int, err error) {
+	return z.decoder.Read(p)
+}
+
+func (z *zstdReadCloser) Close() error {
+	z.decoder.Close()
+	return z.underlyingReader.Close()
+}
```
EdSchouten: This type is quite similar to `zstdReader` that lives in `reference_expanding_blob_access.go`. Do we feel like unifying this, or do we think ["a little copying is better than a little dependency"](https://go-proverbs.github.io)? I'm fine either way. It's annoying that these two don't live in the same package.

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -37,27 +39,49 @@ func (s *byteStreamServer) Read(in *bytestream.ReadRequest, out bytestream.ByteS
...
+			if writeErr := out.Send(&bytestream.ReadResponse{Data: readBuf}); writeErr != nil {
+				return writeErr
+			}
 		}
-		if writeErr := out.Send(&bytestream.ReadResponse{Data: readBuf}); writeErr != nil {
-			return writeErr
+
+	case remoteexecution.Compressor_ZSTD:
+		b := s.blobAccess.Get(out.Context(), digest)
+		zstdWriter, err := zstd.NewWriter(&readStreamWriter{out: out})
```
EdSchouten: But for reading and writing, should we make sure to call `zstd.WithDecoderConcurrency(1)` / `zstd.WithEncoderConcurrency(1)`? We already have concurrency at the request level. It's probably not a good idea to also let this spawn `GOMAXPROCS` goroutines.

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -132,6 +163,95 @@ func (s *byteStreamServer) Write(stream bytestream.ByteStream_WriteServer) error
...
+	n = copy(p, req.Data)
+	r.pendingData = req.Data[n:]
+	return n, nil
+}
+
+func (r *zstdWriteStreamReader) Close() error {
+	return nil
+}
+
+func (s *byteStreamServer) writeZSTD(stream bytestream.ByteStream_WriteServer, request *bytestream.WriteRequest, digest digest.Digest) error {
```
EdSchouten: `writeZstd()`?

pkg/proto/configuration/bb_storage/bb_storage.proto
```diff
@@ -89,6 +90,15 @@ message ApplicationConfiguration {
   // operation. This is hopefully safe, as operation names are hard to guess,
   // and the forwarded-to scheduler should perform its own authorization.
   buildbarn.configuration.auth.AuthorizerConfiguration execute_authorizer = 16;
+
+  // List of compression algorithms supported by the Content Addressable
+  // Storage. Valid values include:
```
EdSchouten: Sorry for coming back to this, but maybe we should extend this description a bit?

> List of compression algorithms supported by the Content Addressable Storage **to announce as part of the server's cache capabilities**.

It's probably also worth mentioning that this option does not actually control which compression algorithms are implemented. Clients are always permitted to issue requests using Zstandard. It's merely that not announcing it will cause clients to not use it.

pkg/blobstore/grpcservers/byte_stream_server_test.go
```diff
@@ -125,6 +127,91 @@ func TestByteStreamServer(t *testing.T) {
...
+			gomock.Any(),
+			digest.MustNewDigest("", remoteexecution.DigestFunction_SHA256, "a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a9b0c1d2e3f4a5b6c7d8e9f0a1b2", 100000),
+		).Return(buffer.NewValidatedBufferFromByteSlice(originalData))
+
+		req, err := client.Read(ctx, &bytestream.ReadRequest{
+			ResourceName: "compressed-blobs/zstd/a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a9b0c1d2e3f4a5b6c7d8e9f0a1b2/100000",
+		})
+		require.NoError(t, err)
+
+		// Collect all compressed data
```
EdSchouten: Minor nit: please end all sentences in comments with a full stop (period). Thanks!

EdSchouten (commented):
This is looking very good, Tyler. I remain skeptical of how compression was added to REv2, but this is certainly an implementation I can live with.

I've left an additional round of (minor) feedback. Be sure to address those, and I'll merge it in a timely manner.

pkg/blobstore/grpcservers/byte_stream_server.go
tyler-french: In this case, I think it's okay to go either way. I'll extract to a common structure, but please let me know if you don't like the abstraction and I'll undo it.

pkg/blobstore/grpcservers/byte_stream_server.go
tyler-french: Yep, good call. Do you think we should also use the Lowmem options?

pkg/proto/configuration/bb_storage/bb_storage.proto
tyler-french: Let me know how the rewording looks, happy to iterate here for clarity.

pkg/util/zstd_reader.go
```diff
@@ -0,0 +1,32 @@
...
+	decoder, err := zstd.NewReader(underlyingReader, options...)
+	if err != nil {
+		return nil, err
+	}
+	return &ZstdReader{Decoder: decoder, underlyingReader: underlyingReader}, nil
+}
+
+// ZstdReader is a decorator for zstd.Decoder that ensures both the
+// decoder and the underlying stream are closed upon completion.
+type ZstdReader struct {
```
EdSchouten: Considering that `NewZstdReadCloser()` returns an `io.ReadCloser`, I suppose this type can be private, right? (Also make sure to remove references to `ZstdReader` from the documentation of `NewZstdReadCloser`.)

EdSchouten:
One tiny remark, and we're good to go!

EdSchouten (changes requested):
Oh, wait! I just realized that this PR is only adding support for compression to the ByteStream service. Before we can actually start announcing support for compression we must also extend the ContentAddressableStorage.Batch*() methods.

tyler-french:
> Oh, wait! I just realized that this PR is only adding support for compression to the ByteStream service. Before we can actually start announcing support for compression we must also extend the ContentAddressableStorage.Batch*() methods.

@EdSchouten We can do this, but it is worth noting that the REv2 API actually announces this separately: https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto#L2245-L2256

So as long as we're not announcing a batch compressor, then we're not lying to the client with our announcement.

`BatchReadBlobs` only takes a supported compressor in the request, and therefore Identity is supported even if do add a supported batch compressor.

`BatchUpdateBlobs` does support compressed writes, but again, this is only expected if its announced as a batch compressor.

Implementation is pretty straightforward:
1. If `acceptable_compressors` includes ZSTD in the read request, then compress the data with `EncodeAll`
2. If upload is compressed, `DecodeAll`.

We should probably have the server configure this separately, which might mean this makes sense in a separate PR. Happy to go either way

tyler-french:
For comparison, in Buildfarm, from how I understand it:

`BatchReadBlobs` is **always uncompressed**, and `BatchUpdateBlobs` only compresses if the request has `compressor` set to ZSTD.

But this will never happen, since it's not announced as a `SupportedBatchUpdateCompressors`: https://github.com/buildfarm/buildfarm/blob/main/src/main/java/build/buildfarm/instance/server/NodeInstance.java#L1736

EdSchouten:
Aaaah, great. Then let’s merge it!

PR #260, opened 2025-07-11
feat(blockdevice): support file mappings on windows

tomgr:
This allows Windows users to use blocksOnBlockDevice, for example. This follows the same approach as the unix implementation, using the Windows counterparts of mmap.

This passes the same block device tests as unix, and I've also run a bb_storage instance locally (with a config essentially the same as the README) and used it to cache the results of building bb_storage with no problems.


pkg/blockdevice/memory_mapped_block_device_unix.go
```diff
@@ -91,3 +91,26 @@ func (bd *memoryMappedBlockDevice) WriteAt(p []byte, off int64) (int, error) {
...
+		}
+		bd.data = nil
+	}
+
+	if bd.fd != -1 {
+		if err := unix.Close(bd.fd); err != nil {
+			errors = append(errors, util.StatusWrap(err, "Failed to close file descriptor"))
+		}
+		bd.fd = -1
+	}
```
EdSchouten: I'd say, we shouldn't implement a function like this in a restartable manner. Even if `Munmap()` and `Close()` fail, we should assume that the underlying resource has been released and nothing can be done to recover. It's not as if the caller is going to write a for-loop around this to retry. Could you remove the if-statements and field assignments, please?

pkg/blockdevice/new_block_device_from_file_disabled.go
```diff
@@ -1,5 +1,5 @@
-//go:build windows
-// +build windows
+//go:build !darwin && !freebsd && !linux && !windows
```
EdSchouten: I tend not to use 'catch-all' files like these, for the reason that if people want to build Buildbarn on other operating systems they should just figure out whether one of the existing implementations can be used. Now that we have block device support for Windows, can you please remove this file entirely? Thanks!

pkg/blockdevice/new_block_device_from_file_windows.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/buildbarn/bb-storage/pkg/util"
+
+	"golang.org/x/sys/windows"
+)
+
+var (
+	kernel32          = windows.NewLazySystemDLL("kernel32.dll")
+	GetDiskFreeSpaceW = kernel32.NewProc("GetDiskFreeSpaceW")
+)
```
EdSchouten: I know that we lazily load the DLL, but does calling NewProc() against it still make it lazy? If not, what are your thoughts on putting the initialization of these fields inside a `sync.Once` in `NewBlockDeviceFromFile()`? That way we only depend on this DLL if block devices are used.

Or are we convinced that `kernel32.dll` is so fundamental, that an unconditional dependency doesn't harm?

pkg/blockdevice/new_block_device_from_file_windows.go
```diff
@@ -0,0 +1,95 @@
...
+		if err != nil {
+			windows.CloseHandle(handle)
+			return nil, 0, 0, util.StatusWrapf(err, "Failed to convert root path %#v to UTF-16", rootPath)
+		}
+
+		// Get the sector size using GetDiskFreeSpace via syscall since it's not in windows package
+		var bytesPerSector uint32
+		r, _, err := GetDiskFreeSpaceW.Call(
+			uintptr(unsafe.Pointer(rootPathPtr)),
+			0,
```
EdSchouten: To improve readability, could you please add some `/* sectorsPerCluster = */ 0,` comments in front of the fields that you're setting to zero? That way it becomes more evident why we do.

pkg/blockdevice/new_block_device_from_file_windows.go
```diff
@@ -0,0 +1,95 @@
...
+			0)
+		if r == 0 {
+			windows.CloseHandle(handle)
+			return nil, 0, 0, util.StatusWrapf(err, "Failed to get disk sector size for %#v", rootPath)
+		}
+
+		sectorSizeBytes = int(bytesPerSector)
+	} else {
+		// For UNC paths, assume a default sector size
+		sectorSizeBytes = 4096
```
EdSchouten: I know very little about Windows. Why doesn't the logic above work for UNC paths? Probably worth explaining. Is there some other method that could be used to determine the sector size instead? Is it even common for our users to let block devices be backed by UNC paths?

EdSchouten (commented):
This looks great, Thomas. Thanks for working on this.

General style remark for consistency with the rest of the code base: can you please make sure to end all comments with a full stop/period? Thanks!

pkg/blockdevice/new_block_device_from_file_windows.go
tomgr: Thankfully a `LazyDLL`'s `NewProc` method returns a `LazyProc`, so I think it's already sufficiently lazy.

EdSchouten: Great! Then it's fine to leave as is. Though could you make GetDiskFreeSpaceW private?

pkg/blockdevice/new_block_device_from_file_windows.go
tomgr: Good point. I've had another look at that and it seems GetDiskFreeSpaceW will work for UNC paths, with some adjustments. I've written a new function to compute the "root path" that we have to pass in to GetDiskFreeSpaceW and added some unit tests for it.

One thing I wasn't sure about was how to make the utility function private whilst still adding unit tests for it so for now I've exported it. If you've got a better suggestion for this please say -- I've not written much go before so!

EdSchouten: Got it! Yeah, feel free to just keep it public. If it turns out we need this more frequently, then we can move it to pkg/filesystem later.

EdSchouten (approved):
LGTM, assuming any remaining CI/linter warnings are addressed.

pkg/blockdevice/new_block_device_from_file_windows.go
tomgr: Oops sorry.  Done!

tomgr:
Brilliant thanks for the quick reviews!

PR #263, opened 2025-07-25
feat(grpc): allow files to be read by add_metadata_jmespath_expression

tomgr:
The motivation for this change is that I would like to be able k8s bound service account tokens and gcloud service account id tokens into outgoing grpc requests. However, these tokens often have relatively short maximum validities and therefore I need a way to get buildbarn to reload these periodically, like it does for TLS certificates. This PR implements a general file-based mechanism for injecting file contents into outgoing gRPC requests. This seemed to me like a reasonably general way of doing it, but it's just an initial idea -- I'm open to doing this in a different way. 

This allows add_metadata_jmespath_expression to refer to the contents of certain pre-declared files which are reloaded on a user-specified schedule. For example, if you want to use k8s bound service account tokens you could add the following to your Pod:
```yaml
  volumes:
  - name: token-vol projected: sources:
      - serviceAccountToken: audience: my-buildbarn-instance expirationSeconds: 3600 path: buildbarn
```
Assuming this is mounted at '/tokens', you could then specify in your buildbarn config:
```jmespath
  addMetadataJmespathExpression: |||
    {
      "authorization": [join(' ', ['bearer', files.token])]
    }
  |||,
  addMetadataJmespathFiles: [
    {
      key: "token",
      path: "/tokens/buildbarn",
      refreshInterval: "1800s",
    }
  ]
```
This is quite useful for k8s service account tokens, as the maximum validity is often capped. Likewise this can also be used for Google service account id tokens, which also have a relatively short maximum validity.


tomgr:
Hi Ed,

That's true - there's nothing special about this. I like your idea. Are you happy for me to migrate all of them? I think this might be the existing list:
```
./pkg/proto/configuration/jwt/jwt.proto:69:  string claims_validation_jmespath_expression = 5;
./pkg/proto/configuration/jwt/jwt.proto:90:  string metadata_extraction_jmespath_expression = 6;
./pkg/proto/configuration/grpc/grpc.proto:90:  string add_metadata_jmespath_expression = 12;
./pkg/proto/configuration/grpc/grpc.proto:368:    string peer_credentials_jmespath_expression = 6;
./pkg/proto/configuration/grpc/grpc.proto:435:  string validation_jmespath_expression = 3;
./pkg/proto/configuration/grpc/grpc.proto:448:  string metadata_extraction_jmespath_expression = 4;
./pkg/proto/configuration/auth/auth.proto:41:    string jmespath_expression = 4;
./pkg/proto/configuration/http/http.proto:149:  string metadata_extraction_jmespath_expression = 6;
```
I presume the intent is that all of those will change type to be `buildbarn.jmespath.Expression`?

EdSchouten:
Exactly!

tomgr:
I've gone ahead and made this change, and I've also added the test vectors whilst I was at it. I'd find that useful.

The code now optionally takes a program.Group. This is wired into a few places, but not into the overall grpc Client construction as that's a larger refactor.

One thing to note here is that the Search function that I have exposed takes a `map[string]any` instead of a `any` (unlike the go-jmespath library). This was fine in every place in the codebase except for the OIDC claims processing in pkg/http/oidc_authenticator.go where I had to change the types that the claims were unmarshled from. However my understanding is that oidc claims are guaranteed to be JSON dictionaries, so I don't think this is an issue.

pkg/jmespath/jmespath.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// NewJMESPathExpression creates a new JMESPath expression.Expression
+// from the provided configuration. This will also evaluate all test vectors
+// and return an error if any of them fail.
+//
+// The group parameter can be optionally supplied: if it is not nil, then
+// any background refresher threads will be run using the group. Otherwise
+// background threads will be executed as a normal goroutine.
+func NewJMESPathExpression(config *pb.Expression, group program.Group) (*Expression, error) {
```
EdSchouten: Could we call this `NewExpressionFromConfiguration()`?

- Remove `JMESPath` from the name, as it is implied from the package name.
- Add `FromConfiguration()` to be more consistent with other functions that construct objects based on a Protobuf configuration message.

pkg/grpc/jmespath_extractor_test.go
```diff
@@ -83,3 +88,68 @@ func TestJMESPathMetadataExtractorAuthMatchToHeterogenousSlice(t *testing.T) {
...
+					ExpectedOutput: util.Must(structpb.NewValue(map[string]any{
+						"authorization": []any{"tv-token-value"},
+					})),
+					FileContents: map[string]string{
+						"token": "tv-token-value",
+					},
+				},
+			},
+		},
+		jmespath.NewTestGroupFromContext(ctx),
```
EdSchouten: Instead of coming up with our own implementation here, what are your thoughts on simply using the existing `program.RunLocal()` function?

pkg/proto/jmespath/jmespath.proto
```diff
@@ -0,0 +1,43 @@
...
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/struct.proto";
+
+option go_package = "github.com/buildbarn/bb-storage/pkg/proto/jmespath";
+
+message Expression {
+  // The JMESPath expression itself.
+  string expression = 1;
+  // Any files that should be made available to the JMESPath expression.
```
EdSchouten: Just for consistency with the rest of the .proto files, could we keep an empty line in between each of the fields? I started doing this, so that fields with long pieces of documentations are easier to read.

pkg/proto/jmespath/jmespath.proto
```diff
@@ -0,0 +1,43 @@
...
+
+message File {
+  // The key under which the file contents will be made available to
+  // the JMESPath expression. For example, if this is token, then the
+  // JMESPath expression can refer to the file contents as files.token.
+  string key = 1;
+  // The path to the file to read.
+  string path = 2;
+  // How often the file should be reloaded.
+  google.protobuf.Duration refresh_interval = 3;
```
EdSchouten: I know that we added a refresh interval for some of the TLS bits, but I do feel like this is a bit too much flexibility. Is there some sane default we could use here? Once every 60 seconds?

pkg/proto/jmespath/jmespath.proto
```diff
@@ -0,0 +1,43 @@
...
+}
+
+message TestVector {
+  // The input to the JMESPath expression. This should be a map with string
+  // keys.
```
EdSchouten: If it should be a map with string keys, then I recommend we use `google.protobuf.Struct`. :-)

pkg/proto/jmespath/jmespath.proto
```diff
@@ -0,0 +1,43 @@
...
+  google.protobuf.Value input = 1;
+  // The expected output of evaluating the JMESPath expression with
+  // input.
+  google.protobuf.Value expected_output = 2;
+  // If the JMESPath expression uses files, then this field should
+  // contain the contents of each of the files. If one of the files is
+  // not provided, then the test vector will fail.
+  map<string, string> file_contents = 3;
```
EdSchouten: Instead of having this be an explicit field in the test vector, couldn't we ask the user to simply provide it as part of the input?
```json
{
  "input": {
    "files": {
        ....
    }
  },
  "expectedOutput": ...
}

File: pkg/proto/configuration/jwt/jwt.proto [hunk 7]
```diff
@@ -66,7 +67,7 @@ message AuthorizationHeaderParserConfiguration {
   // authentication decisions; not authorization decisions. The reason
   // being that validation failures will cause gRPC requests to fail
   // with code UNAUTHENTICATED; not PERMISSION_DENIED.
-  string claims_validation_jmespath_expression = 5;
+  buildbarn.jmespath.Expression claims_validation_jmespath_expression = 5;
```
EdSchouten: I think that this could use some clarification on what the interaction with "files" is. Maybe the right thing to do is to put all of the token's fields in a nested message. Right now the input is of the form:

```json
{"aud": "MyBuildCluster", "exp": 12345, "scopes": [...]}
```

But maybe it's better if we turned that into:
```json
{
  "payload": {"aud": "MyBuildCluster", "exp": 12345, "scopes": [...]},
  "files": {
    ...
  }
}
```

That would also address the issue that you currently cannot access any of the fields in the token's header.

Mixing "files" with the other fields in the token is something we should absolutely not do, because that's bad from a security perspective.

pkg/jmespath/jmespath.go
```diff
@@ -0,0 +1,216 @@
...
+			} else {
+				go refresher(context.Background())
+			}
+		}
+	}
+
+	return provider, nil
+}
+
+func readFile(path string) (string, error) {
```
EdSchouten: From a usability perspective, wouldn't it make more sense to assume the file's contents are JSON and have that parsed? Because right now you can only get all of the data as a single string. This means you can't do something like:

```
contains(username, files.allowed_users)
```

Where the `allowed_users` file contains a JSON list of users.

pkg/grpc/jmespath_extractor.go
```diff
@@ -22,21 +22,22 @@ import (
 //
 //	{
 //	    "authenticationMetadata": value,
+//		"files": map<string, string>,
```
EdSchouten: Inconsistent whitespace?

pkg/proto/configuration/auth/auth.proto
```diff
@@ -38,7 +39,7 @@ message AuthorizerConfiguration {
     //
```
EdSchouten, line 40: This should also mention the existence of "files".

pkg/proto/jmespath/jmespath.proto
tomgr: I've currently implemented this so that if the refresh_interval is zero (i.e. the default), then we don't do any reloading at all. I think that might be a reasonable default. I suspect most files don't need reloading regularly: it will just be token-like things.

What do you think?

pkg/proto/jmespath/jmespath.proto
tomgr: Nice - done!

pkg/proto/configuration/jwt/jwt.proto
tomgr: Good point. I've done this, but this is something quite breaking to users. I guess I ought to at least update the bb-deployments examples after this is merged (if they use these keys at all - I guess they might not).

pkg/jmespath/jmespath.go
tomgr: That's an interesting idea and could be useful. However I think it would be useful to be able to get the entire file's contents as a single string as well. For example, k8s bound service account tokens are JWTs so have to be supplied as-is, and the way kubernetes makes them available would make it difficult to embed them into JSON.

One option might be to have an optional format field in the File struct:
```
enum Format {
  // The file's contents are not parsed at all and are available as a string
  // as `files.key`.
  Raw = 0;
  // The file is assumed to contain a valid JSON document which will be
  // parsed and made available as a nested structure under the key
  // `files.key`.
  JSON = 1;
}
```

tomgr:
Thanks for the review - I've marked the conversations as resolved when I've done the corresponding thing. If you'd prefer a different workflow please let me know.

pkg/proto/jmespath/jmespath.proto
EdSchouten: Wait. If it's really just about supporting those kinds of use cases, can't you just use the existing set of features that Jsonnet provides? For example, use `importstr` to load a token from disk?

tomgr: That works for loading tokens from disk at the point that the buildbarn configuration is read (and is what we’re doing now), but if the buildbarn process is long lived, the token might expire whilst buildbarn is running so it needs to be reloaded whilst buildbarn is running. This is the use case I’m interested in: gcloud tokens expire after 1 hour and our k8s tokens have a maximum lifetime of 24 hours. 

That said, the default of not refreshing the token is equivalent to using importstr. So indeed maybe the default I’ve implemented is not sensible: I could change the default to 60 seconds if you prefer? That might be more appropriate given the likely use case.

EdSchouten: Sure! I'd say, just hardcode 60 seconds. Don't even provide a configuration option for it.

pkg/proto/configuration/jwt/jwt.proto
```diff
@@ -87,5 +102,5 @@ message AuthorizationHeaderParserConfiguration {
   // extracted from the token:
```
EdSchouten, line 123: I think the example above needs to be changed to use `payload.given_name` and `payload.family_name`?

pkg/jmespath/jmespath.go
```diff
@@ -0,0 +1,203 @@
+package jmespath
```
EdSchouten: Maybe this file is better named `pkg/jmespath/expression.go`, as that's the primary type that this file declares. At least, that would be more consistent with the rest of the code base.

pkg/jmespath/jmespath.go
```diff
@@ -0,0 +1,203 @@
...
+	files      *metadataFileProvider
+}
+
+// NewExpressionFromConfiguration creates a new JMESPath expression.Expression
+// from the provided configuration. This will also evaluate all test vectors
+// and return an error if any of them fail.
+//
+// The group parameter can be optionally supplied: if it is not nil, then
+// any background refresher threads will be run using the group. Otherwise
+// background threads will be executed as a normal goroutine.
```
EdSchouten: I'd appreciate it if we used the occasion to plumb through the use of `program.Group` instead.

pkg/jmespath/jmespath.go
EdSchouten: For now, let's just implement the parts that you need. :-)

pkg/proto/jmespath/jmespath.proto
tomgr: I've done this, but I've had to remove a couple of parts of a couple of tests which checked the refreshing was working -- those relied on being able to set the refresh period to something very short.

pkg/proto/configuration/jwt/jwt.proto
```diff
@@ -62,11 +63,25 @@ message AuthorizationHeaderParserConfiguration {
   //
   // More details about JMESPath can be found at https://jmespath.org/.
   //
+  // This JMESPath expression will be evaluated against a JSON object
+  // with the following structure:
+  //
+  //     {
+  //       "files": ...
+  //       "payload": ...
```
EdSchouten: Probably worth documenting this as:

```
"files": map<string, any>,
"payload": any
```

For consistency with the other places we use JMESPath.

pkg/jmespath/jmespath.go
```diff
@@ -0,0 +1,201 @@
...
+			for {
+				select {
+				case <-t.C:
+					contents, err := readFile(file.Path)
+					if err != nil {
+						log.Printf("Failed to reload %s file: %v", file.Path, err)
+					} else {
+						provider.lock.Lock()
+						provider.currentContents[file.Key] = contents
+						provider.lock.Unlock()
```
EdSchouten: What are your thoughts on spawning a single coroutine per expression, as opposed to one per file? Then that goroutine could construct the entirely new "files" map, as opposed modifying the existing map in place. The advantage is that you can then use a `sync.Atomic[map[string]any]` instead of `sync.RWMutex`.

pkg/proto/jmespath/jmespath.proto
EdSchouten: For tests that depend on time/delays, consider changing `NewExpressionFromConfiguration()` to take a `clock.Clock`. Then you can eliminate delays in tests entirely.

tomgr: Thanks - have done this and used a `clock.Clock`. I still needed a small delay as I don't know how quickly the refresh will happen but this does make it testable.

pkg/jmespath/jmespath.go
tomgr: I've done this, and have tried to be consistent with existing code in terms of the `program.Group` argument. It's touched quite a lot of places, in particular inside the blob access creator. This could have been avoided by putting the `program.Group` inside the `grpc.ClientFactory`, but this would have been a bit inconsistent with other existing code.

pkg/jmespath/jmespath.go
tomgr: Sounds like a good idea - I've done this.

cmd/bb_copy/main.go
```diff
@@ -42,6 +42,7 @@ func main() {
 		grpcClientFactory := grpc.NewBaseClientFactory(grpc.BaseClientDialer, nil, nil, nil)
 
 		blobAccessCreator := blobstore_configuration.NewCASBlobAccessCreator(
+			dependenciesGroup,
```
EdSchouten: Why can't this use the `dependenciesGroup` that is already passed to `NewBlobAccessFromConfiguration()`?

cmd/bb_copy/main.go
```diff
@@ -62,7 +63,7 @@ func main() {
 			configuration.Replicator,
 			source.BlobAccess,
 			sink,
-			blobstore_configuration.NewCASBlobReplicatorCreator(grpcClientFactory),
+			blobstore_configuration.NewCASBlobReplicatorCreator(dependenciesGroup, grpcClientFactory),
```
EdSchouten: Same here: I would make `dependenciesGroup` an argument of `NewBlobReplicatorFromConfiguration()`, just like how it's already an argument of `NewBlobAccessFromConfiguration()`.

pkg/jmespath/expression.go
```diff
@@ -0,0 +1,190 @@
...
+// The group parameter is required when there are files and is used to schedule
+// refreshes of the file contents.
+func NewExpressionFromConfiguration(config *pb.Expression, group program.Group, clock clock.Clock) (*Expression, error) {
+	expression, err := jmespath.Compile(config.Expression)
+	if err != nil {
+		return nil, err
+	}
+
+	var files *metadataFileProvider
+	if config.Files != nil {
```
EdSchouten: My recommendation would be to use `len(config.Files) > 0` here. I don't think Protobuf makes a hard guarantee that empty lists unmarshal as `nil` (even though it is the case in practice).

pkg/jmespath/expression.go
```diff
@@ -0,0 +1,190 @@
...
+		err := expr.checkTestVector(t)
+		if err != nil {
+			return nil, util.StatusWrapf(err, "Failed to validate JMESPath expression %q with test vector %q", config.Expression, t.Input)
+		}
+	}
+
+	return expr, nil
+}
+
+// MustCompile creates a Expression from a string, panicing if
```
EdSchouten: s/panicing/panicking/?

pkg/jmespath/expression.go
```diff
@@ -0,0 +1,190 @@
...
+					contents, err := readFile(file.Path)
+					if err != nil {
+						log.Printf("Failed to reload %s file: %v", file.Path, err)
+						newContents[file.Key] = lastContents[file.Key]
+					} else {
+						newContents[file.Key] = contents
+					}
+				}
+				provider.currentContents.Store(&newContents)
+			case <-ctx.Done():
```
EdSchouten, line 162: Nit: If we return before `<-t` triggers, we should call `t.Stop()`.

pkg/jmespath/expression.go
```diff
@@ -0,0 +1,190 @@
...
+		actualJSON, _ := json.Marshal(actual)
+		return status.Errorf(codes.InvalidArgument, "Test vector failed: expected %s, got %s", string(expectedJSON), string(actualJSON))
+	}
+
+	return nil
+}
+
+// metadataFileProvider makes the contents of files available as
+// metadata to JMESPath expressions.
+type metadataFileProvider struct {
```
EdSchouten: Out of curiosity, what's the need for having this struct? Once `Expression` has been constructed and returned:

- `.currentContents` is obviously needed by `Expression.Search()` to populate the "files" field.
- `.files` should only be accessed by our goroutine that refreshes the file contents.

Considering that this goroutine is already launched by `NewExpressionFromConfiguration()`, wouldn't it be simpler to just let this goroutine capture `.files` and only put `.currentContents` in the expression?

pkg/jmespath/expression_test.go
```diff
@@ -0,0 +1,263 @@
...
+	tokenFile := filepath.Join(t.TempDir(), "token.txt")
+	require.NoError(t, os.WriteFile(tokenFile, []byte("secret123"), 0o644))
+
+	ctrl := gomock.NewController(t)
+	clock := mock.NewMockClock(ctrl)
+
+	t.Run("AccessFileContents", func(t *testing.T) {
+		program.RunLocal(context.Background(), func(ctx context.Context, siblingsGroup, dependenciesGroup program.Group) error {
+			timer := mock.NewMockTimer(ctrl)
+			timerChan := make(chan time.Time, 1)
```
EdSchouten: In tests where you need timers that never trigger, feel free to just use `nil`. From [the Go language specification](https://go.dev/ref/spec#Channel_types):

> A nil channel is never ready for communication.

pkg/proto/jmespath/jmespath.proto
```diff
@@ -0,0 +1,43 @@
+syntax = "proto3";
```
EdSchouten: I think this ought to live at pkg/proto/configuration/jmespath, as it only pertains to the configuration aspect of Buildbarn binaries. We never intend to send these messages "across the wire".

EdSchouten (commented):
Thanks for continuing to work on this, @tomgr!

pkg/jmespath/expression.go
tomgr: Good point. I've removed the struct.

cmd/bb_copy/main.go
tomgr: Good point. I've now done this although it required adding the program.Group to `NewCustomBlobAccess`.

pkg/blobstore/configuration/icas_blob_access_creator.go
```diff
@@ -21,7 +22,7 @@ type icasBlobAccessCreator struct {
 // provided to NewBlobAccessFromConfiguration() to construct a
 // BlobAccess that is suitable for accessing the Indirect Content
 // Addressable Storage.
-func NewICASBlobAccessCreator(grpcClientFactory grpc.ClientFactory, maximumMessageSizeBytes int) BlobAccessCreator {
+func NewICASBlobAccessCreator(terminationGroup program.Group, grpcClientFactory grpc.ClientFactory, maximumMessageSizeBytes int) BlobAccessCreator {
```
EdSchouten: Could you revert this change, please? This function argument isn't used.

EdSchouten (commented):
Hey Thomas,

Thanks again for working on this. I have one small comment, and once that's dealt with we can merge this.

pkg/blobstore/configuration/icas_blob_access_creator.go
tomgr: Whoops sorry - done.

PR #264, opened 2025-07-25
Improve windows mmap block performance

tomgr:
I noticed that the write performance was very poor on Windows when running bb_worker. Specifically during the initial fetch of all of the blobs from the CAS, bb_worker was spending a lot of time stuck in the WriteAt function. Writing via the memory map reduced this back down to almost nothing.

Sorry I didn't notice this when I implemented this. bb_storage itself seemed to perform fine, it was only when I started experimenting with bb_worker on our actual codebase that I noticed the performance problems. There the different is stark - the initial fetch took sometimes took 300 seconds with the file handle approach, whereas with the memory map approach it takes less than a second.

### DIFF
--- a/pkg/blockdevice/memory_mapped_block_device_windows.go
+++ b/pkg/blockdevice/memory_mapped_block_device_windows.go
@@ -80,23 +80,11 @@ func (bd *memoryMappedBlockDevice) ReadAt(p []byte, off int64) (n int, err error
 }
 
 func (bd *memoryMappedBlockDevice) WriteAt(p []byte, off int64) (int, error) {
-	// Like the unix implementation, we let write actions go through
-	// the file handle.
-	nTotal := 0
-	for len(p) > 0 {
-		var overlapped windows.Overlapped
-		overlapped.Offset = uint32(off)
-		overlapped.OffsetHigh = uint32(off >> 32)
-		var bytesWritten uint32
-		err := windows.WriteFile(bd.fileHandle, p, &bytesWritten, &overlapped)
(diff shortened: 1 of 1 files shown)


tomgr:
I've looked into this further, and actually on Windows I think we need to avoid WriteFile for correctness. The documentation for CreateFileMapping states:

> A mapped file and a file that is accessed by using the input and output (I/O) functions ([ReadFile](https://learn.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-readfile) and [WriteFile](https://learn.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-writefile)) are not necessarily coherent.

There's also a blog post about this topic at https://devblogs.microsoft.com/oldnewthing/20241107-00/?p=110486 where the conclusion is similar - WriteFile should not be mixed with CreateFileMapping.

In any case, the performance difference is quite staggering. I wrote some benchmarks that did parallel reads and writes on a blockdevice. WriteFile is an order of magnitude slower than copy.

I can't find a way of hinting to the OS that we will write an entire page and thus don't need to do a read. The only idea I have would be to use asynchronous writes instead of synchronous writes, but that would require additional synchronisation as we would have to make sure that before doing a read, any overlapping writes have completed. However, I do not think this is guaranteed to work because of the above comment about WriteFile and a mapped file not necessarily being coherent.

EdSchouten:
Interesting. Yeah, if they don't guarantee coherency between reads/writes, then we must write through the memory map. Merged!

PR #267, opened 2025-08-11
Reduce NewLocalDirectory's desired access on Windows

tomgr:
NewLocalDirectory was previously always requesting GENERIC_WRITE and
FILE_LIST_DIRECTORY, meaning that it would fail on read-only
directories. This change adopts the same approach as the Unix
NewLocalDirectory implementation:

* It requests minimal permissions on the main handle;
* When more permissions are required, it creates a temporary handle to
  the same file or directory with the enhanced permissions.

With this change, bonanza can pass its readiness checks on Windows with
new in-progress WinFSP-based VFS implementation.


EdSchouten (approved):
Looking great, Thomas!

PR #268, opened 2025-08-11
Make NewBlockDeviceFromFile more resiliant on Windows

tomgr:
On Windows / and \ are both accepted as directory separators in most places. This change ensures that the we accept either when specifying the path to a block storage file.

### DIFF
--- a/pkg/filesystem/path/windows_format.go
+++ b/pkg/filesystem/path/windows_format.go
@@ -4,6 +4,9 @@ import (
 	"strings"
 
 	"github.com/buildbarn/bb-storage/pkg/util"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 type windowsFormat struct{}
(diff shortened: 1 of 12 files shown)


EdSchouten:
Hey Thomas,

I'm wondering: instead of doing all of these kinds of normalization using string processing functions, what are your thoughts on leveraging the existing `pkg/filesystem/path` package? Normalizing a pathname string can be done as follows:

```go
joinedPath, scopeWalker := path.EmptyBuilder.Join(path.VoidScopeWalker)
if err := path.Resolve(path.LocalFormat.NewParser(in), scopeWalker); err != nil {
    ...
}
normalizedPath, err := joinedPath.GetWindowString()
if err != nil {
    ...
}
```

I think the only issue is that `pkg/filesystem/path` is currently oblivious to UNC paths. But maybe it's worth adding support for that? Also happy to merge this as is.

EdSchouten:
You're on fire, Thomas!

pkg/filesystem/path/windows_format.go
```diff
@@ -41,6 +43,30 @@ func (p windowsParser) ParseScope(scopeWalker ScopeWalker) (next ComponentWalker
 			}
 			return next, windowsRelativeParser{stripWindowsSeparators(p.path[2:])}, nil
 		}
+
+		if (p.path[0] == '\\' && p.path[1] == '\\') || (p.path[0] == '/' && p.path[1] == '/') {
```
EdSchouten: Just to double check: `\/foo\bar` is not permitted, right? Otherwise we should write something like:
```go
if (p.path[0] == '\\' || p.path[0] == '/') && (p.path[1] == '\\' || p.path[1] == '/') {
```

pkg/filesystem/path/windows_format.go
```diff
@@ -41,6 +43,30 @@ func (p windowsParser) ParseScope(scopeWalker ScopeWalker) (next ComponentWalker
...
+			serverStart := 2
+			serverLen := strings.IndexAny(p.path[serverStart:], "\\/")
+			if serverLen != -1 {
```
EdSchouten: How should the path `\\` be interpreted? Right now we convert it to an absolute path.

pkg/filesystem/path/windows_format.go
```diff
@@ -4,6 +4,8 @@ import (
 	"strings"
 
 	"github.com/buildbarn/bb-storage/pkg/util"
+	"google.golang.org/grpc/codes"
```
EdSchouten, line 8: Very tiny nit: could you add an empty line above this line? Right now we try to keep imports grouped by 2nd level domain. "Why?" is a good question, but staying consistent makes sense.

pkg/filesystem/path/windows_format.go
tomgr: `\\` is definitely not a valid UNC path as a UNC path has to have both a server and share name. So I think keeping the current behaviour of parsing that as an absolute path seems reasonable to me.

EdSchouten: Gotcha! But we don't want that to throw an explicit error?

Should we also check that the server/share name are non-empty? So `if serverLen < 1 {` ?

pkg/filesystem/path/windows_format.go
tomgr: Good question. I've updated it to accept either on the grounds that I don't think it's dangerous: Windows does seemingly normalise slashes quite liberally so I can't imagine we will misparse a path like this.

pkg/filesystem/path/windows_format.go
tomgr: Good point. I've updated it so that we are stricter and give errors on empty server or share names, and also give errors if it's just `//`.

PR #269, opened 2025-08-12
Remove unused proto import and enable --fatal_warnings for protoc

tomgr:
This should mean such mistakes are caught in the future.


tomgr:
Sorry about this - I didn't notice the warning flash past when bazel was running. I've enabled fatal_warnings now which should help, but can back that out if you'd prefer not to.

EdSchouten:
Yes! Having that option enabled makes a lot of sense. (In my opinion it should even be the default.)

Can you regenerate the .proto/BUILD files to address the CI failure? Will merge it as soon as CI is happy.

PR #271, opened 2025-08-12
Fix createUpgradedHandle's use of InvalidHandle

tomgr:
This fixes a bug introduced (by me - sorry) in
48ed13203b8710ffdde5cc4441d7e58033a76957. RootDirectory needs to be 0, not InvalidHandle.


tomgr:
@EdSchouten Sorry about this - I was sure that I ran the tests before pushing but clearly I mixed up something.

Would you be open to me extending the github actions to also run CI on a proper windows runner?

PR #274, opened 2025-08-12
Add a windows CI job

tomgr:


tools/github_workflows/workflows_template.libsonnet
```diff
@@ -2,60 +2,76 @@
   local platforms = [
     {
       name: 'linux_amd64',
-      buildAndTestCommand: 'test --test_output=errors',
       buildJustBinaries: false,
```
EdSchouten: Is `buildJustBinaries: true` still used at this point? If not, probably good to remove this flag entirely?

.github/workflows/pull-requests.yaml
```diff
@@ -1,89 +1,133 @@
...
+                     "bazel_startup_flags": "",
+                     "cross_compile": true,
+                     "lint": true,
+                     "os": "ubuntu-latest",
+                     "platform_name": "linux_amd64",
+                     "upload": true
+                  },
+                  {
+                     "bazel_os": "windows",
+                     "bazel_startup_flags": "--output_base=D:/bazel_output",
```
EdSchouten: To prevent the `bazel` commands from getting messy, would it make sense to write this into `~/.bazelrc` instead?

tomgr: I can do but I'm not sure if it will improve it necessarily. We can't do:
```
startup:windows --output_base=
```
because that would force all Windows users to use that output base, even if no D drive existed.

What we could do is something like:
```
startup:gha-linux ...
startup:gha-windows ...
```
and then always add `--config`. But I'm not necessarily sure if that improves things.

What do you think?

EdSchouten: My thought was that the step that writes this line into .bazelrc could be marked:
```
"if": "matrix.host.platform_name == 'windows_amd64'",
```

tomgr: Good idea - I've just done that.

I've also made one more small change - I've updated the signature of `getWorkflows` to:
```
  getWorkflows(binaries, containers, setupSteps = []): {
```
I'd quite like to get bb-remote-execution to install WinFSP so that it can run the winfsp integration tests on Github CI, hence it will need a setup step to do something like `choco install winfsp`.

EdSchouten (approved):

Issue #158, opened 2022-12-23
Support for compression from buildbarn


eagleonhill:
Bazel remote protocol now have compression option. That would be very useful for cases where the bazel and build-barn is in different clusters, and also saves storage usage.

EdSchouten:
Hey @eagleonhill,

When compression support was added to the Remote Execution protocol, I was opposed to adding it in its given form. The reason being that the transmitting side does not properly send the size of the object to the receiving side. With the way LocalBlobAccess is designed, it would be pretty hard to allocate space for objects as they are being written, as it tries to store objects consecutively. This means that compression would either be limited to just on wire (not at rest), or bb-storage would need to do unnecessary copying of data.

Furthermore, one downside of compression is that it's a lot harder to provide fast random access to blobs. As Buildbarn relies heavily on virtual file systems (based on FUSE/NFSv4), this is something that we need. Even without compression the protocol's story around random access is already a bit messy, as there's no way to guarantee data integrity by reading CAS objects partially.

Because of that, I am currently spending time with the remote API working group to get chunking/decomposition of CAS objects added. See these PRs:

- https://github.com/bazelbuild/remote-apis/pull/236 <- In my opinion good to land
- https://github.com/bazelbuild/remote-apis/pull/235 <- In my opinion also good to land
- https://github.com/bazelbuild/remote-apis/pull/233 <- Still need to be rewritten on top of the two PRs above

Once we have proper support for chunking/decomposition, the story around compression will become a lot simpler. All (primitive) CAS objects will become bounded in size. This means that even if the receiving side doesn't know the size of a compressed CAS object, it can safely ingest it into memory to compute its size, before letting LocalBlobAccess write it to disk.

So in summary, the reason Buildbarn doesn't support compression is not because we don't care. We want to do it right, and that still takes a bit more time to sort out.

bytearth:
Hey @EdSchouten 
compression is also a very useful feature for us.
I have read the code of bazel-remote, in their decision, when receive bytestream.Write request with resource path like 
`{instance_name}/uploads/{uuid}/compressed-blobs/{compressor}/{uncompressed_hash}/{uncompressed_size}{/optional_metadata}`,  they will decode the compressed data and write to the file if use diskCache backend.
zstd supports ”streaming decompression“， so  they can decode and write in a loop, like this:
```
for {
    compressedData = src.Recv()
    cuncompressedData = zstd.Decode(compressedData)
    dst.write(cuncompressedData)
}
```
[...]

PR #277, opened 2025-08-13
DON'T REVIEW/WIP feat(grpcclients): create gRPC CAS client with compression

tyler-french:
This is the same as the `grpc` client, except `Bytestream.Read` and `Bytestream.Write` requests through the remote execution API will be compressed with ZSTD compression.

The server (storage backend) must support compression

e.g.
frontend:
```
  cas: { sharded: { grpcWithCompression: { address: { 'storage-1' } } } } 
```

storage-1
```
  supportedCompressors: [ 'ZSTD' ]
```


EdSchouten:
I'd say, don't make this a separate storage backend. Just make it so that the existing CAS client caches the supported compressors returned by the last call to GetCapabilities() and uses that to determine whether to send requests that use Standard.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -199,6 +199,13 @@ message BlobAccessConfiguration {
     // value. When gRPC calls are timed out a `DEADLINE_EXCEEDED` error
     // code will be returned.
     DeadlineEnforcingBlobAccess deadline_enforcing = 28;
+
+    // Read objects from/write objects to a GRPC service that implements
+    // the remote execution protocol, but uses Zstd compression for
+    // ByteStream.Read and ByteStream.Write. The server must support
+    // Zstd compression. If it is a bb_storage Instance, add 'ZSTD' to
+    // the list of supported compressors.
+    CompressedClientConfiguration compressed = 29;
```
EdSchouten: I'm not a big fan of how this is named. Compression shouldn't be treated as a different kind of backend. It's merely a behavioral aspect of the gRPC CAS client. Right now the problem is that we don't have a dedicated message for that. For all storage backends (AC, CAS, ...) we only provide a `buildbarn.configuration.grpc.ClientConfiguration`. My recommendation would be to do one of the following things instead:

- Don't make any changes to the configuration schema, and simply respect what the server announces via `GetCapabilities()`.
- Add a `GrpcBlobAccessConfiguration` and change the type of field `grpc` to that. This means that options pertaining to compression and such will be available for all backends, but they will be no-ops for !CAS. Though slightly confusing, it does have the advantage that setups that merely want to forward all CAS/AC/... traffic can continue to write something like this:

```jsonnet
{
  local backend = ...,
  contentAddressableStorage: backend,
  actionCache: backend,
  initialSizeClassCache: backend,
  ...
}
```

- Rename `CompressedClientConfiguration` to `GrpcContentAddressableStorageBlobAccessConfiguration` and this `compressed` field to `grpc_content_addressable_storage`.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -946,3 +953,14 @@ message DeadlineEnforcingBlobAccess {
...
+
+message CompressedClientConfiguration {
+  // Base GRPC client configuration. Must be a GRPC client that
+  // announces support for Zstd CAS compression.
+  buildbarn.configuration.grpc.ClientConfiguration grpc = 1;
+
+  // Minimum blob size in bytes to trigger compression.
+  // Blobs smaller than this threshold will use the non-compressed client
+  // for better performance on small objects.
+  int64 compression_threshold_bytes = 2;
```
EdSchouten: Is this feature necessary? Did you perform any measurements that justify adding this?

pkg/blobstore/grpcclients/cas_zstd_blob_access.go
```diff
@@ -0,0 +1,265 @@
+package grpcclients
```
EdSchouten: I don't see why this needs to be a separate backend. Please integrate this functionality into the existing `cas_blob_access.go`.

pkg/blobstore/grpcclients/cas_zstd_blob_access.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func (ba *casWithZstdBlobAccess) Get(ctx context.Context, digest digest.Digest) buffer.Buffer {
+	if digest.GetSizeBytes() < ba.compressionThresholdBytes {
+		return ba.uncompressedBlobAccess.Get(ctx, digest)
+	}
+
+	ctxWithCancel, cancel := context.WithCancel(ctx)
+	resourceName := digest.GetByteStreamReadPath(remoteexecution.Compressor_ZSTD)
+	client, err := ba.byteStreamClient.Read(
```
EdSchouten: This currently fails to do negotiation with the server to see whether compression is supported. It would be nice if the `GetCapabilities()` method of this backend kept an eye on the `SupportedCompressors` returned by the server and have `Get()` use `Put()` respect this. Furthermore, if `Get()` or `Put()` is called before any call to `GetCapabilities()` is made, it may call `GetCapabilities()` directly.

pkg/blobstore/grpcclients/cas_zstd_blob_access.go
```diff
@@ -0,0 +1,265 @@
...
+	if !slices.Contains(cacheCapabilities.SupportedCompressors, remoteexecution.Compressor_ZSTD) {
+		return nil, status.Error(codes.FailedPrecondition, "Server does not support ZSTD compression")
+	}
+
+	// Only return fields that pertain to the Content Addressable
+	// Storage. Include compression support information.
+	return &remoteexecution.ServerCapabilities{
+		CacheCapabilities: &remoteexecution.CacheCapabilities{
+			DigestFunctions:      digest.RemoveUnsupportedDigestFunctions(cacheCapabilities.DigestFunctions),
+			SupportedCompressors: cacheCapabilities.SupportedCompressors,
```
EdSchouten: I don't think that propagating `supported_compressors` here makes sense. Doing so would cause these compressors to be merged with what's specified in bb_storage's configuration.

EdSchouten (changes requested):

aron-muon:
Does this close https://github.com/buildbarn/bb-storage/issues/158?

pkg/proto/configuration/blobstore/blobstore.proto
tyler-french: This matches the configuration in Bazel, but yes, really small files are not worth the overhead to compress. Bazel defaults this to 100. We can either keep it at 100 bytes, or make it configurable.

For certain situations, it might be preferable to set it at something like 64k since the number of packets wont change much.

Here's the analysis: https://github.com/buildbuddy-io/buildbuddy/pull/7009#discussion_r1676117359

pkg/blobstore/grpcclients/cas_blob_access.go
```diff
@@ -62,11 +76,147 @@ func (r *byteStreamChunkReader) Close() {
...
+}
+
+func (w *zstdByteStreamWriter) Close() error {
+	if err := w.client.Send(&bytestream.WriteRequest{
+		ResourceName: w.resourceName,
+		WriteOffset:  w.writeOffset,
+		FinishWrite:  true,
+	}); err != nil {
+		w.cancel()
+		w.client.CloseAndRecv()
```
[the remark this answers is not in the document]
tyler-french: error path ensures we don't block waiting for a response that won't come

pkg/blobstore/grpcclients/cas_zstd_blob_access.go
tyler-french: Get/Put now call GetCapabilities via sync.Once before using compression, and respect the server's SupportedCompressors.

pkg/proto/configuration/blobstore/blobstore.proto
tyler-french: Added GrpcBlobAccessConfiguration with enable_zstd_compression field that's a no-op for non-CAS backends.

aron-muon:
Hi @EdSchouten  - do you have time to re-review this? We are very excited about this implementation

tyler-french:
> > While I was testing this I noticed some OOMs happen sometimes under high load. I'll probably want to do some additional work to make sure we can configure how much memory can be used by compression/uncompression
> 
> 
> 
> I've put together a bounded ZSTD pool implementation that addresses this - it limits concurrent encoders/decoders via semaphores and reuses them via sync.Pool.
> 
> 
> 
> https://github.com/tyler-french/bb-storage/pull/1
> 
> 
> 
> Default config caps memory at ~320MB (16 encoders × 4MB + 32 decoders × 8MB). Happy to adjust based on your feedback!
> 
> 

@aron-muon nice! Before I dig into applying that, just want to check on preferences for these changes.

@EdSchouten do you prefer we merge these changes atomically or in two PRs? And, if separate, would you prefer we first create the ZSTD pool, or first enable support?

pkg/blobstore/configuration/cas_blob_access_creator.go
```diff
@@ -89,14 +89,18 @@ func (bac *casBlobAccessCreator) NewCustomBlobAccess(terminationGroup program.Gr
 			DigestKeyFormat: base.DigestKeyFormat,
 		}, "existence_caching", nil
 	case *pb.BlobAccessConfiguration_Grpc:
-		client, err := bac.grpcClientFactory.NewClientFromConfiguration(backend.Grpc, terminationGroup)
+		client, err := bac.grpcClientFactory.NewClientFromConfiguration(backend.Grpc.Client, terminationGroup)
 		if err != nil {
 			return BlobAccessInfo{}, "", err
 		}
+		compressionThresholdBytes := int64(0)
```
EdSchouten: Have you actually performed measurements that show that having such a threshold is necessary? My gut feeling tells me that compressing 100 bytes of data is far less effort than pushing it through a network socket...

pkg/blobstore/grpcclients/cas_blob_access.go
```diff
@@ -62,11 +77,137 @@ func (r *byteStreamChunkReader) Close() {
...
 const resourceNameHeader = "build.bazel.remote.execution.v2.resource-name"
 
+// shouldUseCompression checks if compression should be used for a blob of the given size.
+// It also ensures GetCapabilities has been called to negotiate compression support.
+func (ba *casBlobAccess) shouldUseCompression(ctx context.Context, digest digest.Digest) bool {
+	if ba.compressionThresholdBytes <= 0 || digest.GetSizeBytes() < ba.compressionThresholdBytes {
+		return false
+	}
+
+	// If GetCapabilities fails, fallback to no compression.
```
EdSchouten: I'm not a big fan of this approach. It means that the order in which processes is started also determines whether compression is used. For example, if bb-frontend is launched before bb-storage and the initial RPCs between the two fail, you don't get compression.

Given that a call to Get() is inherently more expensive than a call to ByteStream Read(), do we even care about coordination/locking? Why can't we just declare `supportedCompressors` as an `atomic.Pointer[[]remoteexecution.Compressor_Value]`? If it returns `nil`, we call `GetCapabilities()` without any in-flight deduplication.

The downside of `sync.Once` is that it's oblivious of `context.Context`. This means that the current code will keep goroutines waiting, even if their context is canceled.

pkg/blobstore/grpcclients/cas_blob_access.go
```diff
@@ -62,11 +77,137 @@ func (r *byteStreamChunkReader) Close() {
...
+	_, err := w.client.CloseAndRecv()
+	w.cancel()
+	return err
+}
+
 const resourceNameHeader = "build.bazel.remote.execution.v2.resource-name"
 
+// shouldUseCompression checks if compression should be used for a blob of the given size.
+// It also ensures GetCapabilities has been called to negotiate compression support.
+func (ba *casBlobAccess) shouldUseCompression(ctx context.Context, digest digest.Digest) bool {
```
EdSchouten: `shouldUseZSTDCompression()` ?

pkg/blobstore/grpcclients/cas_blob_access.go
```diff
@@ -89,19 +239,65 @@ func (ba *casBlobAccess) GetFromComposite(ctx context.Context, parentDigest, chi
...
+			cancel()
+			if _, closeErr := client.CloseAndRecv(); closeErr != nil {
+				return status.Errorf(codes.Internal, "Failed to close client: %v and create zstd writer: %v", closeErr, err)
+			}
+			return status.Errorf(codes.Internal, "Failed to create zstd writer: %v", err)
+		}
+
+		if err := b.IntoWriter(zstdWriter); err != nil {
+			if zstdCloseErr := zstdWriter.Close(); zstdCloseErr != nil {
+				err = errors.Join(err, zstdCloseErr)
```
EdSchouten: The current practice is to use `util.StatusFromMultiple()`, as that ends up preserving gRPC status metadata. However, do we think that the error returned by `zstdWriter.Close()` provides any value in addition to `b.IntoWriter()`? I would argue it doesn't...

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -946,3 +946,13 @@ message DeadlineEnforcingBlobAccess {
   // The backend to which all operations are delegated.
   BlobAccessConfiguration backend = 2;
 }
+
+message GrpcBlobAccessConfiguration {
+  // Base GRPC client configuration.
```
EdSchouten: s/GRPC/gRPC/

pkg/blobstore/configuration/cas_blob_access_creator.go
tyler-french: I have not personally, no. But having Bazel use 100 as a threshold, and also other backends use the same threshold, might cause some confusion if we change it. The intention here is to just have everything the same.

The 100 value came not from comparison with networks, but that the size of the compressed file is generally bigger than the uncompressed file.

EdSchouten: I don't think we really need to worry about the compressed file being larger than the uncompressed one. The reason being that (almost) all high quality compression algorithms detect this. Once they do, they resort to storing the uncompressed data (albeit with a header/trailer). See this comment on StackOverflow by Mark Adler, the author of zlib/gzip:

https://stackoverflow.com/a/23578269

aron-muon:
Hi @EdSchouten , mind re-reviewing?

aron-muon:
Hey @EdSchouten, is this a better week to review?

EdSchouten:
Merged!

PR #278, opened 2025-08-14
Expose some additional types for handling reparse points on Windows

tomgr:
These are needed for the WinFSP file system support.


pkg/filesystem/windowsext/types.go
```diff
@@ -3,7 +3,13 @@
 
 package windowsext
```
EdSchouten, line 4: To be honest, I've never been a big fan of the existence of this package. These are all fairly generic definitions that could also be made part of `golang.org/x/sys/windows`. Would you by any chance be interested in submitting PRs against that package to see how willing they are to merge such changes?

(That said, I'd be more than happy to merge this change if it prevents you from being stuck. Just let me know what to do here.)

EdSchouten (approved):

pkg/filesystem/windowsext/types.go
tomgr: If you could merge as-is please that would help as the WinFSP work requires this. Though if you prefer I could add the necessary extra bits privately in bb-remote-execution. It just seemed nicer to put them here where the other types were.

I'll submit a PR for `golang.org/x/sys/windows` as well and if that gets merged I'll follow-up here and delete this file. But I'm guessing that will take a while as it would require a new go release.

EdSchouten: SGTM. Thanks!

PR #284, opened 2025-08-20
Improve pretty printing of Windows-style paths

tomgr:
This provides a neater way of constructing NT device paths which are required for some system calls (e.g. creation of symlinks).

### DIFF
--- a/pkg/filesystem/local_directory_windows.go
+++ b/pkg/filesystem/local_directory_windows.go
@@ -6,10 +6,8 @@ package filesystem
 import (
 	"io"
 	"os"
-	"path/filepath"
 	"runtime"
 	"sort"
-	"strings"
 	"syscall"
 	"time"
 	"unsafe"
@@ -149,12 +147,11 @@ func NewLocalDirectory(directoryParser path.Parser) (DirectoryCloser, error) {
 	if err := path.Resolve(directoryParser, scopeWalker); err != nil {
 		return nil, util.StatusWrap(err, "Failed to resolve directory")
 	}
-	pathString, err := path.LocalFormat.GetString(directoryPath)
+	pathString, err := directoryPath.GetWindowsString(path.WindowsPathFormatDevicePath)
 	if err != nil {
 		return nil, util.StatusWrap(err, "Failed to create local representation of directory")
 	}
-	absPath := "\\??\\" + pathString
-	return newLocalDirectory(absPath, true)
+	return newLocalDirectory(pathString, true)
 }
 
 func (d *localDirectory) enter(name path.Component, openReparsePoint bool) (*localDirectory, error) {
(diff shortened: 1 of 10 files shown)


pkg/filesystem/path/builder.go
```diff
@@ -73,16 +85,30 @@ func (b *Builder) GetUNIXString() string {
...
+		} else if b.server != "" {
+			out.WriteString(`\??\UNC\`)
+		} else {
+			// We cannot print relative paths as NT device paths, so fallback
+			// to the normal behaviour.
+		}
+	}
+
 	prefix := ""
 	if b.driveLetter != 0 {
```
EdSchouten, line 80: Could the if-statement that you added above ^^ be moved into this one? As in, write something like this:

```go
if b.driveLetter != 0 {
  if format == WindowsPathFormatDevicePath {
    out.WriteString(`\??\`)
  }
  out.WriteString(string(b.driveLetter))
  out.WriteString(":")
  prefix = "\\"
}
```

I think that would improve readability.

pkg/filesystem/path/builder.go
```diff
@@ -6,6 +6,18 @@ import (
 	"github.com/buildbarn/bb-storage/pkg/util"
 )
 
+// WindowsPathFormat represents different format options for Windows path strings.
+type WindowsPathFormat int
```
EdSchouten: A more logical place to house this enum would be `stringer.go`, where we also declare the interface for providing `GetWindowsString()`.

pkg/filesystem/local_directory_windows.go
```diff
@@ -778,17 +777,25 @@ func (d *localDirectory) Symlink(oldNameParser path.Parser, newName path.Compone
 	if err := path.Resolve(oldNameParser, scopeWalker); err != nil {
 		return err
 	}
-	oldName, err := path.LocalFormat.GetString(oldNamePath)
-	if err != nil {
-		return err
-	}
-	// Path with one leading slash (but not UNC) should also be considered absolute.
-	isRelative := !(oldName[0] == '\\' || filepath.IsAbs(oldName))
+	pathKind := oldNamePath.WindowsPathKind()
```
EdSchouten: Even though I'm fine with the changes to `.GetWindowsString()`, I'm not entirely convinced about this part. I have the feeling as if we're doing a bit too much "juggling" going on here. Also I wonder whether it all works correctly. For example, the `filepath.Rel()` will do an implicit `filepath.Clean()`. This means that `foo/../bar` will get simplified to just `bar`. This is not correct if `foo` is a symlink.

What are your thoughts on making this function a lot dumber? With that I mean that we just do something like this:
```go
// Get the path of the current directory.
handlePath, err := getAbsPathByHandle(d.handle)
if err != nil {
	return err
}

// Compute the absolute path of the new symlink's target.
absoluteTargetPath, scopeWalker := path.EmptyBuilder.Join(path.VoidScopeWalker)
if err := path.Resolve(path.WindowsFormat.NewParser(handlePath), scopeWalker); err != nil {
	return err
}
absoluteTargetPath, scopeWalker = absoluteTargetPath.Join(path.VoidScopeWalker)
if err := path.Resolve(oldNameParser, scopeWalker); err != nil {
	return err
}
absoluteTargetPathStr, err := absoluteTargetPath.GetWindowsString(path.WindowsPathFormatDevicePath)
if err != nil {
	return err
}

// Determine whether the target is a directory.
fi, err := os.Stat(targetPath)
isDir = err == nil && fi.IsDir()
return d.createNTFSSymlink(oldName, newName.String(), isRelative, isDir)
```

So it means we call `getAbsPathByHandle()` unconditionally, even if the symlink target path is already absolute. I think that's more maintainable.

PR #288, opened 2025-08-22
Fix a windows performance issue when shutting down block devices

tomgr:
Whilst writing the integration test for buildbarn in bb-deployments, I noticed that bb_storage took a long time to shutdown on Windows. When looking at the system monitor, bb_storage was saturating the disk writing out data on shutdown. This was unexpected, since the test only wrote a small amount of data to the cache, so only a small amount of the block device should have been used (hundreds of MBs out of the GBs in the block device).

It seems that when using SetEndOfFile, Windows will write out the entire file's contents when the process gets shutdown. This is a problem if buildbarn is started with a large block device but only uses a small part of it, like in the deployment tests.

Following the suggestions at https://devblogs.microsoft.com/oldnewthing/20110922-00/?p=9573 we now create a sparse file and instantly zero the entire file's contents (if it's new or is resized). This completely fixes the performance issues.

### DIFF
--- a/pkg/blockdevice/new_block_device_from_file_windows.go
+++ b/pkg/blockdevice/new_block_device_from_file_windows.go
@@ -50,6 +50,15 @@ func NewBlockDeviceFromFile(path string, minimumSizeBytes int, zeroInitialize bo
 		return nil, 0, 0, util.StatusWrapf(err, "Failed to open file %#v", path)
 	}
 
+	bd, sectorSizeBytes, sectorCount, err := createBlockDevice(path, minimumSizeBytes, zeroInitialize, handle)
+	if err != nil {
+		windows.CloseHandle(handle)
+		return nil, 0, 0, err
+	}
+	return bd, sectorSizeBytes, sectorCount, nil
+}
+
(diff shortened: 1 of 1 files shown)


pkg/blockdevice/new_block_device_from_file_windows.go
```diff
@@ -82,25 +89,99 @@ func NewBlockDeviceFromFile(path string, minimumSizeBytes int, zeroInitialize bo
...
+		fsctlSetSparse = 0x000900c4
+		// FSCTL_SET_ZERO_DATA
+		fsctlSetZeroData = 0x000980c8
+	)
+
+	// FILE_ZERO_DATA_INFORMATION
+	type fileZeroDataInformation struct {
+		FileOffset      int64
+		BeyondFinalZero int64
+	}
```
EdSchouten: Can we please submit patches against golang.org/x/sys/windows to add types like these?

tomgr: Yes sure. I found the two constants in there actually. I'm just working on a patch for fileZeroDataInformation.

tomgr: Right I've opened https://github.com/golang/sys/pull/261 for this, and I've also opened https://github.com/golang/sys/pull/260 for the reparse buffer types as we previously discussed.

PR #291, opened 2025-09-05
Fix test vector comparison

tomgr:
It seems that using reflect.DeepEqual on protobuf messages doesn't work correctly; seemingly equal things are different because of internal protobuf fields.

This switches the comparison to use the protocmp library which does proper equality comparison. I've also extended the test cases to cover this.

### DIFF
--- a/pkg/jmespath/expression.go
+++ b/pkg/jmespath/expression.go
@@ -6,7 +6,6 @@ import (
 	"log"
 	"maps"
 	"os"
-	"reflect"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/jmespath/expression.go
```diff
@@ -124,7 +125,7 @@ func (e *Expression) checkTestVector(t *pb.TestVector) error {
 		return util.StatusWrapf(err, "Failed to convert JMESPath result to protobuf value")
 	}
 
-	if !reflect.DeepEqual(t.ExpectedOutput, actualPb) {
+	if !cmp.Equal(t.ExpectedOutput, actualPb, protocmp.Transform()) {
```
EdSchouten: What are your thoughts on just using `proto.Equal()`?
https://pkg.go.dev/google.golang.org/protobuf/proto#Equal

tomgr: Oh much simpler - thanks! Done.

EdSchouten (approved):

PR #292, opened 2025-09-24
Allow oauth client credentials for http clients

jpopadak:
This moves the OAuth configuration from grpc to http package for use on http clients. Due to dependencies / cycles, the http package is split into client and server. Since I split the packages, I renamed `ClientConfiguration` to `client.Configuration`, and `ServerConfiguration` to `server.Configuration`.

Client credentials is currently the only supported OAuth http flow while grpc supports all 3 OAuth cases.

The reason for this change is to provide oauth support for downstream http clients in other buildbarn repos.


EdSchouten:
Hey! I just upgraded all third-party dependencies. I guess that if you were to rebase, that rules_go warning disappears.

cmd/sync_jwks_to_configmap/main.go
```diff
@@ -7,7 +7,7 @@ import (
 	"net/http"
 	"os"
 
-	bb_http "github.com/buildbarn/bb-storage/pkg/http"
+	bb_http "github.com/buildbarn/bb-storage/pkg/http/client"
```
EdSchouten: To improve readability, could you do a sweep of the tree for `bb_http` and replace that with `http_{client, server}` or something?

Edit: Ah, just saw you're already doing that in some places. Could you also do it here for consistency? Whether the `bb_` prefix should remain: I named it `bb_http` only because I wanted to prevent collisions with `net/http`. As that no longer applies, feel free to remove it.

pkg/grpc/base_client_factory.go
```diff
@@ -214,3 +207,27 @@ func (cf baseClientFactory) NewClientFromConfiguration(config *configuration.Cli
 		grpc.WithChainStreamInterceptor(streamInterceptors...))
 	return cf.dialer(context.Background(), config.Address, dialOptions...)
 }
+
+func perRPCCredentials(oauthConfig *http_configuration.OAuthConfiguration) (credentials.PerRPCCredentials, error) {
```
EdSchouten: Could we call this `newPerRPCCredentialsFromConfiguration()` instead, for consistency with some of the other functions in the tree?

pkg/http/client/oauth2_credentials.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/buildbarn/bb-storage/pkg/util"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// TokenSource uses the given OAuthConfiguration to create a token source for HTTP clients.
+func TokenSource(oauthConfig *configuration.OAuthConfiguration) (oauth2.TokenSource, error) {
```
EdSchouten: NewTokenSourceFromConfiguration()

pkg/grpc/base_client_factory.go
```diff
@@ -214,3 +207,27 @@ func (cf baseClientFactory) NewClientFromConfiguration(config *configuration.Cli
...
+	case *http_configuration.OAuthConfiguration_ServiceAccountKey:
+		perRPC, err = oauth.NewServiceAccountFromKey([]byte(credentials.ServiceAccountKey), oauthConfig.Scopes...)
+	case *http_configuration.OAuthConfiguration_ClientCredentials:
+		tokenSource, err := http_client.ClientCredentialsTokenSource(oauthConfig.Scopes, credentials)
+		if err != nil {
+			return nil, err
+		}
+		perRPC = oauth.TokenSource{TokenSource: tokenSource}
+	default:
+		return nil, status.Error(codes.InvalidArgument, "oauth client credentials are wrong: one of googleDefaultCredentials, serviceAccountKey, or clientCredentials should be provided")
```
EdSchouten: Instead of doing this, could we please do this?

```go
default:
    tokenSource, err := http_client.NewTokenSourceFromConfiguration(credentials)
    if err ! = nil {
        return nil, err
    }
    return oauth.TokenSource{TokenSource: tokenSource}, nil
```

That will also allow us to make `ClientCredentialsTokenSource()` private (or inline it into `NewTokenSourceFromConfiguration()`).

pkg/grpc/base_client_factory.go
```diff
@@ -214,3 +207,27 @@ func (cf baseClientFactory) NewClientFromConfiguration(config *configuration.Cli
...
+	}
+	if err != nil {
+		return nil, util.StatusWrap(err, "Failed to create oauth configuration")
+	}
+	return perRPC, err
```
EdSchouten: I'd suggest that we don't do this at the bottom of this function, but repeated it in each of the switch cases. That keeps the control flow simpler, and also allows us to have custom util.StatusWrap()s per case.

pkg/http/client/oauth2_token_source.go
```diff
@@ -0,0 +1,52 @@
+package client
```
EdSchouten, line 1: Given that the functions in this file are mostly about creating an OAuth2 TokenSource, can we call it oauth2_token_source.go instead?

pkg/proto/configuration/http/client/client.proto
```diff
@@ -0,0 +1,82 @@
...
+  // https://github.com/golang/go/issues/36026
+  // https://github.com/golang/go/issues/59690
+  bool disable_http2 = 6;
+
+  // Oauth authentication settings. HTTP only supports OAuthClientCredentials.
+  // More information: https://grpc.io/docs/guides/auth/
+  OAuthConfiguration oauth = 7;
+}
+
+message OAuthConfiguration {
```
EdSchouten: I see that in the code base we have a mixture of "oauth" and "oauth2". Is there anything we can do to tidy that up? Maybe "oauth2" is better, because it also matches golang.org/x/oauth2?

pkg/proto/configuration/http/client/client.proto
```diff
@@ -0,0 +1,82 @@
...
+  oneof credentials {
+    // Use default Google credentials. More information:
+    // https://developers.google.com/accounts/docs/application-default-credentials
+    google.protobuf.Empty google_default_credentials = 1;
+
+    // Service account private key to use to obtain access token.
+    string service_account_key = 2;
```
EdSchouten: Can we please document that these two options currently only work for gRPC?

pkg/grpc/base_client_factory.go
jpopadak: The one downside of this is the error message does not match anymore. 

`NewClientCredentialsTokenSource` returns `oauth http client credentials are wrong: only clientCredentials should be provided`, which is incorrect for this switch as it supports all 3.

pkg/http/client/oauth_adding_round_tripper.go
```diff
@@ -0,0 +1,23 @@
...
+
+	pb "github.com/buildbarn/bb-storage/pkg/proto/configuration/http/client"
+
+	"golang.org/x/oauth2"
+)
+
+// NewOAuthAddingRoundTripper is a decorator for RoundTripper that requests a
+// OAuth2 token from the TokenSource and add its to the HTTP header on all
+// outgoing requests.
+func NewOAuthAddingRoundTripper(base http.RoundTripper, config *pb.OAuth2Configuration) (http.RoundTripper, error) {
```
EdSchouten: NewOAuth2AddingRoundTripper()? Same for the name of the source file.

pkg/http/client/oauth2_token_source.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// NewTokenSourceFromConfiguration uses the given configuration to create a
+// token source for HTTP clients.
+func NewTokenSourceFromConfiguration(oauthConfig *configuration.OAuth2Configuration) (oauth2.TokenSource, error) {
```
EdSchouten: NewOAuth2TokenSourceFromConfiguration()? Probably a bit more accurate, given that this lives in pkg/http/client, which doesn't have "oauth2" in its name.

PR #294, opened 2025-09-26
Reformat script to handle embed, protos, formatting

jpopadak:
### Description
Add the reformat script used within Bonanza to bb-storage, and for use within other buildbarn downstream repos.

### Changes
- Add reformat.sh and `//tools:reformat` target
- Upgrade llvm to 18.1.8 for auto resolve of `darwin/arm64` + windows + linux
- Use runfiles to execute `clang-format`, `gofumt`, `go`, and `gazelle`
- Move lint to its own github workflow. A `no space left on device` kept occurring

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -541,7 +541,6 @@
                   {
                      "bazel_os": "linux",
                      "cross_compile": true,
-                     "lint": true,
                      "os": "ubuntu-latest",
                      "platform_name": "linux_amd64",
                      "upload": true
@@ -549,7 +548,6 @@
                   {
                      "bazel_os": "windows",
                      "cross_compile": false,
-                     "lint": false,
                      "os": "windows-latest",
                      "platform_name": "windows_amd64",
                      "upload": false
@@ -559,8 +557,8 @@
          }
       },
       "lint": {
-         "name": "lint ${{ matrix.host.os }}",
-         "runs-on": "${{ matrix.host.os }}",
+         "name": "lint",
+         "runs-on": "ubuntu-latest",
          "steps": [
             {
                "name": "Check out source code",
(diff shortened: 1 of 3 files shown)


tools/reformat.sh
```diff
@@ -0,0 +1,56 @@
...
+find bazel-bin/pkg/proto -name '*.pb.go' | while read f; do
+  cat $f > $(echo $f | sed -e 's|.*/pkg/proto/|pkg/proto/|')
+done
+#go get -d -u ./... || true
+go mod tidy || true
+
+# Gazelle
+find . -name '*.pb.go' -delete
+rm -f $(find . -name '*.proto' | sed -e 's/[^/]*$/BUILD.bazel/')
+bazel run //:gazelle
```
EdSchouten: Could this also be invoked through `$(location ...)`?

tools/reformat.sh
```diff
@@ -0,0 +1,56 @@
...
+# --- end runfiles.bash initialization v3 ---
+
+# Start in the root directory
+cd "$BUILD_WORKSPACE_DIRECTORY"
+
+# Go dependencies
+find . bazel-bin/pkg/proto -name '*.pb.go' -delete || true
+bazel build -k --output_groups=go_generated_srcs $(bazel query --output=label 'kind("go_proto_library", //...)') || true
+find bazel-bin/pkg/proto -name '*.pb.go' | while read f; do
+  cat $f > $(echo $f | sed -e 's|.*/pkg/proto/|pkg/proto/|')
```
EdSchouten: Just wondering whether there's some way we can make this work, even if .proto files are not stored under /pkg/proto. A hacky solution would be to extract the `gazelle:prefix` from the root BUILD file... (or from go.mod)? Then we can use `bazel-bin/**/${prefix}`

tools/reformat.sh
jpopadak: Doh. Yeah. 😄

tools/reformat.sh
jpopadak: I used `go list -m` to get the module path. Then modified the `find` queries to use `-path` with `*` since `**` didnt work nicely with a variable.

tools/reformat.sh
jpopadak: Gazelle ended up hanging using this. 😞

tools/reformat.sh
```diff
@@ -0,0 +1,80 @@
...
+  source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+  { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
+# --- end runfiles.bash initialization v3 ---
+
+# Check required binaries exist
+go="$(rlocation rules_go++go_sdk+main___download_0/bin/go)"
+if [ -z "$go" ]; then
+  echo "go binary not found"
+  exit 1
+fi
```
EdSchouten: Can we use `runfiles_rlocation_checked`, so that we don't need to write our own error checking? Thanks!

tools/reformat.sh
```diff
@@ -0,0 +1,80 @@
...
+source "${RUNFILES_DIR:-/dev/null}/$f" 2>/dev/null || \
+  source "$(grep -sm1 "^$f " "${RUNFILES_MANIFEST_FILE:-/dev/null}" | cut -f2- -d' ')" 2>/dev/null || \
+  source "$0.runfiles/$f" 2>/dev/null || \
+  source "$(grep -sm1 "^$f " "$0.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+  source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+  { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
+# --- end runfiles.bash initialization v3 ---
+
+# Check required binaries exist
+go="$(rlocation rules_go++go_sdk+main___download_0/bin/go)"
```
EdSchouten: I don't think we should be using canonical repo names here. Apparent repo names should work as well, assuming we `bazel_dep()` or `use_repo()` them. So it should be possible to use `rules_go/bin/go`, `cc_mvdan_gofumpt/gofumpt_/gofumpt`, etc.

tools/reformat.sh
jpopadak: `rlocation` uses `runfiles_rlocation_checked`. Also,`runfiles_rlocation_checked` does not look through dependencies.

EdSchouten (approved):

EdSchouten:
Will merge once CI is happy.

PR #327, opened 2026-03-02
feat: add bounded ZSTD encoder/decoder pool with memory management

aron-muon:
## Summary

The existing CAS gRPC client creates a new `zstd.Encoder` and `zstd.Decoder` for every ByteStream request. Under high concurrency (e.g., large Bazel builds with many parallel actions), this leads to unbounded memory growth, GC pressure, and potential OOM — each encoder allocates ~4MB and each decoder ~8MB, and the `klauspost/compress/zstd` library spawns internal goroutines that leak without explicit `Close()` calls ([klauspost/compress#264](https://github.com/klauspost/compress/issues/264)).

This PR introduces a `pkg/zstd` package with `Pool`, `Encoder`, and `Decoder` interfaces, a bounded pool implementation, and integrates it across the CAS client, ByteStream server, and reference-expanding blob access.

## Approach

**`pkg/zstd/` package** — clean interfaces for ZSTD encoder/decoder lifecycle:
- `Pool` interface with `NewEncoder(ctx, w)` / `NewDecoder(ctx, r)` returning `Encoder`/`Decoder`
- `Encoder.Close()` and `Decoder.Close()` handle flushing and release back to the pool
- `BoundedPool`: reuses encoders/decoders via `sync.Pool`, limits concurrency via semaphores, providing backpressure under load
- `UnboundedPool`: creates fresh encoders/decoders per call, for use when bounding is not required
- `MetricsPool`: decorator exposing Prometheus counters (`pool_acquisitions_total`, `pool_releases_total`, `pool_rejections_total`) and histogram (`pool_wait_duration_seconds`) with `object_type="Encoder"/"Decoder"` label
- `NewPoolFromConfiguration()`: creates a pool from protobuf config, automatically wrapped with metrics
- `NewReadCloser()`: decompressing `io.ReadCloser` backed by the pool

**Configuration** — `PoolConfiguration` proto in `pkg/proto/configuration/zstd/zstd.proto`, referenced from `ApplicationConfiguration.zstd_pool`. When set, creates a process-wide bounded pool. When absent, an unbounded pool is used. Compression is enabled for CAS clients by providing a non-nil pool (no separate bool flag).

**Integration:**
- CAS client (`cas_blob_access.go`): `Get` acquires a decoder eagerly in `Get()`; `Put` acquires a pooled encoder with `defer encoder.Close()`
- ByteStream server (`byte_stream_server.go`): uses pool for both compressed reads (encoding) and writes (decoding via `NewReadCloser`)
- Reference-expanding blob access: uses `NewReadCloser` with the shared pool
- All zstd usage consolidated — `pkg/util/zstd_reader.go` removed

## Testing

- Pool unit tests: acquire/release, concurrency limits, context cancellation, concurrent stress (50 goroutines), double-close safety
- CAS integration tests: pool exhaustion returning `ResourceExhausted`, encoder release (sequential puts with pool size 1), decoder release (sequential gets with pool size 1)
- Benchmarks: pooled vs non-pooled encoder performance


pkg/blobstore/grpcclients/cas_blob_access.go
```diff
@@ -42,13 +74,23 @@ type casBlobAccess struct {
...
 func NewCASBlobAccess(client grpc.ClientConnInterface, uuidGenerator util.UUIDGenerator, readChunkSize int, enableZSTDCompression bool) blobstore.BlobAccess {
+	return NewCASBlobAccessWithPool(client, uuidGenerator, readChunkSize, enableZSTDCompression, nil)
+}
+
+// NewCASBlobAccessWithPool creates a BlobAccess handle with a custom ZSTD pool.
+// If pool is nil, the default shared pool will be used.
+func NewCASBlobAccessWithPool(client grpc.ClientConnInterface, uuidGenerator util.UUIDGenerator, readChunkSize int, enableZSTDCompression bool, pool *BoundedZstdPool) blobstore.BlobAccess {
+	if pool == nil {
+		pool = GetDefaultZstdPool()
+	}
```
EdSchouten: The Buildbarn codebase uses the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern) everywhere. Please don't add constructs like these. Callers should just pass in the proper object.

pkg/blobstore/grpcclients/cas_blob_access.go
```diff
@@ -23,6 +23,37 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+// DefaultZstdPool is a shared pool for ZSTD encoders/decoders.
+// It is initialized with default settings on first use.
+// This can be overridden via SetDefaultZstdPool for custom configurations.
+var (
+	defaultZstdPool     *BoundedZstdPool
```
EdSchouten: Why do we need global state like this?

pkg/blobstore/grpcclients/zstd_config.go
```diff
@@ -0,0 +1,154 @@
...
+
+	// DefaultEncoderMemoryBytes is the estimated memory per encoder.
+	DefaultEncoderMemoryBytes = 4 * 1024 * 1024
+
+	// DefaultDecoderMemoryBytes is the estimated memory per decoder.
+	DefaultDecoderMemoryBytes = 8 * 1024 * 1024
+)
+
+// ZstdPoolConfig holds validated configuration for a BoundedZstdPool.
+type ZstdPoolConfig struct {
```
EdSchouten: What's the point in adding this entire structure if we don't provide any facilities for controlling these values?

pkg/blobstore/grpcclients/zstd_config.go
```diff
@@ -0,0 +1,154 @@
...
+
+	// DefaultDecoderMaxWindow is the default maximum decoder window size (8MB).
+	// Must be at least as large as the encoder window size used by remote peers.
+	DefaultDecoderMaxWindow = 8 << 20
+
+	// DefaultEncoderMemoryBytes is the estimated memory per encoder.
+	DefaultEncoderMemoryBytes = 4 * 1024 * 1024
+
+	// DefaultDecoderMemoryBytes is the estimated memory per decoder.
+	DefaultDecoderMemoryBytes = 8 * 1024 * 1024
```
EdSchouten: People may run bb_storage on a Raspberry Pi. Or they may run it on a 128 core EC2 instance.  I therefore think we shouldn't have any sane defaults like this.

pkg/blobstore/grpcclients/zstd_config.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return config
+}
+
+// ParseEncoderLevel converts a string level name to zstd.EncoderLevel.
+// Valid values: "fastest", "default", "better", "best"
+// Returns zstd.SpeedDefault for unrecognized values.
+func ParseEncoderLevel(level string) zstd.EncoderLevel {
+	switch level {
+	case "fastest":
```
EdSchouten: This function doesn't seem to get called anywhere. Also, Buildbarn uses Protobuf messages for specifying configuration options. Why do we need to do string parsing here? Can't we just match enum values?

pkg/blobstore/grpcclients/zstd_pool.go
```diff
@@ -0,0 +1,263 @@
...
+	p.encoderPool.New = func() interface{} {
+		enc, err := zstd.NewWriter(nil, p.encoderOptions...)
+		if err != nil {
+			// This should only fail with invalid options, which is a programming error
+			panic("failed to create ZSTD encoder: " + err.Error())
+		}
+		ew := &EncoderWrapper{Encoder: enc}
+
+		// CRITICAL: Finalizer prevents goroutine/memory leaks if caller
+		// forgets to return encoder to pool. The zstd.Encoder spawns
```
EdSchouten: We own the code. Why would we forget to return the encoder to the pool?

EdSchouten (commented):
I know you're trying to make the best out of the entire situation. But let me use this opportunity to go on record and say this:

This is a clear demonstration of why I so absolutely **hate** the way the Remote API working group added support for compression to REv2. Having all of this bloat in bb_storage is just awful.

pkg/blobstore/grpcclients/zstd_pool.go
aron-muon: Yeah good point, I removed the finalizers entirely

pkg/blobstore/grpcclients/zstd_config.go
aron-muon: Removed ParseEncoderLevel and the entire `zstd_config.go` file

pkg/blobstore/grpcclients/zstd_config.go
aron-muon: Agreed, removed hardcoded defaults. The pool is created at the configuration site with explicit values

pkg/blobstore/grpcclients/zstd_config.go
aron-muon: Removed ZstdPoolConfig entirely. The pool parameters are now passed directly within the protobuf configuration fields

pkg/blobstore/grpcclients/cas_blob_access.go
aron-muon: Removed all global state, `defaultZstdPool`, `sync.Once`, and `SetDefaultZstdPool`. The pool is created and passed explicitly by the caller.

pkg/blobstore/grpcclients/cas_blob_access.go
aron-muon: Removed the `if pool == nil` defaulting. Callers now pass the pool explicitly, and nil is only used when compression is disabled.

aron-muon:
> I know you're trying to make the best out of the entire situation. But let me use this opportunity to go on record and say this:
> 
> This is a clear demonstration of why I so absolutely **hate** the way the Remote API working group added support for compression to REv2. Having all of this bloat in bb_storage is just awful.

I can easily understand that. The compression support in REv2 adds a lot of plumbing. Happy to keep the implementation as lean as possible

aron-muon:
Hey @EdSchouten , mind letting the workflows run?
We are running the changes here in production within our Bazel cache, and are really pleased with the memory management

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -931,9 +931,48 @@ message GrpcBlobAccessConfiguration {
...
+
+  reserved 2;
+}
+
+message ZstdCompressionConfiguration {
+  // Maximum number of concurrent ZSTD encoding operations. Each
+  // encoder allocates memory proportional to the configured window
+  // size. When all encoder slots are in use, new Put operations will
+  // block until a slot becomes available, providing backpressure.
+  int64 max_encoders = 1;
```
EdSchouten: s/max_/maximum_/ for consistency with other configuration options.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -931,9 +931,48 @@ message GrpcBlobAccessConfiguration {
...
+  // Optional ZSTD compression configuration for ByteStream operations
+  // in the Content Addressable Storage (CAS). When set, ZSTD
+  // compression will be used if the server supports it. Has no effect
+  // for other storage types (AC, ICAS, etc.).
+  //
+  // If not set, compression is disabled and no encoder/decoder pools
+  // are allocated.
+  ZstdCompressionConfiguration zstd = 3;
+
+  reserved 2;
```
EdSchouten: Please add a comment above along these lines:

```proto
// Was 'enable_compression'. Use 'zstd' instead.
```

That said: does it really make sense to configure this on a per-BlobAccess basis? If you use sharding/mirroring, then you end up creating n zstd pools instead of just one. Also, in the Bytestream gRPC server we also need to perform compression and decompression. Wouldn't we want to use the same pool type there as well?

This makes me think: maybe this should be configured on the top level?

pkg/proto/configuration/blobstore/blobstore.proto
aron-muon: Renamed to `maximum_encoders` and `maximum_decoders`

pkg/proto/configuration/blobstore/blobstore.proto
aron-muon: Yes excellent point. I moved `ZstdCompressionConfiguration` to `ApplicationConfiguration.zstd_compression`. A single pool is now created at startup in `main.go` and shared by all gRPC CAS clients and the ByteStream server. Per-backend config is just `bool enable_zstd_compression` to opt in. (Also added the reserved field comment)

EdSchouten:
Hey @aron-muon,

What if we add a pkg/zstd/pool.go with inside of that:

```go
type Decoder interface {
   Close() error
   // All of the other stuff that we use from klauspost.
   ....
}

type Encoder interface {
    Close() error
   // All of the other stuff that we use from klauspost.
   ....
}

type Pool interface {
    NewDecoder(ctx context.Context, r io.Reader) (Decoder, error)
    NewEncoder(ctx context.Context, w io.Writer) (Encoder, error)
}
```

Then we take that BoundedZstdPool implementation of yours, but place it in pkg/zstd/bounded_pool.go and rename it to NewBoundedPool() + boundedPool. Instead of ReleaseDecoder()/ReleaseEncoder(), we override {Decoder,Encoder}.Close() to do the right thing.

I saw that you patched up the bytestream server to conditionally call into the pool or not. With a change like this, this would no longer be necessary. If we wanted to continue to support non-pooled usage of zstd, we could just provide a different implementation of Pool.

I also saw that reference_expanding_blob_access.go also talks to klauspost directly. We should probably patch that up as well. There is also pkg/util/zstd_reader.go. Probably worth moving that to pkg/zstd/read_closer.go, and removing Zstd from the function/type names.

One question I do have about this approach though: by having a maximum decoder/encoder count in place, isn't there a risk that this change leads to deadlocks/lack of forward progress? One user inside a Starbucks with poor wifi could issue a bunch of Read()/Write() calls and cause encoders/decoders to get exhausted.

aron-muon:
> Hey @aron-muon,
> 
> What if we add a pkg/zstd/pool.go with inside of that:
> 
> ```go
> type Decoder interface {
>    Close() error
>    // All of the other stuff that we use from klauspost.
>    ....
> }
> 
> type Encoder interface {
>     Close() error
>    // All of the other stuff that we use from klauspost.
>    ....
> }
> 
> type Pool interface {
>     NewDecoder(ctx context.Context, r io.Reader) (Decoder, error)
>     NewEncoder(ctx context.Context, w io.Writer) (Encoder, error)
> }
> ```
> 
> Then we take that BoundedZstdPool implementation of yours, but place it in pkg/zstd/bounded_pool.go and rename it to NewBoundedPool() + boundedPool. Instead of ReleaseDecoder()/ReleaseEncoder(), we override {Decoder,Encoder}.Close() to do the right thing.
> 
> I saw that you patched up the bytestream server to conditionally call into the pool or not. With a change like this, this would no longer be necessary. If we wanted to continue to support non-pooled usage of zstd, we could just provide a different implementation of Pool.
> 
> I also saw that reference_expanding_blob_access.go also talks to klauspost directly. We should probably patch that up as well. There is also pkg/util/zstd_reader.go. Probably worth moving that to pkg/zstd/read_closer.go, and removing Zstd from the function/type names.
> 
> One question I do have about this approach though: by having a maximum decoder/encoder count in place, isn't there a risk that this change leads to deadlocks/lack of forward progress? One user inside a Starbucks with poor wifi could issue a bunch of Read()/Write() calls and cause encoders/decoders to get exhausted.

Yep that's a really clean choice, thank you for the comment!

- Created `pkg/zstd/` with `Pool`, `Encoder`, `Decoder` interfaces exactly as you described
- `BoundedPool` where `Close()` flushes and releases back to pool, no more separate `Release` methods
- `UnboundedPool` that creates fresh encoders/decoders per call, eliminating the conditional pool/non-pool paths in the ByteStream server
- Moved `pkg/util/zstd_reader.go` → `pkg/zstd/read_closer.go` (renamed to `NewReadCloser`)
- Patched `reference_expanding_blob_access.go` to use `bb_zstd.NewReadCloser()`

On the deadlock/starvation risk: you're right that a slow client holds a decoder for the full upload duration on the write path (since the decoder reads directly from the gRPC stream). Context deadlines prevent indefinite blocking on *acquire*, but don't help with slots already held by slow clients. The alternative (unbounded allocation) is what causes OOM under load, so bounded-with-backpressure is strictly better, but it needs to be observable and tunable. I'm thinking we expose Prometheus metrics on the pool:

- `buildbarn_zstd_pool_encoders_in_use` / `decoders_in_use` — are we near capacity?
- `acquisitions_total` / `rejections_total` — are requests getting dropped?
- `wait_duration_seconds` — how long are requests blocking for a slot?

This lets operators see the Starbucks scenario happening (in-use pinned at max, wait duration spiking) and tune pool size accordingly. Since the buildbarn tools are not opinionated by default, I think putting the onus on operators to tune the setup is appropriate. What do you think?

pkg/zstd/unbounded_pool.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &unboundedDecoder{Decoder: dec}, nil
+}
+
+// unboundedDecoder wraps *zstd.Decoder to implement the Decoder
+// interface. zstd.Decoder.Close() does not return an error, so this
+// adapter adds the error return.
+type unboundedDecoder struct {
```
EdSchouten, line 38: Instead of adding this type, can we please change Decoder.Close() to not return an error? I don't think the bounded pool has an implementation that can ever fail, right?

pkg/zstd/read_closer.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"io"
+
+	"github.com/klauspost/compress/zstd"
+)
+
+// NewReadCloser creates a new io.ReadCloser that wraps an underlying
+// reader and decompresses the data using Zstandard. The reader will
+// close both the decoder and the underlying reader when it is closed.
+func NewReadCloser(underlyingReader io.ReadCloser, options ...zstd.DOption) (io.ReadCloser, error) {
```
EdSchouten: I was thinking that this function could also be changed to take a Pool instead of the decoder options. So let ReferenceExpandingBlobAccess take a zstd.Pool as well.

pkg/zstd/bounded_pool.go
```diff
@@ -0,0 +1,126 @@
...
+			panic("failed to create ZSTD decoder: " + err.Error())
+		}
+		return dec
+	}
+
+	return p
+}
+
+func (p *boundedPool) NewEncoder(ctx context.Context, w io.Writer) (Encoder, error) {
+	if err := p.encoderSem.Acquire(ctx, 1); err != nil {
```
EdSchouten: Could we change this to use util.AcquireSemaphore()? Semaphore's Acquire() method is a bit silly in that it completely ignores `ctx` in the case the semaphore can be acquired. This means that it doesn't consistently act as a cancelation point.

pkg/blobstore/grpcclients/cas_blob_access.go
```diff
@@ -269,21 +299,31 @@ func (ba *casBlobAccess) Put(ctx context.Context, digest digest.Digest, b buffer
...
+			}
+			return status.Errorf(codes.ResourceExhausted, "Failed to acquire ZSTD encoder: %v", err)
 		}
 
-		if err := b.IntoWriter(zstdWriter); err != nil {
-			zstdWriter.Close()
-			byteStreamWriter.Close()
+		if err := b.IntoWriter(encoder); err != nil {
+			if zstdCloseErr := encoder.Close(); zstdCloseErr != nil {
+				err = errors.Join(err, zstdCloseErr)
```
EdSchouten: The convention in the Buildbarn tree is to use util.StatusFromMultiple(). errors.Join() may discard metadata that's attached to gRPC status objects.

That said: I don't think there's any real value in propagating Zstd closure errors. It likely only adds noise, as it's not the source of the failure.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -931,9 +931,48 @@ message GrpcBlobAccessConfiguration {
...
+  // improve compression ratio at the cost of memory per encoder.
+  // Total encoder memory budget is approximately
+  // maximum_encoders * encoder_window_size_bytes.
+  int32 encoder_window_size_bytes = 3;
+
+  // Maximum window size accepted by ZSTD decoders, in bytes. Streams
+  // compressed with a larger window than this will be rejected. Total
+  // decoder memory budget is approximately
+  // maximum_decoders * decoder_max_window_size_bytes.
+  int32 decoder_max_window_size_bytes = 4;
```
EdSchouten: s/max/maximum/ ? Or maybe omit it entirely for symmetry with the encoder window size?

cmd/bb_copy/main.go
```diff
@@ -43,7 +43,8 @@ func main() {
 
 		blobAccessCreator := blobstore_configuration.NewCASBlobAccessCreator(
 			grpcClientFactory,
-			int(configuration.MaximumMessageSizeBytes))
+			int(configuration.MaximumMessageSizeBytes),
+			nil)
```
EdSchouten: I suspect that this argument is no longer optional, right? We should at least pass in an unbounded zstd pool.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -931,9 +931,48 @@ message GrpcBlobAccessConfiguration {
...
-  // used if the server supports it. Has no effect for other storage
-  // types (AC, ICAS, etc.).
-  bool enable_compression = 2;
+  // Addressable Storage (CAS). Requires zstd_compression to be
+  // configured at the application level. Has no effect for other
+  // storage types (AC, ICAS, etc.).
+  bool enable_zstd_compression = 4;
+}
+
+message ZstdCompressionConfiguration {
```
EdSchouten: Now that we've created a pkg/zstd, we may argue that this has nothing to do with blobstore. What are your thoughts on creating a pkg/proto/configuration/zstd/zstd.proto and put this message in there under the name PoolConfiguration?

Then we can add a pkg/zstd/configuration.go or something that provides a NewPoolFromConfiguration(), similar to what we have in pkg/jwt/configuration.go. That function we can call from within cmd/bb_copy/main.go, etc. to create a pool similar to what we currently do in cmd/bb_storage/main.go.

EdSchouten:
> I'm thinking we expose Prometheus metrics on the pool:
> 
> * `buildbarn_zstd_pool_encoders_in_use` / `decoders_in_use` — are we near capacity?
> * `acquisitions_total` / `rejections_total` — are requests getting dropped?
> * `wait_duration_seconds` — how long are requests blocking for a slot?
> 
> This lets operators see the Starbucks scenario happening (in-use pinned at max, wait duration spiking) and tune pool size accordingly. Since the buildbarn tools are not opinionated by default, I think putting the onus on operators to tune the setup is appropriate. What do you think?

SGTM! Would it be possible to implement that in the form of a MetricsPool or something? So as a wrapper around Pool. Just like we already have for BlobAccess, pkg/eviction's Set, etc.

In terms of metrics: be sure to use counters over gauges. So encoders_in_use is something I'd avoid. Just add counters for the number of acquisitions and releases. You can always subtract those two to get the number in use.

pkg/zstd/unbounded_pool.go
aron-muon: Decoder.Close() no error:** Done, `Decoder.Close()` now returns nothing. Removed the `unboundedDecoder` wrapper entirely since `*zstd.Decoder.Close()` already has no error return.

pkg/zstd/read_closer.go
aron-muon: Done, `NewReadCloser(pool, underlyingReader)` now takes a `Pool` instead of raw decoder options. `ReferenceExpandingBlobAccess` now takes a `zstd.Pool` parameter.

pkg/zstd/bounded_pool.go
aron-muon: Nice - changed to `util.AcquireSemaphore`

pkg/blobstore/grpcclients/cas_blob_access.go
aron-muon: Removed `errors.Join` entirely, since they just add noise as you noted.

pkg/proto/configuration/blobstore/blobstore.proto
aron-muon: Renamed to `decoder_window_size_bytes` for symmetry with `encoder_window_size_bytes`

cmd/bb_copy/main.go
aron-muon: Yep, nice catch- all callers now pass `NewPoolFromConfiguration(nil)` which returns an unbounded pool with default options

pkg/proto/configuration/blobstore/blobstore.proto
aron-muon: Yes that checks out! created `pkg/proto/configuration/zstd/zstd.proto` with `PoolConfiguration`, and `pkg/zstd/configuration.go` with `NewPoolFromConfiguration()`. `bb_storage`, `bb_copy`, `bb_replicator`, and `NewCASAndACBlobAccessFromConfiguration` all use it now.

aron-muon:
> SGTM! Would it be possible to implement that in the form of a MetricsPool or something? So as a wrapper around Pool. Just like we already have for BlobAccess, pkg/eviction's Set, etc.
> 
> In terms of metrics: be sure to use counters over gauges. So encoders_in_use is something I'd avoid. Just add counters for the number of acquisitions and releases. You can always subtract those two to get the number in use.

I've implemented `MetricsPool` as a wrapper around `Pool`, same pattern as `MetricsBlobAccess` and eviction's `MetricsSet`. Uses counters as you suggested:
- `buildbarn_zstd_pool_operations_total` with labels for `EncoderAcquire`, `EncoderRelease`, `EncoderReject`, `DecoderAcquire`, `DecoderRelease`, `DecoderReject`
- `buildbarn_zstd_pool_wait_duration_seconds` histogram for acquire latency

All pools get metrics automatically!

pkg/blobstore/configuration/new_blob_access.go
EdSchouten: I think this should also be changed to take a `zstdPool`. Ideally we should only have calls to `NewPoolFromConfiguration()` inside `cmd/*/main.go`.

pkg/blobstore/grpcclients/BUILD.bazel
```diff
@@ -33,14 +33,17 @@ go_library(
 
 go_test(
     name = "grpcclients_test",
-    srcs = ["cas_blob_access_test.go"],
+    srcs = [
+        "cas_blob_access_test.go",
+    ],
```
EdSchouten: Nit: Can you revert this whitespace only change? Just to make sure people don't run into merge conflicts.

pkg/blobstore/grpcclients/cas_blob_access.go
```diff
@@ -74,47 +77,65 @@ func (r *byteStreamChunkReader) Close() {
...
 	wg            sync.WaitGroup
+	initOnce      sync.Once
+	initErr       error
 }
 
-func (r *zstdByteStreamChunkReader) Read() ([]byte, error) {
-	if r.zstdReader == nil {
-		pr, pw := io.Pipe()
+func (r *zstdByteStreamChunkReader) init(ctx context.Context) error {
+	r.initOnce.Do(func() {
```
EdSchouten: Is it really necessary to be this lazy here? Can't we just call NewDecoder() inside of `casBlobAccess.Get()`?

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -60,12 +61,13 @@ func (s *byteStreamServer) Read(in *bytestream.ReadRequest, out bytestream.ByteS
 
 	case remoteexecution.Compressor_ZSTD:
 		b := s.blobAccess.Get(out.Context(), digest)
```
EdSchouten: Could you add the following line above this one, and use `ctx` everywhere? Thanks!
```go
ctx := out.Context()
```

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -214,23 +216,34 @@ func (s *byteStreamServer) writeZstd(stream bytestream.ByteStream_WriteServer, r
 		pendingData: request.Data,
 	}
 
-	zstdReader, err := util.NewZstdReadCloser(streamReader, zstd.WithDecoderConcurrency(1))
+	decoder, err := s.zstdPool.NewDecoder(stream.Context(), streamReader)
```
EdSchouten: Same here: don't call `stream.Context()` repeatedly.

Also, we should be calling `zstd.NewReadCloser()` here. That way we can get rid of `decoderReadCloser` below.

pkg/zstd/read_closer.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+	"io"
+)
+
+// NewReadCloser creates a new io.ReadCloser that decompresses data
+// using a decoder from the provided Pool. Closing the returned reader
+// releases the decoder back to the pool and closes the underlying
+// reader.
+func NewReadCloser(pool Pool, underlyingReader io.ReadCloser) (io.ReadCloser, error) {
```
EdSchouten: This should be extended to take a `ctx`.

pkg/proto/configuration/bb_storage/bb_storage.proto
```diff
@@ -102,6 +103,13 @@ message ApplicationConfiguration {
   // Support for IDENTITY (i.e., no compression) is implied.
   repeated build.bazel.remote.execution.v2.Compressor.Value
       supported_compressors = 20;
+
+  // ZSTD compression pool configuration. When set, creates a process-
+  // wide pool of ZSTD encoders and decoders shared by all gRPC CAS
+  // clients (those with enable_zstd_compression set) and the ByteStream
+  // server. This ensures bounded, predictable memory usage regardless
+  // of the number of backends or concurrent operations.
+  buildbarn.configuration.zstd.PoolConfiguration zstd_compression = 21;
```
EdSchouten: Maybe call this `zstd_pool`? Just to make it explicit that this does not enable/disable Zstandard compression. It only sets up a pool.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -931,9 +931,13 @@ message GrpcBlobAccessConfiguration {
...
+
   // Enable ZSTD compression for ByteStream operations in the Content
-  // Addressable Storage (CAS). When enabled, ZSTD compression will be
-  // used if the server supports it. Has no effect for other storage
-  // types (AC, ICAS, etc.).
-  bool enable_compression = 2;
+  // Addressable Storage (CAS). Requires zstd_compression to be
+  // configured at the application level. Has no effect for other
+  // storage types (AC, ICAS, etc.).
+  bool enable_zstd_compression = 4;
```
EdSchouten: I guess we can now revert changes to this file altogether?

pkg/blobstore/grpcclients/cas_blob_access.go
```diff
@@ -40,15 +41,17 @@ type casBlobAccess struct {
 // Addressable Storage.
 //
 // If enableZSTDCompression is true, the client will use ZSTD compression
-// for ByteStream operations if the server supports it.
-func NewCASBlobAccess(client grpc.ClientConnInterface, uuidGenerator util.UUIDGenerator, readChunkSize int, enableZSTDCompression bool) blobstore.BlobAccess {
+// for ByteStream operations if the server supports it. In that case,
+// zstdPool must be provided for pooling encoders/decoders.
+func NewCASBlobAccess(client grpc.ClientConnInterface, uuidGenerator util.UUIDGenerator, readChunkSize int, enableZSTDCompression bool, zstdPool bb_zstd.Pool) blobstore.BlobAccess {
```
EdSchouten: It may be cleaner if we remove `enableZSTDCompression` entirely, and use `zstdPool != nil` to indicate whether compression should be used?

pkg/zstd/metrics_pool.go
```diff
@@ -0,0 +1,125 @@
...
+	poolMetricsPrometheusMetrics sync.Once
+
+	poolOperationsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "buildbarn",
+			Subsystem: "zstd",
+			Name:      "pool_operations_total",
+			Help:      "Total number of encoder/decoder pool operations.",
+		},
+		[]string{"name", "operation"})
```
EdSchouten: With metrics it's always an interesting discussion whether to use separate metrics, or have a single metric with labels. When making these tradeoffs, it's always work asking: "does it make sense to aggregate these?" For decoders and encoders it may make sense to do such aggregation. "I just want to know how many times we use Zstandard." However, for acquisitions/releases it sort of doesn't. You want to subtract those from each other.

Maybe it's better to have pool_acquisitions_total, pool_releases_total, etc. But then have an object_type="Encoder"/"Decoder" label?

EdSchouten (commented):
This is starting to look pretty good, @aron-muon. Thanks for your perseverance!

pkg/blobstore/configuration/new_blob_access.go
aron-muon: It takes a `zstdPool` parameter now, `NewPoolFromConfiguration()` calls only live in `cmd/*/main.go`.

pkg/blobstore/grpcclients/BUILD.bazel
aron-muon: Done

pkg/blobstore/grpcclients/cas_blob_access.go
aron-muon: Done, `NewDecoder()` is now called eagerly in `casBlobAccess.Get()`. Removed `initOnce`, `initErr`, and the `init()` method entirely.

pkg/blobstore/grpcservers/byte_stream_server.go
aron-muon: All set, `ctx := out.Context()` extracted above the switch.

pkg/blobstore/grpcservers/byte_stream_server.go
aron-muon: Extracted `ctx := stream.Context()` and switched to `zstd.NewReadCloser()`, which eliminated the `decoderReadCloser` type.

pkg/zstd/read_closer.go
aron-muon: Done

pkg/proto/configuration/bb_storage/bb_storage.proto
aron-muon: Done

pkg/blobstore/grpcclients/cas_blob_access.go
aron-muon: Nice, `NewCASBlobAccess` no longer takes a bool. `zstdPool != nil` controls whether compression is used.

pkg/zstd/metrics_pool.go
aron-muon: Perfect - separate metrics `pool_acquisitions_total`, `pool_releases_total`, `pool_rejections_total`, `pool_wait_duration_seconds`, each with `object_type="Encoder"/"Decoder"` label. (Done)

pkg/proto/configuration/blobstore/blobstore.proto
aron-muon: Reverted to upstream. The per-backend `enable_zstd_compression` field is gone entirely since compression is now controlled by whether a pool is provided

pkg/blobstore/configuration/cas_blob_access_creator.go
```diff
@@ -96,7 +99,7 @@ func (bac *casBlobAccessCreator) NewCustomBlobAccess(terminationGroup program.Gr
 		// TODO: Should we provide a configuration option, so
 		// that digest.KeyWithoutInstance can be used?
 		return BlobAccessInfo{
```
EdSchouten, line 105: Doesn't this need something along these lines?

```go
var zstdPool *zstd.Pool
if configuration.EnableCompression {
    zstdPool = bac.zstdPool
}
```

aron-muon: Ahh good catch. Updated - the pool is now only passed to NewCASBlobAccess when backend.Grpc.EnableCompression is true, so per-backend opt-in is preserved.

EdSchouten (approved):
I think this looks great, @aron-muon. Thanks for taking the time to make these refactoring changes. I've just kicked off CI. Assuming CI is happy, I will merge it tomorrow morning.

In the meantime, can you please roll out this change on your end and double check that it doesn't cause any obvious regressions for things not covered by tests? Thanks!

aron-muon:
> I think this looks great, @aron-muon. Thanks for taking the time to make these refactoring changes. I've just kicked off CI. Assuming CI is happy, I will merge it tomorrow morning.
> 
> In the meantime, can you please roll out this change on your end and double check that it doesn't cause any obvious regressions for things not covered by tests? Thanks!

CI was failing lint - I fixed that in the latest commit. I'll launch this on our side and will report back on whether we find any issues!

ivangolub-oai:
@aron-muon thank you for this implementation! Do you have any recommendations what are good numbers to pick for maximumEncoders/maximumDecoders? (ps: asked in https://join.slack.com/t/buildteamworld/shared_invite/zt-3epswjmkj-bsCAPUtTmXEtlPcBl5C_lQ but looks like you are no there right now)

PR #329, opened 2026-03-09
Improve windows path parsing

tomgr:
This PR is necessary to fix some bugs with the WinFSP vfs integration in bb-remote-execution (see https://github.com/buildbarn/bb-remote-execution/pull/213).

Firstly it changes the pretty printing of windows paths to append `\.` as a bare trailing backslash is not a valid pathname component.

Secondly this neatens up the parsing of Windows file paths and adds handling for paths starting with `\\.\`.

Claude helped write some of this -- particularly the tests (it has more patience than I do at converting escape sequences!).

### DIFF
--- a/pkg/filesystem/path/builder.go
+++ b/pkg/filesystem/path/builder.go
@@ -109,14 +109,23 @@ func (b *Builder) GetWindowsString(format WindowsPathFormat) (string, error) {
 		prefix = "\\"
 	}
 
-	// Emit trailing slash in case the path refers to a directory,
-	// or a dot or slash if the path is empty. The suffix is been
-	// constructed by platform-independent code that uses forward
-	// slashes. To construct a Windows path we must use a
+	// If the path refers to a directory we should emit a trailing slash,
+	// and since the suffix has been constructed by platform-independent code
+	// that uses forward slashes we need to convert the path to use a
 	// backslash.
+	//
+	// Unfortunately, a trailing backslash is not a valid pathname component:
+	// https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/ffb795f3-027d-4a3c-997d-3085f2332f6f
+	// Instead we emit `\.` and rely on win32 path normalization. However,
+	// for WindowsPathFormatDevicePath we cannot do this as device paths
+	// bypass win32 path normalization so have to just emit a backlash.
 	suffix := b.suffix
 	if suffix == "/" {
-		suffix = "\\"
+		if format == WindowsPathFormatDevicePath {
+			suffix = "\\"
+		} else {
+			suffix = "\\."
+		}
 	}
 	out.WriteString(suffix)
 	return out.String(), nil
--- a/pkg/filesystem/path/builder_test.go
+++ b/pkg/filesystem/path/builder_test.go
@@ -63,14 +63,14 @@ func TestBuilder(t *testing.T) {
 		for _, data := range [][]string{
 			{".", "."},
 			{"..", ".."},
-			{"/", "\\"},
+			{"/", "\\."},
 			{"hello", "hello"},
-			{"hello/", "hello\\"},
+			{"hello/", "hello\\."},
 			{"hello/..", "hello\\.."},
-			{"/hello/", "\\hello\\"},
+			{"/hello/", "\\hello\\."},
 			{"/hello/..", "\\hello\\.."},
 			{"/hello/../world", "\\hello\\..\\world"},
-			{"/hello/../world/", "\\hello\\..\\world\\"},
+			{"/hello/../world/", "\\hello\\..\\world\\."},
 			{"/hello/../world/foo", "\\hello\\..\\world\\foo"},
 		} {
 			p := data[0]
@@ -92,16 +92,16 @@ func TestBuilder(t *testing.T) {
 
 	t.Run("WindowsIdentity", func(t *testing.T) {
 		for _, p := range []string{
-			"C:\\",
-			"C:\\hello\\",
+			"C:\\.",
+			"C:\\hello\\.",
 			"C:\\hello\\..",
 			"C:\\hello\\..\\world",
-			"C:\\hello\\..\\world\\",
+			"C:\\hello\\..\\world\\.",
 			"C:\\hello\\..\\world\\foo",
 			"C:\\hello\\..\\world\\foo",
-			"\\\\server\\share\\hello\\",
+			"\\\\server\\share\\hello\\.",
 			"\\\\server\\share\\hello\\..\\world",
-			"\\\\server\\share\\hello\\..\\world\\",
+			"\\\\server\\share\\hello\\..\\world\\.",
 			"\\\\server\\share\\hello\\..\\world\\foo",
 		} {
 			t.Run(p, func(t *testing.T) {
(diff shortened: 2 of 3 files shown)


pkg/filesystem/path/scope_walker.go
```diff
@@ -7,27 +7,59 @@ type ScopeWalker interface {
...
+	// VirtualRootScopeWalkerFactory may only call it when the path
+	// is known to be valid. Absence of calls to OnAbsolute(),
+	// OnRelative() or OnWindowsRoot() are used to indicate that the
+	// provided path does not resolve to a location inside the file
 	// system.
 	OnAbsolute() (ComponentWalker, error)
 	OnRelative() (ComponentWalker, error)
-	OnDriveLetter(drive rune) (ComponentWalker, error)
-	OnShare(server, share string) (ComponentWalker, error)
+	OnWindowsRoot(root WindowsRootKind) (ComponentWalker, error)
```
EdSchouten: So I get you're trying to solve the issue that path.Builder isn't always emitting paths in the right format. But to me it's unclear why this specific API change is needed. Could you please explain a bit more why this part is necessary?

pkg/filesystem/path/stringer.go
```diff
@@ -4,12 +4,37 @@ package path
...
 	// Note that not all paths can be printed like this (e.g. relative paths
 	// cannot), so this will fallback to WindowsPathFormatStandard.
 	WindowsPathFormatDevicePath
+	// WindowsPathFormatNoTrailingSeparator is like
+	// WindowsPathFormatStandard, but suppresses the trailing "\"
+	// that is normally emitted when the path refers to a
+	// directory. This is needed when constructing symlink
+	// substitute names for Windows reparse points, where a
+	// trailing backslash can produce consecutive backslashes
+	// after path substitution.
```
EdSchouten: I just want to get clarified: is this a limitation/feature of Windows, or is it really just WinFSP that freaks out in case symlink targets have trailing slashes? What happens if you create a symlink on NTFS that has a trailing slash?

In the PR description I see you mention that the issue is in FspFileSystemResolveReparsePointsInternal. Maybe instead of fixing this on the Buildbarn side, we should report an issue against WinFSP that they fix this?

EdSchouten: Alternatively, what if instead of a trailing slash we append \\. ? Is that valid?

tomgr: I think the Windows standards require symlink reparse buffers not to have a trailing slash. [The symbolic link reparse data buffer docs](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/b41f1cbf-10df-4a47-98d4-1c52a833d913) state that the substitute name has to be a [pathname](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/ffb795f3-027d-4a3c-997d-3085f2332f6f). However, the pathname documentation does not allow an empty final component: each component of the path has to have at least one character.

`\.` is valid; WinFSP would form `\.\` after substitution which does seem to get handled correctly. However I thought that wouldn't be as elegant as that would mean that whenever directory paths were printed with WindowsPathFormatStandard (which is the default) they would have a trailing `\.\`. There's also a minor concern which is that Windows has a fairly low limit on path sizes, so keeping them short is always a good idea.

pkg/filesystem/path/scope_walker.go
tomgr: I could have kept the separate OnDriveLetter and OnShare calls, but both of them would have needed a `prefix WindowsPrefixKind` parameter added as we need to record the exact prefix that was used. It seemed a bit nicer to me to have a single Windows-specific callback as I noticed that quite a few of the walkers were just passing through the calls (e.g. voidScopeWalker and absoluteScopeWalker as examples). I thought that might allow more code to ignore Windows when it doesn't care about it.

I realise this will create a bit of churn in the other buildbarn repos but I'm happy to handle that straight after this is merged. And in either case we'd need to make modifications as we'd need the extra prefix parameter to be added.

pkg/filesystem/path/stringer.go
EdSchouten: I would still have a preference for appending `\.` instead of removing the suffix entirely. One intentional design goal of the pathname library was that it does not relax paths in any way.

That MS-FSCC link that you shared is interesting. It raises the question: is it ever valid for us to generate Windows pathnames that end with slashes? If not, maybe we should just make a change like this?

```diff
diff --git pkg/filesystem/path/builder.go pkg/filesystem/path/builder.go
index 80be90a..a4ff01d 100644
--- pkg/filesystem/path/builder.go
+++ pkg/filesystem/path/builder.go
@@ -116,7 +116,7 @@ func (b *Builder) GetWindowsString(format WindowsPathFormat) (string, error) {
        // backslash.
        suffix := b.suffix
        if suffix == "/" {
-               suffix = "\\"
+               suffix = "\\."
        }
        out.WriteString(suffix)
        return out.String(), nil
```

tomgr: Ok I've done this. I think it should be ok given the definition of pathname. And the winfsp tests all pass in bb-remote-execution with this.

pkg/filesystem/path/scope_walker.go
EdSchouten: Gotcha!

One of the things I want to prevent is that we end up with an API design that's somewhere in between. If the goal is that we're able to easily forward/describe/store the scope of a path, then we should in my opinion add a dedicated `Scope` type, and simplify this interface to just `OnScope(Scope) (ComponentWalker, error)`. Or there is no need to do that, and we continue to make this interface fatter. It would in my opinion be a bit odd if for "Windows roots" there is a mechanism to do it, but not for absolute/relative paths.

Could you also please remind me why we need to preserve the namespace prefix in literal form? Take extended length prefix `\\?\`. My guess would be that there's no need for us to actually parse/preserve that. We could just jump over it while parsing. But when `GetWindowsString()` is called and the resulting string is longer than n bytes, then we could/should prepend it. Or we always just prepend it in contexts where it's allowed to do so.

You can concatenate many paths that don't have the extended length prefix. But the resulting path can be long enough that the use of such a prefix becomes required.

tomgr: The problem is the namespace prefix affects exactly how the path is interpreted by Windows: `\??\` and`\\?\` are not normalised by Windows, so slashes are left in whatever form they were originally in for example. In contrast, no prefix or `\\.\` would mean that normalisation would still be applied.

What makes this a bit confusing is that the builder is doing normalisation. Arguably we should just be preserving `\??\` and`\\?\` paths  verbatim and we should not even be transforming slash directions. This doesn't fit that well with the path library though, particularly the builder which is designed to normalise.

The original motivation for this change is bb-remote-execution's WinFSP layer. I now wonder if my fix to that is in the wrong place. I think the underlying issue with bb-remote-execution is that REv2 SymlinkNode target field is documented as using `/` as the path separator. bb-remote-execution preserves that when creating the Symlinks but that's arguably incorrect on Windows. It mostly works because Windows mostly permits either kind of slash, except when resolving symlinks.

I think a better fix is to change bb-remote-execution so that symlinks are created from the SymlinkNode with `\` instead. Then the WinFSP reparse buffer construction can just take the raw bytes returned from VirtualReadlink.

As for where this leaves this PR - I think there are a few useful cleanups that could be extracted but there is no longer a need to preserve the prefix. Unless you think otherwise I'll implement the above fix for bb-remote-execution and I'll change this PR just to include some minor improvements to the windows path parsing code. Does that direction sound good to you?

EdSchouten: > I think the underlying issue with bb-remote-execution is that REv2 SymlinkNode target field is documented as using / as the path separator. 

Indeed. SymlinkNode.target, Command.output_paths, Command.working_directory, etc. are all intended to be UNIX-style paths, regardless of the pathname convention of the operating system on which the action is going to run. This is intentional, as otherwise it's hard to build tools for inspecting actions.

So there is no proper/standard way of putting symlinks in input roots that refer to other drives, network shares, etc.

tomgr: I've changed this PR to only have what's needed for the bb-remote-execution fixes.

pkg/filesystem/path/builder.go
```diff
@@ -116,7 +116,14 @@ func (b *Builder) GetWindowsString(format WindowsPathFormat) (string, error) {
 	// backslash.
 	suffix := b.suffix
 	if suffix == "/" {
-		suffix = "\\"
+		// We append "\." because a bare trailing backslash is not a valid pathname component:
+		// https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/ffb795f3-027d-4a3c-997d-3085f2332f6f
+		if format == WindowsPathFormatDevicePath {
+			// But device paths bypass win32 path normalization, so we can't use `\\.`
```
EdSchouten: Can you please remove these two comments, and update/augment the larger comment above it to fully describe why the code is structured the way it is?

pkg/filesystem/path/builder.go
```diff
@@ -116,7 +116,14 @@ func (b *Builder) GetWindowsString(format WindowsPathFormat) (string, error) {
...
+			suffix = "\\"
+		} else {
+			suffix = "\\."
```
EdSchouten, line 127: So something I do not understand is the following: even if we fix this specific issue, we may still emit paths containing ".." components, right? And if my interpretation of https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/ffb795f3-027d-4a3c-997d-3085f2332f6f is correct, this is also not allowed. How should we deal with that?

tomgr: I think that's allowed. Each pathname component can be a dot directory name:

> dot directory name: In a pathname, a directory name component of "." (single period) or ".." (two periods). For more details, see [[MS-FSCC]](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/efbfe127-73ad-4140-9967-ec6500e66d5e) section 2.1.5.1.

EdSchouten: d'oh! You're correct. I guess I glossed over that, because on UNIX those are explicitly called "dotdots". They aren't grouped together with "dots".

pkg/filesystem/path/builder.go
tomgr: sure done!

EdSchouten (approved):

PR #330, opened 2026-03-12
fix: prevent double-close inflating zstd pool release counters

aron-muon:
## Summary

Hey @EdSchouten, firstly I want to say that overall in our testing, the changes from #327 have been working very well! A tremendous thank you for taking your time to work with me and refine this approach into a very nice technical solution 🥳

image

After integrating with the new metrics available, I've noticed that the observed releases did not match the observed acquisitions. This PR introduces a patch fix for this issue.

- Remove redundant `defer zstdReader.Close()` in `writeZstd()`, the buffer takes ownership of the reader via `NewCASBufferFromReader` and closes it, so the defer was a second close
- Add double-close guards to `metricsEncoder.Close()` and `metricsDecoder.Close()` as a safety net, matching the existing pattern in `pooledEncoder`/`pooledDecoder`

The `writeZstd()` path double-closes the decoder: once when the buffer consumes and closes the reader, and again via `defer`. Since `metricsDecoder.Close()` lacks double-close protection (unlike the underlying `pooledDecoder`), each compressed ByteStream write increments `releases_total` an extra time. Over time this causes `releases_total` to permanently exceed `acquisitions_total` (observed: 336,865 releases vs 333,001 acquisitions = 3,864 excess, corresponding to the number of compressed ByteStream writes).

## Test plan

- [x] `bazel test //pkg/zstd/... //pkg/blobstore/grpcservers/...` — all pass
- [x] `bazel test //...` — all 30 tests pass
- [ ] Verify `releases_total` no longer exceeds `acquisitions_total` after deployment

### DIFF
--- a/pkg/blobstore/grpcservers/byte_stream_server.go
+++ b/pkg/blobstore/grpcservers/byte_stream_server.go
@@ -222,7 +222,6 @@ func (s *byteStreamServer) writeZstd(stream bytestream.ByteStream_WriteServer, r
 	if err != nil {
 		return status.Errorf(codes.ResourceExhausted, "Failed to acquire ZSTD decoder: %v", err)
 	}
-	defer zstdReader.Close()
 
 	if err := s.blobAccess.Put(
 		ctx,
--- a/pkg/blobstore/grpcservers/byte_stream_server.go
+++ b/pkg/blobstore/grpcservers/byte_stream_server.go
@@ -9,6 +9,7 @@ import (
 	"github.com/buildbarn/bb-storage/pkg/blobstore"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
 	"github.com/buildbarn/bb-storage/pkg/digest"
+	"github.com/buildbarn/bb-storage/pkg/util"
 	bb_zstd "github.com/buildbarn/bb-storage/pkg/zstd"
 
 	"google.golang.org/genproto/googleapis/bytestream"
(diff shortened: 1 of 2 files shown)


aron-muon:
Visualizing the issue further:
image

pkg/zstd/metrics_pool.go
```diff
@@ -123,9 +123,14 @@ func (p *metricsPool) NewDecoder(ctx context.Context, r io.Reader) (Decoder, err
 type metricsEncoder struct {
 	Encoder
 	releases prometheus.Counter
+	closed   bool
 }
 
 func (e *metricsEncoder) Close() error {
+	if e.closed {
```
EdSchouten: Instead of doing this, could we set `e.Encoder = nil` instead? That way we can be 100% certain we don't attempt to call any other methods against the object after it's been closed. That also prevents the need for having a separate boolean field.

aron-muon: Yeah, makes sense - updated

EdSchouten (approved):

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -222,7 +222,6 @@ func (s *byteStreamServer) writeZstd(stream bytestream.ByteStream_WriteServer, r
 	if err != nil {
 		return status.Errorf(codes.ResourceExhausted, "Failed to acquire ZSTD decoder: %v", err)
```
EdSchouten: Sorry for not catching this in an earlier review, but could you please change this to:
```go
return util.StatusWrap(err, "Failed to acquire ZSTD decoder")
```
This ensures that the gRPC code coming from `Pool.NewDecoder()` is returned. So we'll get `codes.Canceled` properly in case of a client cancelation. It also means that gRPC error metadata (if any) doesn't get discarded. Thanks!

EdSchouten (changes requested):

pkg/blobstore/grpcservers/byte_stream_server.go
aron-muon: Fixed in latest pushed!

EdSchouten (approved):

PR #332, opened 2026-03-24
Forward termination signals to child process

oscar-meroton:
Commit d2a3a16763ee8a62125d9c57b7a12bf5d9ae0789,
"On Linux, relaunch the current process when running as PID 1 (#311)", relaunches the command in a child process and then waits for it to exit, if the parent process has PID 1. If the command is ran in a container, any termination signal sent to the container is forwarded to the parent process. This unfortunately prevents the persistent state files from being saved to disk if the container is killed, as the child process never receives the shutdown signal.
This commit forwards `SIGINT` and `SIGTERM` signals to the child process, which then saves the state files to disk, exits, and triggers a graceful exit in PID 1.

### DIFF
--- a/pkg/program/run_main_linux.go
+++ b/pkg/program/run_main_linux.go
@@ -38,6 +38,18 @@ func relaunchIfPID1(currentPID int) {
 			log.Fatal("Failed to relaunch current process: ", err)
 		}
 
+		// Forward incoming termination signals to child process.
+		signalChan := make(chan os.Signal, 1)
+		signal.Notify(signalChan, terminationSignals...)
+		go func() {
+			for {
+				receivedSignal := <-signalChan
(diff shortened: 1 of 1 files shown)


oscar-meroton:
@EdSchouten,
1. Do you have any thoughts on whether to forward *any* signal, or just the two listed in `terminationSignals`?
2. Currently the `error` from `syscall.Kill` is not handled, should the program behave differently if an error is returned?

EdSchouten (approved):

EdSchouten:
1. Sticking to the termination signals is all right.
2. What you have in terms of error handling is also all right!

\o/

oscar-meroton:
@EdSchouten Great! Is there anything else or can we go ahead and merge?

EdSchouten:
Oh sorry. Was waiting for CI to complete, and forgot to merge.

PR #333, opened 2026-04-10
Tune IO performance

prime000000:
This patch adds a couple of tunable, which I found useful in our buildbarn setup.

We have 10 bare metal hosts managed by k8s. Each host has two 8TiB NVME SSD linked in software mirror raid. These SSDs support write speed of about 2.5Gb/s. We deploy buildbarn with memory limit set to 64Gi (rest is used for CI) and locate 1TiB of disk space for it, sharded over 10 files.

1. I think it always make sense to call fallocate, when creating the /blocks file. ext4 filesystem on linux default to lazy allocation, and if we create 100Tb file, by default 0 blocks are allocated upfront. Instead, allocation code is running during Pwrite, or even during writeback. Causing all sort of problems with fragmentation and lock contention inside the kernel.

2. I do not think mmap is always useful for file read, when file does not fit into memory (which is our case). Pread should give more predictable performance, since OS sees full range of requested data in a single syscall. Instead of second guessing how much data it needs to prefetch from the page-fault pattern.

3. It makes sense to sync data to disk after Pwrite. NVME is fast enough, so there is no degradation in performance. But disk write is issued synchronously from buildbarn thread. Without this, dirty pages are piled up in memory, and write to disk is started at random point in time in the future, when process tries allocating new memory, hits container memory limit and starts kernel memory reclaim.

With all three options applied I was able to get full underlying NVME write speed in synthetic benchmarks with consistent timings.

Without these options, buildbarn died under our CI load, when we tries switching to buildbarn from our current bazel-remote setup.

I did not have enough time to do a clear before/after experiment in our CI setup, but all my experience of working with storages tells me that these changes are worthwhile.


pkg/proto/configuration/blockdevice/blockdevice.proto
```diff
@@ -11,6 +11,15 @@ message FileConfiguration {
...
   int64 size_bytes = 2;
+
+  // When set to true (or left unset), use fallocate() to preallocate
+  // disk space for the file at creation time. This ensures that no
+  // out-of-space errors occur later during writes, and may reduce file
+  // fragmentation.
+  //
+  // This option is only supported on Linux and FreeBSD. On other platforms
+  // it is silently ignored.
+  optional bool preallocate = 3;
```
meroton-benjamin, line 22: The typical usage of using a file for block device is for development purposes, in those circumstances it is generally not desirable to preallocate the file. The way this flag is named also makes its unset behavior somewhat unpredictable (not setting preallocate implies preallocate: true).

pkg/proto/configuration/blockdevice/blockdevice.proto
```diff
@@ -39,4 +48,20 @@ message Configuration {
   // this process to crash, due to the Go runtime being unable to create
   // any more threads.
   int64 write_concurrency_limit = 3;
+
+  // When set to true (or left unset), use a memory map (mmap) for
+  // reads. When set to false, use pread() system calls instead. Using
+  // pread avoids page fault overhead and provides more predictable I/O
+  // behavior.
+  optional bool mmap = 4;
```
meroton-benjamin, line 56: Similarly to the above naming issue this configuration treats mmap unset as mmap: true.

We don't have any data here but it looks like an interesting avenue, in a real world workload one of the strategies would be faster overall (paying for a syscall for every pread vs pay for page misses when mmap heuristic fails). For very large reads the syscall overhead would be negligeable and the kernel will be able to make more informed choices about which pages to page in with a pread than with its mmap heuristics. But for smaller reads the syscall overhead may be very large in comparison.

I think it would be preferable if we figured out which strategy that was better and simply used that one rather than expecting cluster operators to make that choice.

pkg/proto/configuration/blockdevice/blockdevice.proto
```diff
@@ -39,4 +48,20 @@ message Configuration {
...
+
+  // When set to true, call sync_file_range() after each pwrite() to
+  // initiate writeback of the written data range. This reduces the
+  // amount of dirty data that accumulates in the page cache and can
+  // prevent latency spikes caused by the kernel flushing large amounts
+  // of dirty data at once.
+  //
+  // This option is only supported on Linux. On other platforms it is
+  // silently ignored.
+  bool sync_after_write = 5;
```
meroton-benjamin, line 66: This sounds like a dubious feature, sure when writing a decently large blob syncing it immediately makes sense but for small writes you would have massive write amplification as the kernel can only flush entire pages. Additionally nvme:s can only perform erasure on ~2MB blocks and have to internally accumulate these tiny 4KB chunks with almost 0 delta until the nvme controller decides to garbage collect them.

I'm not sure an explicit sync_after_write strategy for any sized blobs will outperform the kernel in this context.

pkg/blockdevice/new_block_device_from_file_unix.go
```diff
@@ -41,7 +41,14 @@ func NewBlockDeviceFromFile(path string, minimumSizeBytes int, zeroInitialize bo
 		return nil, 0, 0, util.StatusWrapf(err, "Failed to truncate file %#v to %d bytes", path, sizeBytes)
 	}
 
-	bd, err := newMemoryMappedBlockDevice(fd, int(sizeBytes))
+	if preallocate {
```
meroton-benjamin, line 44: This code has semantically different behavior on failure depending on operating system. In a linux filesystem which doesn't support fallocate or which fails to fallocate it returns an error, in a macos system it simply continues.

pkg/blockdevice/memory_mapped_block_device_unix.go
```diff
@@ -15,54 +15,77 @@ import (
...
+		}
+		bd.data = data
+	}
+	return bd, nil
 }
 
-func (bd *memoryMappedBlockDevice) ReadAt(p []byte, off int64) (n int, err error) {
-	// Let read actions go through the memory map to prevent system
-	// call overhead for commonly requested objects.
+func (bd *fdBlockDevice) ReadAt(p []byte, off int64) (n int, err error) {
```
meroton-benjamin, line 45: this is an if else over two different functions where the implementation is chosen at constructor time, this strongly indicates that this should be two different concrete types

meroton-benjamin (commented):
Hi and welcome to Buildbarn thanks for your contribution, I've left some comments on the PR for you to look at. Feel free to ask any followup questions or join us in slack to discuss it.

prime000000:
@meroton-benjamin thank you for the feedback. It seems that I need to bring actual benchmarks, instead of hand-waving. I'll be back with real numbers.

pkg/proto/configuration/blockdevice/blockdevice.proto
prime000000: > The typical usage of using a file for block device is for development purposes

Seems counter intuitive to me. Allocating and managing block device is much harder, compared to just creating a file. I bet most deployments out there are using files.

pkg/proto/configuration/blockdevice/blockdevice.proto
prime000000: Do you believe it is typical to have CAS full of very small files?

We have C++ codebase, I was operating under the assumption that CAS is mostly caching object files (~100Kb) and resulting binaries (~100-1000Mb).

prime000000: AC might store small protobufs, but I assumed that load from AC is negligible anyway.

meroton-benjamin: Besides the output files themselves the CAS also has to store a very large number of tiny artifacts such as directory objects, input files etc. These can often be just a few hundred bytes. A quick sampling on one of our clusters show that ~1/3 of all get requests that hit the storage layer are for blobs that fit in a single page.

pkg/proto/configuration/blockdevice/blockdevice.proto
meroton-benjamin: Fair enough I don't actually have any numbers on it, it is just the way we use it. For development purposes we have a file to emulate a block device while all our production environments gets an entire block device.

meroton-benjamin:
Related to your initial setup description, but unrelated to the pull request, you mention that you run your raid stack in mirror mode on your 10 hosts, as buildbarn has application support for mirroring and data recovery so your cluster as a whole would be getting better performance from the same amount of storage but instead running a 2 mirrors x 5 shards configuration where the underlying storage is striped instead.

prime000000:
Tried testing write speed with FIO.

It seems that my hypothesis about writeback being bottleneck is wrong, and all speedup I saw is caused by the other change I made.

I used sharding backend locally, to sprint IO between 10 files. File IO is serialized by inode sem in kernel, seems like it was limiting the throughput. And all issues with extent allocation and writeback are insignificant (and even if they are, they are mitigated by splitting writes between files).

When using storage on top of block device, writes are not serialized, and probably that is why other people did not hit this issue.

@meroton-benjamin sorry for wasting your time. Thank you for the feedback again.

> you mention that you run your raid stack in mirror mode on your 10 hosts, as buildbarn has application support for mirroring and data recovery so your cluster as a whole would be getting better performance from the same amount of storage but instead running a 2 mirrors x 5 shards configuration where the underlying storage is striped instead.

I think it would not work in my case. At least, not it that configuration.

buildbarn can't handle node going down (sharding backends does not have fault tolerance). And mirror backend in our testing does not handle faults either (we tried mirroring between two nodes). So I added fallback logic using k8s. When storage node goes down, k8s would recreate pod on another node while losing all the data for the shard. So there is no point in having local mirror.

PR #335, opened 2026-05-03
Implement a fallback storage configuration

vtsao-openai:
This configuration allows you to specify a primary and secondary storage backend, where if the primary goes down, reads & writes will go to the secondary.

Writes to the primary are async, best-effort replicated to the secondary. Writes to the secondary are not replicated back to the primary. This backend provides more availability at the cost of consistency. Although most of the time we don't expect the secondary to be actually used.

The main use case of this backend, at least for us, is to be able to update our backend storage deployment without down time. We currently use a sharded backend storage on k8s via a stateful set. So any deployments currently cause down time as k8s has to take down a pod (storage shard) when updating it. This also helps in general if k8s decides to restart a pod for any reason.

### DIFF
--- a/pkg/blobstore/configuration/BUILD.bazel
+++ b/pkg/blobstore/configuration/BUILD.bazel
@@ -22,6 +22,7 @@ go_library(
     deps = [
         "//pkg/blobstore",
         "//pkg/blobstore/completenesschecking",
+        "//pkg/blobstore/fallback",
         "//pkg/blobstore/grpcclients",
         "//pkg/blobstore/local",
         "//pkg/blobstore/mirrored",
--- a/pkg/blobstore/configuration/new_blob_access.go
+++ b/pkg/blobstore/configuration/new_blob_access.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/buildbarn/bb-storage/pkg/blobstore"
+	"github.com/buildbarn/bb-storage/pkg/blobstore/fallback"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/local"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/mirrored"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/readcaching"
@@ -379,6 +380,19 @@ func (nc *simpleNestedBlobAccessCreator) newNestedBlobAccessBare(configuration *
 			BlobAccess:      readfallback.NewReadFallbackBlobAccess(primary.BlobAccess, secondary.BlobAccess, replicator),
 			DigestKeyFormat: primary.DigestKeyFormat.Combine(secondary.DigestKeyFormat),
 		}, "read_fallback", nil
+	case *pb.BlobAccessConfiguration_Fallback:
+		primary, err := nc.NewNestedBlobAccess(backend.Fallback.Primary, creator)
+		if err != nil {
+			return BlobAccessInfo{}, "", err
+		}
+		secondary, err := nc.NewNestedBlobAccess(backend.Fallback.Secondary, creator)
+		if err != nil {
+			return BlobAccessInfo{}, "", err
+		}
+		return BlobAccessInfo{
+			BlobAccess:      fallback.NewFallbackBlobAccess(primary.BlobAccess, secondary.BlobAccess),
+			DigestKeyFormat: primary.DigestKeyFormat.Combine(secondary.DigestKeyFormat),
+		}, "fallback", nil
 	case *pb.BlobAccessConfiguration_Demultiplexing:
 		// Construct a trie for each of the backends specified
 		// in the configuration indexed by instance name prefix.
--- a/pkg/blobstore/fallback/BUILD.bazel
+++ b/pkg/blobstore/fallback/BUILD.bazel
@@ -1,0 +1,34 @@
+load("@rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "fallback",
+    srcs = ["fallback_blob_access.go"],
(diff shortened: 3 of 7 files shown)


EdSchouten:
> The main use case of this backend, at least for us, is to be able to update our backend storage deployment without down time. We currently use a sharded backend storage on k8s via a stateful set. So any deployments currently cause down time as k8s has to take down a pod (storage shard) when updating it. This also helps in general if k8s decides to restart a pod for any reason.

Out of curiosity, why are you doing this? Why not run bb-storage as a simple deployment? That way Kubernetes is capable of spinning up replacements before shutting down the old pod.

vtsao-openai:
> > The main use case of this backend, at least for us, is to be able to update our backend storage deployment without down time. We currently use a sharded backend storage on k8s via a stateful set. So any deployments currently cause down time as k8s has to take down a pod (storage shard) when updating it. This also helps in general if k8s decides to restart a pod for any reason.
> 
> Out of curiosity, why are you doing this? Why not run bb-storage as a simple deployment? That way Kubernetes is capable of spinning up replacements before shutting down the old pod.

Hey @EdSchouten it's because we're using PVs for our disk because ephemeral storage isn't enough for us. So I think we'd run into the same issue whether we're using deployments or stateful sets - even with deployments the PVC can only be mounted to a single pod so I think we'd run into the same issue - at some point we have to switch the PVC mount over to the new pod (which will result in downtime). And the disk types we require (for latency reasons) are all `ReadWriteOnce`, so only one pod can be mounted to them.

So I'm happy to be wrong, but I'm not sure this can be solved purely with k8s. I think we need this kind of fallback mechanism natively in Buildbarn.

artyrian:
We have a similar Buildbarn setup with multiple shards (and also tried additional replicas per shard), it doesn’t provide true HA in k8s terms.
I also reviewed the ADR (https://github.com/buildbarn/bb-adrs/blob/main/0002-storage.md#adding-fault-tolerance), but didn't find a straightforward way to achieve fast shard failover with PVCs without downtime.

vtsao-openai:
Yeah the mirrored backend does not actually provide HA, I think it's even documented that it does not in the proto comments.

@EdSchouten another benefit of this fallback approach is not just for deployments, but in case the storage shard actually goes down for whatever reason, builds will not fail. Yes it's at the cost of consistency (which in our case is probably fine - it should be no different than if the digests just didn't exist in the cache in the first place).

Also this allows people not using k8s to be able to achieve some more availability if they want to use this backend.

moroten:
> EdSchouten another benefit of this fallback approach is not just for deployments, but in case the storage shard actually goes down for whatever reason, builds will not fail. Yes it's at the cost of consistency (which in our case is probably fine - it should be no different than if the digests just didn't exist in the cache in the first place).

Consider the case where Bazel asks the remote cluster to execute an action. The output is stored in mirror A because B is down. 5 minutes later, Bazel wants to use the output but A is down and B doesn't have it. The difference from a blob missing from the start is that Bazel also knows it is missing. In this case, Bazel should be able to assume that the blob still exists, it did 5 minutes ago.

artyrian:
> The output is stored in mirror A because B is down. 5 minutes later, Bazel wants to use the output but A is down and B doesn't have it. The difference from a blob missing from the start is that Bazel also knows it is missing.

Isn't this inconsistency equivalent to a normal cache eviction? If a blob gets evicted between the time Bazel stores it and tries to reuse it, the action cache would similarly point to a missing CAS entry. Bazel handles by reexecuting, so why is the fallback case worse? Or is the concern about bazel clients that don't gracefully handle (AC hit + CAS miss) and fail hard?

artyrian:
friendly bump, we have the same StatefulSet + RWO PVC setup and this PR is what we need for zero-downtime rolling updates.

One thing: will adopting fallback also require rebuilding bb-frontend, bb-scheduler, etc?

Happy to help test or push this forward if there's anything blocking. Thanks!

vtsao-openai:
> > The output is stored in mirror A because B is down. 5 minutes later, Bazel wants to use the output but A is down and B doesn't have it. The difference from a blob missing from the start is that Bazel also knows it is missing.
> 
> Isn't this inconsistency equivalent to a normal cache eviction? If a blob gets evicted between the time Bazel stores it and tries to reuse it, the action cache would similarly point to a missing CAS entry. Bazel handles by reexecuting, so why is the fallback case worse? Or is the concern about bazel clients that don't gracefully handle (AC hit + CAS miss) and fail hard?

Yeah I'm also thinking something similar - in the worst case scenario it's a cache miss and Bazel tries again right? I understand it's not correct if these mirrors aren't completely consistent, but that's the trade off here for a more highly available storage - and for our use case it's something we're okay with for the higher availability.

EdSchouten:
Sorry, but I'd rather not merge changes like these. The downside of backends like these is that they behave poorly in case of 'split brain' situations, which also happen to be hard to diagnose.

Imagine the case you have 2 storage backend processes, and 2 frontend processes in front of it that use this fallback option. Now assume that some network issue occurs where frontend A is only capable of reaching backend A, and frontend B is only capable of reaching backend B. Then in that case the frontends will behave inconsistently, and will continue to do so.

Right now we depend on Kubernetes to make the decision which storage nodes are authoritative. The API servers will use Raft/Paxos internally to reach consensus, and all frontends will end up using the same set of processes by using service routing.

I fully understand that Kubernetes StatefulSets have some inherent limitations that make them annoying to use for what we're trying to achieve. My recommendation would therefore be to not use StatefulSets, but n (or 2*n in case of mirroring) separate deployments. If this is undesirable (e.g., because it causes us to leave performance/cost savings/... on the table), please work with the Kubernetes team to extend StatefulSets to provide the features you want.

PR #336, opened 2026-05-07
feat(jwt): support RSA-PSS signatures (PS256/PS384/PS512)

chagui:
This PR adds RSA-PSS (PS256/PS384/PS512) verification to the JWT validator, alongside the existing PKCS#1 v1.5 (RS256/RS384/RS512) support.
We want to support an internal identity provider that signs tokens with PSS.

### DIFF
--- a/pkg/jwt/rsa_sha_signature_validator.go
+++ b/pkg/jwt/rsa_sha_signature_validator.go
@@ -33,15 +33,6 @@ func (sv *rsaSHASignatureValidator) ValidateSignature(algorithm string, keyID *s
 	var hasher hash.Hash
 	var pssOpts *rsa.PSSOptions
 	switch algorithm {
-	case "RS256":
-		hashType = crypto.SHA256
-		hasher = sha256.New()
-	case "RS384":
-		hashType = crypto.SHA384
-		hasher = sha512.New384()
-	case "RS512":
-		hashType = crypto.SHA512
-		hasher = sha512.New()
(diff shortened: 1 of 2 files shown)


pkg/jwt/rsa_sha_signature_validator.go
```diff
@@ -40,9 +42,25 @@ func (sv *rsaSHASignatureValidator) ValidateSignature(algorithm string, keyID *s
 	case "RS512":
 		hashType = crypto.SHA512
 		hasher = sha512.New()
+	case "PS256":
```
EdSchouten, line 36: Could we keep this sorted alphabetically (i.e., PS* before RS*)?

EdSchouten (commented):
Just checking: RS* and PS* always use the same JSON Web Keys? No distinct formats between the two? If so, then it's fine to keep this all implemented by a single type.

chagui:
> Just checking: RS* and PS* always use the same JSON Web Keys? No distinct formats between the two? If so, then it's fine to keep this all implemented by a single type.

That's my understanding, asked an LLM to be sure:
> RFC 7518 §6.3 defines a single RSA JWK shape (kty=RSA, params n/e) with no distinction between RS* and PS*; padding is selected by the JWS header alg. go-jose decodes both into *rsa.PublicKey, so a single validator type is the right fit.

Pushed the alphabetical sort as new commit, feel free to squash-merge

PR #338, opened 2026-05-14
perf: reduce allocations on the digest and buffer hot paths

MarshallOfSound:
> [!NOTE]
> This PR was authored with AI assistance and reviewed by me before opening.

Four small allocation/CPU fixes on the digest, buffer, and key-location-map hot paths. Each commit is independent and explains its own win; this is the summary.

**`digest: avoid fmt.Sprintf in newDigestUnchecked`** — every digest constructed from a proto goes through this. `fmt.Sprintf` parses the format string and reflects on its arguments at runtime; replacing it with a pre-sized `[]byte` + `strconv.AppendInt` is ~3.5× faster per call (76.9 → 22.1 ns). Output is byte-identical, including two-digit enum values.

**`digest: pre-size SetBuilder map and use slices.SortFunc`** — `NewSetBuilder()` now takes a capacity argument so callers can pre-size the backing map. Callers that always fill the set to a known size (request parsing, per-directory file sets, the completeness checker's bounded queue, per-shard partitions) pass that size; callers that build a result of unknown size — chiefly the `missing` set returned by `FindMissing`, where the count depends on the caller's hit rate — pass zero, since pre-allocating the worst case would waste memory and CPU on the common high-hit-rate path. Also replaces `sort.Sort` (interface dispatch per comparison) with `slices.SortFunc`. Sort order unchanged.

**`blobstore/buffer: pre-allocate ToByteSlice result instead of io.ReadAll`** — both `ToByteSlice` implementations know the blob size but used `io.ReadAll`, which append-doubles from 512 bytes (≈2N transient bytes, ~N redundant memcpy). Pre-allocate and `io.ReadFull` instead, matching what `toByteSliceViaChunkReader` already does. Zero-byte blobs need a single explicit `Read()` to drive the validating reader since `io.ReadFull` won't issue one for an empty buffer. Slight behavior change for non-empty blobs: a short read now returns `io.ErrUnexpectedEOF` instead of silently returning a truncated slice — on this path the size is trusted, so a short read indicates a corrupted backend and erroring seems more correct, but flagging it.

**`blobstore/local: share a sentinel error for key-location map misses`** — `hashingKeyLocationMap.Get()` constructed a fresh `status.Error(codes.NotFound, …)` on every lookup miss. Since `Get()` runs once per digest in `FindMissing()`, a CPU profile shows it accounting for ~8% of total CPU at a 50% hit rate. Replaced with a package-level sentinel; every caller only inspects `status.Code()`, never the message or error identity.

Measured against a real `bb_storage` binary on localhost (in-memory and file-backed-mmap CAS configs), driven by a load generator over gRPC, 3 interleaved A/B runs per workload, mean ±σ:

| workload | baseline | patched | Δ |
|---|---|---|---|
| `FindMissingBlobs`, 2000 dig/RPC, 1% miss, in-memory CAS | 1157K ±25K dig/s | 1347K ±9K dig/s | +16.4% |
| `FindMissingBlobs`, 2000 dig/RPC, 50% miss, in-memory CAS | 974K ±6K dig/s | 1136K ±6K dig/s | +16.6% |
| `FindMissingBlobs`, 2000 dig/RPC, 99% miss, in-memory CAS | 881K ±5K dig/s | 1040K ±2K dig/s | +18.1% |
| `BatchReadBlobs`, 200×64 KiB/RPC, file-backed CAS | 1965 ±10 MB/s | 2400 ±6 MB/s | +22.2% |
| `BatchReadBlobs`, 200×64 KiB/RPC, in-memory CAS | 2761 ±50 MB/s | 2798 ±21 MB/s | +1.4% |

The `FindMissingBlobs` win is consistent across the full range of miss rates (1% → 99% all show +15–19%, p50 latency −13–16%) — the sentinel error matters more at higher miss rates while the digest/SetBuilder changes are constant. The in-memory `BatchReadBlobs` row barely moves because in-memory blocks return zero-copy `[]byte` slices, so the `io.ReadAll` path never executes there. The file-backed config (which I think is closer to a typical deployment) is where the buffer fix lands.

`pprof` allocation diff over the file-backed read benchmark shows `io.ReadAll`'s contribution dropping from ~110 GB to zero, and `runtime.memmove` flat CPU dropping from 11.0% to 4.4%.

> [!NOTE]
> The `digest: add test coverage` commit pins the digest string encoding for size 0, `math.MaxInt64`, and an empty instance name, plus tests for `NewSetBuilder()` capacity and `Build()` ordering.

### DIFF
--- a/pkg/digest/set_builder.go
+++ b/pkg/digest/set_builder.go
@@ -1,7 +1,8 @@
 package digest
 
 import (
-	"sort"
+	"slices"
+	"strings"
 )
 
 // SetBuilder is a builder for Set objects.
@@ -9,10 +10,15 @@ type SetBuilder struct {
 	digests map[Digest]struct{}
 }
 
-// NewSetBuilder creates a SetBuilder that contains no initial elements.
-func NewSetBuilder() SetBuilder {
+// NewSetBuilder creates a SetBuilder that contains no initial
+// elements. The capacity argument pre-sizes the underlying map so that
+// adding up to that many elements does not require rehashing. Callers
+// should provide a sensible estimate based on the context in which the
+// set is built (e.g., the number of digests in an incoming request).
+// Pass zero when no estimate is available.
+func NewSetBuilder(capacity int) SetBuilder {
 	return SetBuilder{
-		digests: map[Digest]struct{}{},
+		digests: make(map[Digest]struct{}, capacity),
 	}
 }
 
@@ -37,29 +43,17 @@ func (sb SetBuilder) Build() Set {
 	}
 
 	// Store all digests in a list.
-	digests := make(digestList, 0, len(sb.digests))
+	digests := make([]Digest, 0, len(sb.digests))
 	for digest := range sb.digests {
 		digests = append(digests, digest)
 	}
 
 	// Sort the list for determinism and to allow linear time
 	// implementations of GetDifferenceAndIntersection() and
-	// GetUnion().
-	sort.Sort(digests)
+	// GetUnion(). slices.SortFunc avoids the per-comparison
+	// interface dispatch of sort.Sort.
+	slices.SortFunc(digests, func(a, b Digest) int {
+		return strings.Compare(a.value, b.value)
+	})
 	return Set{digests: digests}
-}
-
-// digestList implements a list of digests that is sortable.
-type digestList []Digest
-
-func (l digestList) Len() int {
-	return len(l)
-}
-
-func (l digestList) Less(i, j int) bool {
-	return l[i].String() < l[j].String()
-}
-
-func (l digestList) Swap(i, j int) {
-	l[i], l[j] = l[j], l[i]
 }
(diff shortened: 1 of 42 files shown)


pkg/digest/set_builder.go
```diff
@@ -16,6 +17,16 @@ func NewSetBuilder() SetBuilder {
 	}
 }
 
+// NewSetBuilderWithCapacity creates a SetBuilder that contains no
+// initial elements, pre-sized to hold approximately n elements without
+// rehashing. Use this when the number of elements is known up front,
+// such as when constructing a Set from a FindMissingBlobs request.
+func NewSetBuilderWithCapacity(n int) SetBuilder {
```
EdSchouten: Instead of adding a new function, can we please just change `NewSetBuilder()` to take the capacity? I think that in (almost) every case we should be able to come up with some kind of sensible initial capacity based on the context in which the sets are created. Thanks!

EdSchouten (commented):
Great work! One tiny remark.

pkg/digest/set_builder.go
MarshallOfSound: Sounds good, there are a couple of places we're likely to slightly-over-allocate E.g. `digests.Length()` when 0 -> N could be missing) but loose benchmarks show that's fine (trading tiny additional memory pressure for CPU cycles)

MarshallOfSound: @EdSchouten updated and ran benchmarks to ensure various hit / miss / fill % don't regress, looking good

PR #339, opened 2026-05-14
perf: reduce allocations on the chunk reading hot path

matshch:
> [!NOTE]
> This PR was authored with AI assistance and reviewed by me before opening.

`readerBackedChunkReader` previously allocated a fresh `maximumChunkSizeBytes` slice on every `Read()` call. With the default 64 KiB chunk size and production traffic around 20-30 KRPS, this allocation site accounted for the bulk of GC pressure on bb_storage.

The reader now owns a single buffer for its entire lifetime, acquired lazily on first `Read()` and returned to a package-private `sync.Pool` on `Close()`. The `ChunkReader` contract is tightened to reflect that returned slices are only valid until the next `Read()`/`Close()` call. The only place where this change mattered is gRPC server, now we use `grpc.PreparedMsg` to move the marshaling of the message in the main loop, which removes references to the allocated buffers.

The overall effect depends heavily on the load pattern, in our case we are seeing up to 3x CPU usage reduction on production cluster.

### DIFF
--- a/pkg/blobstore/buffer/reader_backed_chunk_reader.go
+++ b/pkg/blobstore/buffer/reader_backed_chunk_reader.go
@@ -8,6 +8,7 @@ type readerBackedChunkReader struct {
 	r                     io.ReadCloser
 	maximumChunkSizeBytes int
 
+	buf *[]byte
 	err error
 }
 
@@ -23,8 +24,11 @@ func newReaderBackedChunkReader(r io.ReadCloser, maximumChunkSizeBytes int) Chun
 
 func (r *readerBackedChunkReader) Read() ([]byte, error) {
 	if r.err == nil {
-		b := make([]byte, r.maximumChunkSizeBytes)
-		n, err := io.ReadFull(r.r, b[:])
+		if r.buf == nil {
+			r.buf = getChunkBuffer(r.maximumChunkSizeBytes)
+		}
+		b := *r.buf
+		n, err := io.ReadFull(r.r, b)
 		if err == io.ErrUnexpectedEOF {
 			r.err = io.EOF
 		} else {
@@ -38,5 +42,9 @@ func (r *readerBackedChunkReader) Read() ([]byte, error) {
 }
 
 func (r *readerBackedChunkReader) Close() {
+	if r.buf != nil {
+		putChunkBuffer(r.buf)
+		r.buf = nil
+	}
 	r.r.Close()
 }
--- a/pkg/blobstore/grpcservers/byte_stream_server.go
+++ b/pkg/blobstore/grpcservers/byte_stream_server.go
@@ -13,6 +13,7 @@ import (
 	bb_zstd "github.com/buildbarn/bb-storage/pkg/zstd"
 
 	"google.golang.org/genproto/googleapis/bytestream"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
@@ -56,7 +57,18 @@ func (s *byteStreamServer) Read(in *bytestream.ReadRequest, out bytestream.ByteS
 			if readErr != nil {
 				return readErr
 			}
-			if writeErr := out.Send(&bytestream.ReadResponse{Data: readBuf}); writeErr != nil {
+			// A ChunkReader may reuse the slice returned by
+			// Read() across calls. Encoding into a grpc.PreparedMsg
+			// copies the wire-format bytes synchronously, so the
+			// underlying buffer is no longer referenced by gRPC after
+			// SendMsg returns. This is the only contract-safe way to
+			// hand a reusable buffer to grpc.SendMsg; see
+			// google.golang.org/grpc issues #5857 and #8186.
+			var prepared grpc.PreparedMsg
+			if err := prepared.Encode(out, &bytestream.ReadResponse{Data: readBuf}); err != nil {
+				return err
+			}
+			if writeErr := out.SendMsg(&prepared); writeErr != nil {
 				return writeErr
 			}
 		}
@@ -81,7 +93,14 @@ type readStreamWriter struct {
 }
 
 func (w *readStreamWriter) Write(p []byte) (int, error) {
-	if err := w.out.Send(&bytestream.ReadResponse{Data: p}); err != nil {
+	// Each Write encodes the chunk into a grpc.PreparedMsg before sending,
+	// so the caller's buffer is no longer referenced by gRPC after Write
+	// returns. See the IDENTITY path above for rationale.
+	var prepared grpc.PreparedMsg
+	if err := prepared.Encode(w.out, &bytestream.ReadResponse{Data: p}); err != nil {
+		return 0, err
+	}
+	if err := w.out.SendMsg(&prepared); err != nil {
 		return 0, err
 	}
 	return len(p), nil
--- a/pkg/blobstore/buffer/BUILD.bazel
+++ b/pkg/blobstore/buffer/BUILD.bazel
@@ -11,6 +11,7 @@ go_library(
         "cas_reader_buffer.go",
         "cas_validating_chunk_reader.go",
         "cas_validating_reader.go",
+        "chunk_buffer_pool.go",
         "chunk_reader.go",
         "chunk_reader_backed_reader.go",
         "common_conversions.go",
@@ -56,6 +57,7 @@ go_test(
         "new_validated_buffer_from_byte_slice_test.go",
         "new_validated_buffer_from_reader_at_test.go",
         "reader_backed_chunk_reader_bench_test.go",
+        "reader_backed_chunk_reader_test.go",
         "with_error_handler_test.go",
     ],
     deps = [
--- a/pkg/blobstore/buffer/BUILD.bazel
+++ b/pkg/blobstore/buffer/BUILD.bazel
@@ -55,6 +55,7 @@ go_test(
         "new_proto_buffer_from_proto_test.go",
         "new_validated_buffer_from_byte_slice_test.go",
         "new_validated_buffer_from_reader_at_test.go",
+        "reader_backed_chunk_reader_bench_test.go",
         "with_error_handler_test.go",
     ],
     deps = [
--- a/pkg/blobstore/buffer/chunk_buffer_pool.go
+++ b/pkg/blobstore/buffer/chunk_buffer_pool.go
@@ -1,0 +1,26 @@
+package buffer
+
+import "sync"
+
+// chunkBufferPool caches []byte buffers used by ChunkReader
+// implementations that allocate a fixed-size working buffer per
+// stream. Buffers are stored as *[]byte to avoid the interface-boxing
+// allocation that occurs when storing a non-pointer value in
+// sync.Pool.
+var chunkBufferPool sync.Pool
+
+func getChunkBuffer(size int) *[]byte {
+	if v := chunkBufferPool.Get(); v != nil {
+		b := v.(*[]byte)
+		if cap(*b) >= size {
+			*b = (*b)[:size]
+			return b
+		}
+	}
+	b := make([]byte, size)
+	return &b
+}
+
+func putChunkBuffer(b *[]byte) {
+	chunkBufferPool.Put(b)
+}
--- a/pkg/blobstore/buffer/chunk_reader.go
+++ b/pkg/blobstore/buffer/chunk_reader.go
@@ -10,6 +10,12 @@ import (
 // responsible for providing space for the data. This interface is
 // similar to how frame-based transfer protocols work, including the
 // Bytestream protocol used by REv2.
+//
+// The byte slice returned by Read() is only guaranteed to be valid
+// until the next call to Read() or Close() on the same ChunkReader.
+// Implementations are free to reuse the backing storage across Read()
+// calls in order to minimize allocations. Callers that need to retain
+// the contents past the next Read()/Close() must copy them.
 type ChunkReader interface {
 	Read() ([]byte, error)
 	Close()
--- a/pkg/blobstore/buffer/multiplexed_chunk_reader.go
+++ b/pkg/blobstore/buffer/multiplexed_chunk_reader.go
@@ -32,6 +32,15 @@ func newMultiplexedChunkReader(r ChunkReader, additionalConsumers int) ChunkRead
 }
 
 func (r *multiplexedChunkReader) readAndShareWithOthers(currentConsumerContinues int) ([]byte, error) {
+	// All waiting consumers receive the exact same slice returned by
+	// the underlying ChunkReader.Read(). This is safe with respect
+	// to the "valid until next Read()/Close()" contract because the
+	// pendingConsumers counter ensures every consumer must call
+	// Read() (or Close()) again before another underlying Read()
+	// can be issued. By the time the next underlying Read()
+	// executes, every consumer has therefore released its reference
+	// to the previously returned slice, so the underlying reader is
+	// free to reuse the backing storage.
 	data, err := r.r.Read()
 	for _, c := range r.waitingConsumers {
 		c <- readResult{data: data, err: err}
--- a/pkg/blobstore/buffer/reader_backed_chunk_reader_bench_test.go
+++ b/pkg/blobstore/buffer/reader_backed_chunk_reader_bench_test.go
@@ -1,0 +1,82 @@
+package buffer_test
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
+)
+
+type readAtCloser struct{ *bytes.Reader }
+
+func (readAtCloser) Close() error { return nil }
+
+const benchChunkSize = 1 << 16
+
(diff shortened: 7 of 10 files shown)


EdSchouten (commented):
Are we sure that this is safe? In the bytestream server we place these chunks in messages that we pass to gRPC's `.Send()`. Is it safe to mutate/reuse these buffers immediately afterwards?

matshch:
From the practical point of view it is safe: the current implementation of grpc-go always marshals the message before returning from `Send()`, that is what my AI agent noticed and used to come to the conclusion that this code is safe.

But it looks like the contract actually does not allow that, https://pkg.go.dev/github.com/grpc/grpc-go#ServerStream.SendMsg:
> It is not safe to modify the message after calling `SendMsg`. Tracing libraries and stats handlers may use the message lazily.

(and that is the method called in https://github.com/googleapis/go-genproto/blob/3700d4141b60f18b5e65fb06d6b8ccebd3ff6384/googleapis/bytestream/bytestream.pb.go#L842)

The exact same issue was brought up in grpc/grpc-go#8186, and the maintainers do recommend using `grpc.PreparedMsg` as a workaround. It moves the marshaling outside of the `Send` call, and passes the already marshaled message around. And during marshaling all the bytes are copied into the new buffer, so it should be safe to change the old one.

Let me know if that works for you, or if you deem this too unsafe to merge.

MarshallOfSound:
Fwiw this was a follow up to my last perf PR that I was still bashing around a bit in a prod cluster.

I can confirm the prod wins just the safety of it gave me pause. I'll compare what I've been running to this PR and see if there's any edge case differences

matshch:
We've rolled out pre-grpc.PreparedMsg version in production and it looks functional. We need more time to say for sure, but it looks like both #338 and this one together give us 2x performance boost.

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -48,6 +49,14 @@ func (s *byteStreamServer) Read(in *bytestream.ReadRequest, out bytestream.ByteS
...
 		defer r.Close()
 
+		// readerBackedChunkReader may reuse the slice returned by
+		// Read() across calls. Encoding into a grpc.PreparedMsg
+		// copies the wire-format bytes synchronously, so the
+		// underlying buffer is no longer referenced by gRPC after
+		// SendMsg returns. This is the only contract-safe way to
+		// hand a reusable buffer to grpc.SendMsg; see
+		// google.golang.org/grpc issues #5857 and #8186.
+		var prepared grpc.PreparedMsg
```
tamird: isn't this just as unsafe? this code now mutates `prepared` after calling `SendMsg` but the documentation on `SendMsg` says explicitly:

```
	// It is not safe to modify the message after calling SendMsg. Tracing
	// libraries and stats handlers may use the message lazily.
```

which means that this `prepared` needs to be inside the loop body, where escape analysis will put it on the heap and performance will be basically unchanged.

https://pkg.go.dev/google.golang.org/grpc#ServerStream

it seems that `PreparedMsg` is in fact useful only in fanout scenarios where the same encoded message is meant to be sent on multiple transports.

tamird: I had my coding agent test this: try to reuse `PreparedMsg` on both client and server while polling `/debug/requests` which renders lazy trace payloads; as expected it worked fine in a normal build and blew up under the race detector. So this optimization seems to be unsound.

matshch: Oh, you are right, we need to move the `prepared` inside of the loop.

matshch: You really need to keep an eye on all these AI agents 😅

tamird: Indeed. The trouble is that with `prepared` moved inside the loop all the performance gains are erased.

matshch: Not really — we had the same message marshaling inside of the `SendMsg`, we just moved it out of the function call. So in the end:
* This PR optimized allocation of buffers during chunk reading.
* This PR does not affect allocation of gRPC messages — that is something we can work on in the future.

tamird (changes requested):
Directionally I think this PR is not right. `sync.Pool` is basically a hack here to attempt to improve the performance of an implementation of `ChunkReader` but that interface is inherently broken. It is simply not reasonable or possible in the general case to return owned byte slices from an interface - any interface - without forcing oodles of allocations.

The second problem is that the amortized allocations do not actually vanish in this PR because of the continuously allocated `PreparedMsg`. This PR is just moving the allocations around: before they were in `readerBackedChunkReader` and now they are in the `grpc.PreparedMsg.Encode` call.

The benchmarks in this PR are misleading; microbenchmarking around the broken interface doesn't demonstrate anything useful.

I'm planning to send another PR - which is sadly performance-neutral - that removes the broken `ChunkReader` interface entirely.

pkg/blobstore/grpcservers/byte_stream_server.go
tamird: I don't think that's true, and there are no benchmarks in this PR to prove it.

matshch: I have checked the version without prepared messages with pprof and I don't see any issues with allocations in the chunk reading path that were there before. I'll run the new PR version in production soon and see how it behaves under load.

matshch:
> `sync.Pool` is basically a hack here to attempt to improve the performance of an implementation of `ChunkReader` but that interface is inherently broken. It is simply not reasonable or possible in the general case to return owned byte slices from an interface - any interface - without forcing oodles of allocations.

I agree, but in the program which whole job is to read some chunks from disk and then chug them into the network — there is not whole lot to do except to manage memory, so that is what we have to optimize.

I'm not quite sure how to approach this issue without owning the buffers, I would be glad to see your PR.

> The second problem is that the amortized allocations do not actually vanish in this PR because of the continuously allocated `PreparedMsg`. This PR is just moving the allocations around: before they were in `readerBackedChunkReader` and now they are in the `grpc.PreparedMsg.Encode` call.

You are mistaking two different allocation paths:
* Allocations from `readerBackedChunkReader` are gone.
* Allocations from `ServerStream.SendMsg` have moved to `PreparedMsg.Encode`.

So this PR should be performance beneficial, but I am not sure how to demonstrate that easily. I guess we need another benchmark that will hook this server to some client to test out the whole pipeline? I'll measure it empirically on our production soon.

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -56,7 +57,18 @@ func (s *byteStreamServer) Read(in *bytestream.ReadRequest, out bytestream.ByteS
 			if readErr != nil {
 				return readErr
 			}
-			if writeErr := out.Send(&bytestream.ReadResponse{Data: readBuf}); writeErr != nil {
+			// readerBackedChunkReader may reuse the slice returned by
```
EdSchouten: s/readerBackedChunkReader/ChunkReader/

The fact that there is a type named readerBackedChunkReader is an implementation of the existing buffer layer. I wouldn't refer to that type here.

pkg/blobstore/buffer/reader_backed_chunk_reader.go
```diff
@@ -23,8 +24,11 @@ func newReaderBackedChunkReader(r io.ReadCloser, maximumChunkSizeBytes int) Chun
 
 func (r *readerBackedChunkReader) Read() ([]byte, error) {
 	if r.err == nil {
-		b := make([]byte, r.maximumChunkSizeBytes)
-		n, err := io.ReadFull(r.r, b[:])
+		if r.buf == nil {
+			r.buf = getChunkBuffer(r.maximumChunkSizeBytes)
```
EdSchouten, line 28: As far as I understand, you're trying to solve the issue that transmitting a large file causes O(n) allocations. However, with this specific change you're trying to reduce this to 'less-than-1' allocation by using a sync.Pool. Is that really necessary?

In other words, why don't we just write this instead?
```go
r.buf = make([]byte, r.maximumChunkSizeBytes)
```

Sure, it's one more allocation. But are we sure that that's actually part of the bottleneck?

matshch: The main issue is that bb-storage on our workload is GC-bound:
Stock bb-storage CPU flamegraph

And the biggest source of allocations is this place:
Stock bb-storage allocations flamegraph

We have all kinds of blob sizes in CAS. Many of them do fit in the default chunk size, but we also have a huge tail of blobs which are in the realm of gigabytes:
image

With this distribution of sizes it looks like we need to optimize both small and huge files, that is why I've looked into `sync.Pool` first, and then found that we can actually reuse one buffer (except for the gRPC part, which is really unfortunate).

matshch:
Here is a new flame graph of allocations: we can see that both `readerBackedChunkReader` and `serverStream.SendMsg` are gone, now we only see `grpc.PreparedMsg.Encode`:
PR bb-storage allocations flame graph

<details><summary>Old flame graph to compare side by side</summary>Stock bb-storage allocations flame graph</details>

That means:
* `readerBackedChunkReader` is successfully optimized.
* `grpc.PreparedMsg` does not introduce additional allocations, we just moved allocations outside of `serverStream.SendMsg` and now control the allocations.


If we speak about CPU load, here is a new CPU flame graph:
PR bb-storage CPU flame graph

<details><summary>Old flame graph to compare side by side</summary>Stock bb-storage CPU flame graph</details>

Now GC takes only ~25% of CPU time. The main win here is that it is <=25%:
* The GC mark phase is limited to 25% of allocatable CPU in Go runtime. This limit is hardcoded in the runtime, you cannot really affect it in any way.
* If you are GC-bound and GC takes more than 25% of your CPU time, that effectively means that you cannot utilize all allocated CPU, as the GC mark phase cannot take more than 25% of allocatable CPU, and you are waiting for mark phase to finish before you can continue serving other requests.
* That means that previously bb-storage would never reach the allocated CPU limit, but it would still throttle on GC and make some requests time out.
* Now GC as the whole is less than 25%, so you can better control the allocated CPU.

I'll run this PR in our cluster a couple more days and report back how we feel, but overall it looks like a huge win for us.

matshch:
Hi, thanks for the patience! Last week we were diving deeper and deeper into our bb-storage instance performance. I now have a lot better understanding of what is happening there, and the fix in this PR definitely helps in our scenario, but it is not a silver bullet as it might have seem from initial micro-benchmarks.

The main issue that we are observing is that when a bb-storage instance is overwhelmed with requests, it responds slower to each of those requests (due to limited resources available to the instance), which means that with high RPS we are getting more and more in-flight requests, which means more allocated resources. Go GC runs when there is a certain ratio of new allocations reached, regardless of the memory still available to the process, so in this runaway scenario each allocation creates additional load on GC _and_ makes us run GC more often, which does not help.

We have not yet found specific bottlenecks that we are hitting, but GC contention is visible in pprof, and reducing it helps with overall CPU usage. So this PR looks to me like a good optimization to have, if you agree with new contracts imposed.

Here is a chart of CPU usage over the number of requests processed in our cluster. We have deployed a version from this PR over the previous weekend, so the left half of the chart is the base branch and the right half is this PR branch. You can see that while there are still a lot of noticeable peaks in CPU usage, they are lower overall, as well as the median CPU usage.
Chart showing bb-storage CPU usage over the number of requests processed

I have also added some benchmarks that exercise the whole gRPC server loop. They still don't represent the workload ideally, as the network latency is not modeled, but at least they can be used to compare the effects of this PR more directly. One of the benchmarks tries to simulate the GC contention described above by setting low memory limit. It is not quite the same thing as happens in the real life, but it allows to see a 10x throughput improvement on 1MiB blob size (and no degradation on the rest of the cases).

Let me know what you think!

tamird:
It would be great to update the PR description and remove the micro benchmark which is, as discussed, rather meaningless.

matshch:
Updated the description to only reflect the main gist of this PR.

pkg/blobstore/buffer/reader_backed_chunk_reader.go
EdSchouten: Again, can't we first do some benchmarking to see whether making a single allocation is sufficient?


I am not a fan of using sync.Pool here. I get why sync.Pool exists. Namely, you often want to use it for holding objects that are complex to initialize. But in our case that shouldn't need to apply. We're just talking about simple byte arrays here. If we already need to use sync.Pool for storing simple byte slices, then Go's memory management must be _really_ bad.

I also suspect that a big amount of waste is coming from the fact that we always allocate `maximumChunkSizeBytes` instead of  something like:

```
max(min(maximumChunkSizeBytes, digest.GetSizeBytes()), 1)
```

tamird: Over-allocating is wasteful, but not the kind of waste that introduces GC overhead. GC overhead is sensitive to the number of allocations and not especially to their size.

I agree that `sync.Pool` is a hack here.

matshch: > I get why sync.Pool exists. Namely, you often want to use it for holding objects that are complex to initialize.

That's not exactly true. https://pkg.go.dev/sync#Pool:
> Pool's purpose is to cache allocated but unused items for later reuse, relieving pressure on the garbage collector. [...]
> 
> An appropriate use of a Pool is to manage a group of temporary items silently shared among and potentially reused by concurrent independent clients of a package. Pool provides a way to amortize allocation overhead across many clients.
> 
> An example of good use of a Pool is in the fmt package, which maintains a dynamically-sized store of temporary output buffers.

https://victoriametrics.com/blog/go-sync-pool/ provides several examples where the standard library utilizes pools for simple objects. https://cs.opensource.google/go/go/+/master:src/net/http/server.go;l=857-870;drc=15b9fc2659f77608548cb279c5e0565b0664cfca;bpv=1;bpt=1 is basically our exact case — just allocating a buffer of constant size.

The point of the pool here is to make sure that we allocate less buffers overall, so GC needs to do less work overall. It is usually negligible when we have other work to do, but in case of bb-storage it might be noticeable depending on the workload.

I have done some benchmarking with both `sync.Pool` and without it:
* For large blobs pool does not matter, we are getting the main benefit from reusing the buffer inside the reader.
* For small blobs pool does matter, we are saving ~30% of GC CPU, which means ~30% of total CPU usage when we are GC bound.

matshch: > I also suspect that a big amount of waste is coming from the fact that we always allocate `maximumChunkSizeBytes` instead of something like:
> 
> ```
> max(min(maximumChunkSizeBytes, digest.GetSizeBytes()), 1)
> ```

It might be a big amount of waste from the point of allocated memory, but I wouldn't say we have significant problems in that regard. And speaking about CPU usage:
* The main driver of GC workload is the number of objects. Having smaller objects probably improves data locality slightly, but I don't think it is significant enough.
* Having many different sizes might actually make GC/allocator work slightly more complex, as these objects will use different size classes.
* It is not obvious how to recycle different-sized objects via `sync.Pool`.

But there is another interesting angle to all of that — Go has a significantly more complex data path for allocations > 32 KB. Maybe it makes sense to decrease the size of these buffers to 32 KB to get lock-free allocations.

matshch: I have conducted several more experiments:
* Decreasing buffer sizes to 32 KB (which are better supported by Go runtime) does not help, as we are getting more overhead from more gRPC messages than from heavier allocation/GC paths.
* `sync.Pool` definitely helps to reduce CPU usage compared to allocating new constant-size buffers each time, especially in case of smaller blobs, as we can reuse them quicker.
* Replacing allocations with new purpose-sized buffers helps a bit, but `sync.Pool` still outperforms this idea by 10-30% on blobs <64 KB.

tamird (changes requested):
Reviewed because @matshch re-requested my review on b993b6789.

The central objection remains unresolved: compare one buffer owned for a reader's lifetime against the global sync.Pool on the complete gRPC path, including the per-message grpc.PreparedMsg. Establishing that the old per-Read allocation was expensive does not establish that pooling improves the simpler lifetime-owned buffer once required gRPC encoding is included. Please publish reproducible before-and-after timing, bytes, allocations, and GC CPU for realistic small and large blobs, and omit the pool unless those results justify it. This also answers Ed's outstanding question.

Please remove the chunk-reader-only microbenchmark; it still excludes gRPC and cannot establish the production-path result. The new 424-line gRPC benchmark is also disproportionately expensive: its largest case allocates 64 × 64 MiB, or 4 GiB, just for the blob fixture, while the GC-pressure variant configures 1,024 concurrent reads and a roughly 100 GiB memory limit. A focused end-to-end comparison with bounded fixtures is sufficient.

TestReaderBackedChunkReaderClosePoolsBuffer treats a pooled pointer appearing within 100 attempts as a correctness property. Go expressly allows sync.Pool to discard an item at any time, so retries cannot make this reliable. Please test the deterministic reader-lifetime and buffer-reuse contract; if the pool remains, do not require that a particular pooled object be returned.

[ tamirdex ]

PR #345, opened 2026-05-31
Add support for SplitBlob and SpliceBlob methods

meroton-benjamin:
This commit adds support for the SplitBlob and SpliceBlob methods from the Remote Execution v2 (REv2) api. SplitBlob and SpliceBlob can be used to facilitate uploads and downloads of large files but a naïve implementation like this has some major drawbacks as well.

The blobs must exist in both their chunked and non chunked form, which may significantly increase storage requirements for large blobs. The protocol gives no guarantee that a large blob stored in the CAS exists in its chunked form which forces you to perform a fairly heavy Split call that loads the entire large blob in order to decomposition it into its chunks.

This implementation mostly exists as a stepping stone for a different implementation where Buildbarn internally manages all blobs as chunked blobs.


pkg/blobstore/grpcclients/cls_blob_access.go
```diff
@@ -0,0 +1,71 @@
...
+func (ba *clsBlobAccess) Put(ctx context.Context, digest digest.Digest, b buffer.Buffer) error {
+	splitBlobResponse, err := b.ToProto(&remoteexecution.SplitBlobResponse{}, ba.maximumMessageSizeBytes)
+	if err != nil {
+		return err
+	}
+	digestFunction := digest.GetDigestFunction()
+	_, err = ba.contentAddressableStorageClient.SpliceBlob(ctx, &remoteexecution.SpliceBlobRequest{
+		InstanceName:   digestFunction.GetInstanceName().String(),
+		DigestFunction: digestFunction.GetEnumValue(),
+		ChunkDigests:   splitBlobResponse.(*remoteexecution.SplitBlobResponse).GetChunkDigests(),
```
EdSchouten: Shouldn't this also set `ChunkingFunction`?

pkg/blobstore/grpcservers/content_addressable_storage_server.go
```diff
@@ -150,10 +152,59 @@ func (contentAddressableStorageServer) GetTree(in *remoteexecution.GetTreeReques
 	return status.Error(codes.Unimplemented, "This service does not support downloading directory trees")
 }
 
-func (contentAddressableStorageServer) SpliceBlob(ctx context.Context, in *remoteexecution.SpliceBlobRequest) (*remoteexecution.SpliceBlobResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "This service does not support splicing blobs")
+func (s *contentAddressableStorageServer) SpliceBlob(ctx context.Context, in *remoteexecution.SpliceBlobRequest) (*remoteexecution.SpliceBlobResponse, error) {
+	if s.chunkListStorage == nil {
```
EdSchouten, line 187: Instead of doing this, why don't we require that ChunkListStorage is non-nil, and that an `ErrorBlobAccess` is passed in?

pkg/blobstore/grpcclients/cls_blob_access.go
```diff
@@ -0,0 +1,71 @@
...
+		BlobDigest:     digest.GetProto(),
+	})
+	return err
+}
+
+func (clsBlobAccess) FindMissing(ctx context.Context, digests digest.Set) (digest.Set, error) {
+	return digest.EmptySet, status.Error(codes.Unimplemented, "Chunk Mapping does not support bulk existence checking")
```
EdSchouten: I don't think that this is correct. This should likely issue a call against FindMissingBlobs() to make sure all ChunkLists exist and are touched.

pkg/blobstore/grpcservers/content_addressable_storage_server.go
EdSchouten, line 32: Shouldn't this also be patched up to call chunkListStorage.FindMissing()?

pkg/capabilities/chunking_provider.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+type chunkingProvider struct {
+	Provider
+	chunkingParameters *remoteexecution.RepMaxCdcParams
+}
+
+// NewChunkingProvider decorates a capabilities provider with support
+// for the split and splice blob api with chunking parameters.
+func NewChunkingProvider(base Provider, chunkingParameters *remoteexecution.RepMaxCdcParams) Provider {
```
EdSchouten: Why can't we use StaticProvider in combination with MergingProvider for this?

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,325 @@
...
+	// Check if we have already computed the result for this blob.
+	if result := ba.checkSplitResult(ctx, d); result != nil {
+		return result
+	}
+
+	// Fallthrough case, compute the chunk list, upload the chunks and
+	// store the chunk list.
+	blobReader := ba.contentAddressableStorage.Get(ctx, d).ToReader()
+	defer blobReader.Close()
+	chunker := NewReaderChunker(d.GetDigestFunction(), blobReader, int64(params.MinChunkSizeBytes), int64(params.HorizonSizeBytes))
```
EdSchouten, line 133: I think we need to have some safety belts in place that these parameters are within bounds. For example, MinChunkSizeBytes needs to be at least 64, as the Gear hashing doesn't work otherwise.

pkg/blobstore/chunklistvalidating/fake_blob_access_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
+	"github.com/buildbarn/bb-storage/pkg/digest"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// fakeBlobAccess provides a thread-safe, in-memory BlobAccess for
+// testing.
+type fakeBlobAccess struct {
```
EdSchouten, line 18: Can't we just use gomock for consistency with most of the other tests in Buildbarn?

cmd/bb_storage/main.go
```diff
@@ -84,6 +84,25 @@ func main() {
...
+					contentAddressableStorageInfo,
+					grpcClientFactory,
+					int(configuration.MaximumMessageSizeBytes),
+				),
+				grpcClientFactory,
+			)
+			if err != nil {
+				return util.StatusWrap(err, "Failed to create Chunk Map")
+			}
+			chunkListStorage = authorizedBackend
```
EdSchouten, line 105: I would have imagined that this added a capabilities provider to `cacheCapabilitiesAuthorizers`. Why doesn't t do that?

pkg/blobstore/grpcclients/cls_blob_access.go
meroton-benjamin: ack

pkg/blobstore/grpcservers/content_addressable_storage_server.go
meroton-benjamin: We could but then the `ContentAddressableStorageServer` would be unable to behave differently when the ChunkListStorage is not defined. If we want the `FindMissingBlobs` implementation from below then we need to know when we need to verify chunk list storage and when we do not.

pkg/blobstore/grpcclients/cls_blob_access.go
meroton-benjamin: The chunk lists are stored in the CLS rather than the CAS and we don't have an api method to request bulk existence checking. We could do it one by one for `blobs > maxChunkSize` (requiring FindMissing to also to call GetCapabilities) or we could add a custom api for bulk existence checking chunk lists.

pkg/blobstore/grpcservers/content_addressable_storage_server.go
meroton-benjamin: It makes sense to renew the lifetime of the chunk lists when calling FMB for their blob but if we replace the nil CLS with an ErrorBlobAccess how would we know which mode of operation to use?

pkg/blobstore/chunklistvalidating/fake_blob_access_test.go
meroton-benjamin: Eh we sort of could but it becomes noisy, for some of the tests a straight up `EXPECT().Return()` construct flows well but for the more complicated tests such as the ones where we want to verify chunking and that files have been properly touched it becomes very noisy.

For those tests we could use `.EXPECT().DoAndReturn().AnyTimes()` to implement pretty much the same thing as in fakeBlobAccess already does but that would basically be the same thing as fakeBlobAccess just with lambdas instead of functions.

If you prefer that anyway I can modify the code such that it uses gomock for consistency but at least in my mind fakeBlobAccess was easier to understand.

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
meroton-benjamin: Sure, I'll add the requirement when getting the capabilities above, erroring if minChunkSizeBytes is outside of [64, maxMessageSizeBytes/2]

pkg/capabilities/chunking_provider.go
meroton-benjamin: I don't see why not, I'll add it

cmd/bb_storage/main.go
meroton-benjamin: I don't need to add it because Capabilities are propagated from the CAS and not from the CLS. Do you want us to propagate the parameters from the CLS instead?

cmd/bb_storage/main.go
```diff
@@ -84,6 +84,27 @@ func main() {
...
+				configuration.ChunkListStorage,
+				blobstore_configuration.NewCLSBlobAccessCreator(
+					contentAddressableStorageInfo,
+					grpcClientFactory,
+					int(configuration.MaximumMessageSizeBytes),
+				),
+				grpcClientFactory,
+			)
+			if err != nil {
+				return util.StatusWrap(err, "Failed to create Chunk Map")
```
EdSchouten: Failed to create Chunk List Storage?

pkg/blobstore/grpcservers/content_addressable_storage_server.go
EdSchouten: Maybe good to at least add a `// TODO: Require that s.chunkListStorage is non-null once we require chunking.` ?

pkg/blobstore/grpcservers/content_addressable_storage_server.go
```diff
@@ -56,6 +58,36 @@ func (s *contentAddressableStorageServer) FindMissingBlobs(ctx context.Context,
...
+		for _, partialDigest := range in.BlobDigests {
+			if partialDigest.GetSizeBytes() > int64(maxChunkSize) {
+				digest, err := digestFunction.NewDigestFromProto(partialDigest)
+				if err != nil {
+					return nil, err
+				}
+				bigBlobDigests.Add(digest)
+			}
+		}
+		_, _ = s.chunkListStorage.FindMissing(ctx, bigBlobDigests.Build())
```
EdSchouten, line 88: Sure we don't need any error/results handling here...?

pkg/blobstore/configuration/cls_blob_access_creator.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (clsBlobAccessCreator) GetDefaultCapabilitiesProvider() capabilities.Provider {
+	return capabilities.NewStaticProvider(&remoteexecution.ServerCapabilities{})
+}
+
+func (bac *clsBlobAccessCreator) NewCustomBlobAccess(terminationGroup program.Group, configuration *pb.BlobAccessConfiguration, nestedCreator NestedBlobAccessCreator) (BlobAccessInfo, string, error) {
+	switch backend := configuration.Backend.(type) {
+	case *pb.BlobAccessConfiguration_ChunkListValidating:
+		if bac.contentAddressableStorage == nil {
+			return BlobAccessInfo{}, "", status.Error(codes.InvalidArgument, "Action Cache completeness checking can only be enabled if a Content Addressable Storage is configured")
```
EdSchouten: s/Action Cache completeness checking/.../ ?

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,371 @@
...
+	if err != nil {
+		return nil, util.StatusWrap(err, "Unable to GetCapabilities to determine chunking parameters")
+	}
+
+	params := capabilities.GetCacheCapabilities().GetRepMaxCdcParams()
+	if params == nil {
+		return nil, status.Error(codes.Unimplemented, "This backend only supports upstream servers with rep max cdc support.")
+	}
+	if params.MinChunkSizeBytes < 64 {
+		return nil, status.Errorf(codes.Internal, "MinChunkSizeBytes was %d but a minimum of 64 is required.", params.MinChunkSizeBytes)
```
EdSchouten: "RepMaxCDC minimum chunk size was %d bytes, but a minimum of 64 bytes is required" ?

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,371 @@
...
+// implementation.
+func (ba *chunkListValidatingBlobAccess) getValidChunkingParameters(ctx context.Context, instanceName digest.InstanceName) (*remoteexecution.RepMaxCdcParams, error) {
+	capabilities, err := ba.BlobAccess.GetCapabilities(ctx, instanceName)
+	if err != nil {
+		return nil, util.StatusWrap(err, "Unable to GetCapabilities to determine chunking parameters")
+	}
+
+	params := capabilities.GetCacheCapabilities().GetRepMaxCdcParams()
+	if params == nil {
+		return nil, status.Error(codes.Unimplemented, "This backend only supports upstream servers with rep max cdc support.")
```
EdSchouten: s/rep max cdc/RepMaxCDC/

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,371 @@
...
+
+// Fetch the chunking parameters from the GetCapabilities
+// implementation.
+func (ba *chunkListValidatingBlobAccess) getValidChunkingParameters(ctx context.Context, instanceName digest.InstanceName) (*remoteexecution.RepMaxCdcParams, error) {
+	capabilities, err := ba.BlobAccess.GetCapabilities(ctx, instanceName)
+	if err != nil {
+		return nil, util.StatusWrap(err, "Unable to GetCapabilities to determine chunking parameters")
+	}
+
+	params := capabilities.GetCacheCapabilities().GetRepMaxCdcParams()
```
EdSchouten: s/GetCacheCapabilities()/CacheCapabilities/

In the Buildbarn tree I only try to use `.Get*()` in cases where the parent may be nil. That way it's sort of self-documenting what can be nil or not. In this case the top-level message is always non-nil.

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,371 @@
...
+	if params == nil {
+		return nil, status.Error(codes.Unimplemented, "This backend only supports upstream servers with rep max cdc support.")
+	}
+	if params.MinChunkSizeBytes < 64 {
+		return nil, status.Errorf(codes.Internal, "MinChunkSizeBytes was %d but a minimum of 64 is required.", params.MinChunkSizeBytes)
+	}
+	maxMinChunkSize := (ba.maximumMessageSizeBytes + 1) / 2
+	if params.MinChunkSizeBytes > uint64(maxMinChunkSize) {
+		return nil, status.Errorf(codes.Internal, "MinChunkSizeBytes was %d but a maximum of %d is supported with the configured maximum message size.", params.MinChunkSizeBytes, maxMinChunkSize)
```
EdSchouten: Please make sure to remove trailing dots from the error messages.

cmd/bb_storage/main.go
meroton-benjamin: ack

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
meroton-benjamin: ack

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
meroton-benjamin: ack

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
meroton-benjamin: ack

pkg/blobstore/chunklistvalidating/fake_blob_access_test.go
meroton-benjamin: Do you have any input on if we want to turn fakeBlobAccess into a lambda construct with gomock for consistency or if it's fine as is?

pkg/blobstore/configuration/cls_blob_access_creator.go
meroton-benjamin: Replaced with "Chunk list validation can only be enabled if a Content Addressable Storage is configured"

pkg/blobstore/grpcservers/content_addressable_storage_server.go
meroton-benjamin: We perform the call towards the chunk list storage in order to prevent the chunk list storage itself from being dropped from the cache if they exist. As we've yet to make the chunking mandatory semantically the blob is not missing even if it's chunk list is missing.

With regards to errors, I'm not sure how we would treat errors at this point.

pkg/blobstore/grpcservers/content_addressable_storage_server.go
meroton-benjamin: done

PR #347, opened 2026-06-22
Replace CAS with CS and CLS

meroton-benjamin:
This commit builds on top of our split and splice blob support to make
it a mandatory first class feature in Buildbarn. With this commit the
Content Addressable Storage (CAS) is created from two Storage
configurations that work in tandem. A Chunk Storage (CS) which is
content addressed and contains chunks of blobs, and a Chunk List Storage
(CLS) which is addressed by a blob digest and contains a manifest
describing the chunks that make up the blob.

All api calls are automatically translated to use Chunk Lists created
with RepMaxCDC. Effectively this means that large blobs no longer exists
in the storage layer, individual chunks of the large blobs are in turn
deduplicated in such a manner that the chunks are stored only once.

The automatic translation makes certain that clients that are not cdc
aware can still continue to use the storage backend without performing
any changes. Clients which support RepMaxCDC also gets a significant
reduction in the amount of blobs to transfer as they only need to
transfer modified chunks rather than the entire blob.


cmd/bb_storage/main.go
```diff
@@ -54,34 +58,88 @@ func main() {
...
+			if casConfiguration.ChunkStorage == nil {
+				return status.Error(codes.InvalidArgument, "The Chunk Storage is a mandatory part of the Content Addressable Storage.")
+			}
+			if casConfiguration.ChunkListStorage == nil {
+				return status.Error(codes.InvalidArgument, "The Chunk List Storage is a mandatory part of the Content Addressable Storage.")
+			}
+
+			var parameterCache *cdc.TTLCache[cdc.Parameters]
+			if casConfiguration.ContentDefinedChunkingParameterCache != nil {
+				parameterCacheConfiguraiton := casConfiguration.ContentDefinedChunkingParameterCache
```
EdSchouten: Typo: Configuraiton

cmd/bb_storage/main.go
```diff
@@ -54,34 +58,88 @@ func main() {
...
+				evictionSet, err := eviction.NewSetFromConfiguration[string](parameterCacheConfiguraiton.CacheReplacementPolicy)
+				if err != nil {
+					return err
+				}
+				parameterCache = cdc.NewTTLCache[cdc.Parameters](
```
EdSchouten: So this is a cache that maps instance names to CDC parameters, right? I don't think it's correct to create such a cache globally. If you use `DemultiplexingBlobAccess`, you can rewrite instance names. Wouldn't that lead to potential collisions?

Also in the case of centralized storage nodes I don't think it makes sense to have any caching of CDC parameters. There's nothing to cache, as the storage node would just a single configuration globally. It's only the gRPC client backend that needs a cache.

Maybe better to just extend BlobAccess to have a new `GetCDCParameters()` method or something? Then let individual storage backends be responsible for caching this information (or not).

cmd/bb_storage/main.go
```diff
@@ -54,34 +58,88 @@ func main() {
...
+					clock.SystemClock,
+					evictionSet,
+					int(parameterCacheConfiguraiton.GetCacheSize()),
+					parameterCacheConfiguraiton.CacheDuration.AsDuration(),
```
EdSchouten: Missing `.CacheDuration.CheckValid()`?

cmd/bb_storage/main.go
```diff
@@ -54,34 +58,88 @@ func main() {
...
+				blobstore_configuration.NewCSBlobAccessCreator(
 					grpcClientFactory,
 					int(configuration.MaximumMessageSizeBytes),
 					zstdPool,
 				),
 				grpcClientFactory,
 			)
 			if err != nil {
-				return util.StatusWrap(err, "Failed to create Content Addressable Storage")
+				return util.StatusWrap(err, "Failed to create Content Addressable Storage: Failed to create Chunk Storage")
```
EdSchouten: "Failed to create Chunk Storage for Content Addressable Storage"?

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,309 @@
...
+
+	// Check that all referenced chunks are present in storage.
+	userChunks, err = ba.flattenChunks(ctx, params, userChunks)
+	if err != nil {
+		return status.Error(codes.NotFound, "At least one chunk is missing from storage.")
+	}
+
+	// Chunk list is marked for validation bypass, push it directy to
+	// downstream blob store.
+	if cdc.ChunkListValidationBypassed(ctx) {
```
EdSchouten, line 164: To me the notion of whether a chunk list is known to be valid isn't really a properly of the calling context. It's more of a property of the chunk list that's passed in. Maybe better to either change the signature of `BlobAccess.Put()`, or add this to `buffer.Buffer`?

I take it that this logic is added to make sure that if a client performs a legacy `Write()` for a large object, that the frontend doesn't re-read the objects just to make sure that the ChunkList is valid, right? If so, how does this actually relate to `buffer.Source`? Maybe we should just treat ChunkLists created by bb-frontend itself as being `buffer.BackendProvided()`, and use that to skip validation?

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -199,6 +200,29 @@ message BlobAccessConfiguration {
     // value. When gRPC calls are timed out a `DEADLINE_EXCEEDED` error
     // code will be returned.
     DeadlineEnforcingBlobAccess deadline_enforcing = 28;
+
+    // Validate that chunk list requests fulfill the api specification.
+    // Chunk List Storage (CLS) is used to implement the SplitBlob and
```
EdSchouten: s/api/API/

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -199,6 +200,29 @@ message BlobAccessConfiguration {
...
+    // SpliceBlob methods from the REv2 api. The protocol has several
+    // demands on the state of the Content Addressable Storage (CAS)
+    // after those methods have been called.
+    //
+    // SplitBlob requires that that the blob as well as all chunks of
```
EdSchouten: s/that that/that/

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -199,6 +200,29 @@ message BlobAccessConfiguration {
...
+    // renewed and that the resulting chunk list composes into the blob.
+    // This is validated by the Get path by splitting the Blob on demand
+    // if required.
+    //
+    // SpliceBlob requires that the blob as well as all chunks of the
+    // blob are present in storage, they have their lifetime renewed and
+    // that the supplied chunk list composes into the blob. Notably it
+    // does not require the chunks to follow any particular chunking
+    // algorithm but our implementation ensures that after any call a
+    // proper rep max cdc chunk list is verified even if the caller
```
EdSchouten: s/rep max cdc/RepMaxCDC/

pkg/proto/configuration/bb_storage/bb_storage.proto
```diff
@@ -57,9 +58,9 @@ message ApplicationConfiguration {
   // Common configuration options that apply to all Buildbarn binaries.
   buildbarn.configuration.global.Configuration global = 9;
 
-  // Optional: Blobstore configuration for the Content Addressable
-  // Storage (CAS).
-  ScannableBlobAccessConfiguration content_addressable_storage = 17;
+  // Optional: Blobstore configurations for the Content
+  // AddressableContentAddressa Storage (CAS).
```
EdSchouten: Typo :)

pkg/proto/configuration/bb_storage/bb_storage.proto
```diff
@@ -109,6 +110,29 @@ message ApplicationConfiguration {
   // ByteStream server, and enables ZSTD compression for ByteStream
   // operations where the server supports it.
   buildbarn.configuration.zstd.PoolConfiguration zstd_pool = 21;
+
+  // Was 'chunk_list_storage'. Has been moved into the
+  // content_addressable_storage.
+  reserved 22;
```
EdSchouten: This can be removed, right?

cmd/bb_storage/main.go
```diff
@@ -54,34 +58,88 @@ func main() {
...
+		var chunkListStorage blobstore.BlobAccess
 		if configuration.ContentAddressableStorage != nil {
-			info, authorizedBackend, allAuthorizers, err := newScannableBlobAccess(
+			casConfiguration := configuration.ContentAddressableStorage
+			if casConfiguration.ChunkStorage == nil {
+				return status.Error(codes.InvalidArgument, "The Chunk Storage is a mandatory part of the Content Addressable Storage.")
+			}
+			if casConfiguration.ChunkListStorage == nil {
+				return status.Error(codes.InvalidArgument, "The Chunk List Storage is a mandatory part of the Content Addressable Storage.")
+			}
```
EdSchouten: Is this really a necessary requirement? I can imagine that for frontends that implement the full REv2 API it's necessary that both are provided. But for individual shards of our storage backends there is no requirement that each node provides both a CS and CLS.

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -642,6 +674,18 @@ message LocalBlobAccessConfiguration {
...
+  // The chunking parameters advertised via GetCapabilities, setting
+  // this announces support for the SplitBlob and SpliceBlob api calls.
+  //
+  // Adding the chunking parameters does not make the underlying
+  // blobstore enforce the chunking parameters or implement the split
+  // and splice blob. There must be a configured Chunk List Storage with
+  // an outer ChunkValidatingBlobAccess that has view of the entire
+  // Content Addressable Storage (CAS).
+  //
+  // This option is only supported for the CLS.
```
EdSchouten: This option is only supported for the CS?

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -936,4 +980,15 @@ message GrpcBlobAccessConfiguration {
   // used if the server supports it. Has no effect for other storage
   // types (AC, ICAS, etc.).
   bool enable_compression = 2;
+
+  // Optional: In memory cache for getting the capabilities for a
+  // specific instance name. Can be set to a relatively low value like
+  // '60s'.
```
EdSchouten: This is only used by the CS?

cmd/bb_copy/main.go
```diff
@@ -42,7 +42,7 @@ func main() {
 
 		grpcClientFactory := grpc.NewBaseClientFactory(grpc.BaseClientDialer, nil, nil, nil)
 
-		blobAccessCreator := blobstore_configuration.NewCASBlobAccessCreator(
+		blobAccessCreator := blobstore_configuration.NewCSBlobAccessCreator(
```
EdSchouten, line 45: Keep in mind that bb_copy can, for example, be used to make backups of actions. This means that for bb_copy it's not sufficient to only use a Chunk Store. It should also preserve/expand chunk lists, etc.

cmd/bb_replicator/main.go
```diff
@@ -33,7 +33,7 @@ func main() {
 			return util.StatusWrap(err, "Failed to apply global configuration options")
```
EdSchouten, line 33: Now that we only deal with chunks and no longer do any streaming, do you think we still need bb_replicator? Maybe it's fine to just throw it out altogether, and let bb_frontend do the fixups directly?

cmd/bb_storage/main.go
EdSchouten: This concern still holds. :-)

pkg/blobstore/buffer/cas_chunk_concatenating_buffer.go
```diff
@@ -0,0 +1,257 @@
...
+	chunkList   ChunkList
+	source      Source
+}
+
+// NewValidatedCASChunkListConcatenatingBuffer creates an optimized
+// buffer for an object stored in the Content Addressable Storage
+// composed of a validated list of chunks. Because the chunk list is
+// treated as correct, validation is limited to validating any
+// underlying chunks read allowing for random access reads.
+func NewValidatedCASChunkListConcatenatingBuffer(ctx context.Context, blobDigest digest.Digest, chunkList ChunkList, chunkGetter ChunkGetter, source Source) Buffer {
```
EdSchouten: So I take it that in the final version of your PR this implementation will be gone, right? It shouldn't be necessary to create `buffer.Buffer` objects that have concatenated data.

pkg/blobstore/cdc/bypass.go
```diff
@@ -0,0 +1,22 @@
...
+
+import "context"
+
+type chunkListValidationBypassKey struct{}
+
+// NewContextWithChunkListValidationBypass creates a derived context
+// that signals downstream storage layers that the chunk list has
+// already been validated (or was freshly generated) and does not need
+// expensive re-validation.
+func NewContextWithChunkListValidationBypass(ctx context.Context) context.Context {
```
EdSchouten, line 11: Instead of using `context.Context` for this, why can't we pass through this information as regular function arguments?

pkg/blobstore/cdc/cas_chunking_blob_access.go
```diff
@@ -0,0 +1,185 @@
...
+	chunkGetter             buffer.ChunkGetter
+	chunkListFetcher        buffer.ChunkListFetcher
+	chunkListStorage        blobstore.BlobAccess
+	maximumMessageSizeBytes int
+}
+
+// NewCasChunkingBlobAccess creates a cas blob access configuration that
+// constructs large case objects from the chunks described in the chunk
+// list.
+func NewCasChunkingBlobAccess(chunkStorage, chunkListStorage blobstore.BlobAccess, chunkListFetcher buffer.ChunkListFetcher, maximumMessageSizeBytes int) blobstore.BlobAccess {
```
EdSchouten: This type would also be removed, right?

pkg/blobstore/cdc/cdc_parameters.go
```diff
@@ -0,0 +1,40 @@
...
+	capabilities, err := capabilitiesProvider.GetCapabilities(ctx, instanceName)
+	if err != nil {
+		return Parameters{}, util.StatusWrap(err, "Unable to GetCapabilities to determine chunking parameters")
+	}
+
+	params := capabilities.CacheCapabilities.GetRepMaxCdcParams()
+	if params == nil {
+		return Parameters{}, status.Error(codes.Unimplemented, "This server implementation is only compatible with RepMaxCDC")
+	}
+	if params.MinChunkSizeBytes < 64 {
```
EdSchouten: Just as a safety belt, also require that MaxChunkSizeBytes isn't larger than some big constant? The same thing with the HorizonSizeBytes, which should also be limited. Maybe better to also require a MinChunSizeBytes of more than 64? Seems pretty evil if clients can be tricked into using a ridiculously small chunk size that would cause very high memory usage.

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	// Return upstream split result if complete.
+	return ba.getComplete(ctx, d)
+}
+
+func (ba *chunkListValidatingBlobAccess) matchesStoredChunkList(ctx context.Context, d digest.Digest, userResponse *remoteexecution.SplitBlobResponse) bool {
+	existingMsg, err := ba.BlobAccess.Get(ctx, d).ToProto(&remoteexecution.SplitBlobResponse{}, ba.maximumMessageSizeBytes)
+	if err != nil {
+		return false
```
EdSchouten, line 118: This suppresses errors. Is that intentional?

pkg/blobstore/configuration/new_blob_access.go
```diff
@@ -352,11 +356,26 @@ func (nc *simpleNestedBlobAccessCreator) newNestedBlobAccessBare(configuration *
...
+				capabilitiesProvider,
+				capabilities.NewStaticProvider(&remoteexecution.ServerCapabilities{
+					CacheCapabilities: &remoteexecution.CacheCapabilities{
+						SplitBlobSupport:  true,
+						SpliceBlobSupport: true,
+						RepMaxCdcParams:   chunkingParameters,
+					},
+				}),
+			})
+		}
```
EdSchouten, line 374: I think that instead of doing this, we should:

- Rename `GetDefaultCapabilitiesProvider()` to something like `GetLeafCapabilitiesProvider()`. Or `GetBaseCapabilitiesProvider()`. That way it's more explicit that it should only be used by 'leaf'/'base' data stores.
- Extend that method to take a `ChunkingParameters` argument.
- For the non-CS data stores, let it fail if chunking parameters are provided.
- For the CS, do the inverse: require that chunking parameters are given.

pkg/blobstore/configuration/new_blob_access.go
```diff
@@ -645,3 +660,47 @@ func NewCASAndACBlobAccessFromConfiguration(terminationGroup program.Group, conf
 
 	return contentAddressableStorage.BlobAccess, actionCache.BlobAccess, nil
 }
+
+// NewCASBlobAccessFromConfiguration is a convenience function to create
+// BlobAccessInfo for Content Addressable Storage (CAS) and BlobAccess
+// objects for its constituent blobstores Chunk Storage (CS) and Chunk
+// List Storage (CLS).
+func NewCASBlobAccessFromConfiguration(terminationGroup program.Group, configuration *pb.ContentAddressableStorage, grpcClientFactory grpc.ClientFactory, maximumMessageSizeBytes int, zstdPool bb_zstd.Pool) (contentAddressableStorage BlobAccessInfo, chunkStorage, chunkListStorage blobstore.BlobAccess, err error) {
```
EdSchouten: This also needs to be removed, right?

snakethatlovesstaticlibs:
Closes #360 

edit: ah that didn't too anything, oh well

meroton-benjamin:
Replaced by #366

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
meroton-benjamin: fixed

Issue #350, opened 2026-06-27
ghcr.io/buildbarn/bb-storage points to wrong binary


BoleynSu:
```
$ podman run --rm -it ghcr.io/buildbarn/bb-storage:20260623T170048Z-b4bd798
Trying to pull ghcr.io/buildbarn/bb-storage:20260623T170048Z-b4bd798...
Getting image source signatures
Copying blob 688513194d7a done   | 
Copying blob 3214acf345c0 done   | 
Copying blob efa9d1d5d3a2 done   | 
Copying blob a62778643d56 done   | 
Copying blob 7c12895b777b done   | 
Copying blob bfb59b82a9b6 done   | 
Copying blob 5664b15f108b done   | 
Copying blob 0bab15eea81d done   | 
Copying blob 4aa0ea1413d3 done   | 
Copying blob da7816fa955e done   | 
Copying blob 9aee425378d2 done   | 
Copying blob af044092d7eb done   | 
Copying config 4a0fd46c05 done   | 
Writing manifest to image destination
2026/06/27 13:41:48 Usage: partition_ephemeral_disks partition_ephemeral_disks.jsonnet
```

PR #351, opened 2026-06-27
Fix container publishing of partition_ephemeral_disks

BoleynSu:


BoleynSu:
Fixes https://github.com/buildbarn/bb-storage/issues/350

PR #353, opened 2026-07-03
Generate test key material at runtime instead of embedding it

ma-oli:
The RSA, ECDSA and Ed25519 private keys, and the TLS certificate/key pairs, were hard-coded in test files. Security scanners flag any embedded private key regardless of it being throwaway test material.

Generate all key material at test time instead: the JWT tests now generate a key and sign each algorithm at runtime (flipping a byte for the invalid cases), and the TLS tests generate self-signed certificate and key pairs via a shared helper, using a realistic validity window (NotBefore = now - 10m, NotAfter = now + 1h). No private keys remain in source.


EdSchouten:
Using static keys and tokens is intentional. See https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html .

PR #356, opened 2026-07-31
fix: use unix.O_RDONLY in darwin oflagsearch

francine-blanc:
`O_SEARCH` and `O_PATH` give search-only directory permissions on FreeBSD and Linux OS respectively (NB. for Linux this is more restrictive - the handle produced cannot be used for any regular R/W/IO).

`O_SEARCH` is not supported by darwin, so `unix.O_SEARCH` does not exist for darwin build tags in Go; its presence caused a compile failure when attempting to build bb-storage for macOS.

This PR therefore changes the flag to `O_RDONLY` which although is more permissive (it grants search and read access) is the closest equivalent.

### DIFF
--- a/pkg/filesystem/local_directory_darwin.go
+++ b/pkg/filesystem/local_directory_darwin.go
@@ -16,7 +16,7 @@ import (
 // rawDeviceNumber is the equivalent of POSIX dev_t.
 type rawDeviceNumber = int32
 
-const oflagSearch = unix.O_SEARCH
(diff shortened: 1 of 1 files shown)


EdSchouten:
O_SEARCH is supported by all recent versions of Darwin. It's just that golang.org/x/sys/unix is somewhat out of date. We apply a local patch in bb-storage to add this specific constant.

PR #362, opened 2026-08-20
Add ZSTD `ReadOffset` support

### LINKED ISSUE #360: ByteStream ZSTD reads ignore read_offset, corrupting resumed downloads
Full disclosure, AI was used to find this in our dev system, but I looked at the finding before filing

 ## Summary

The ZSTD implementation of `ByteStream.Read` does not honor `ReadRequest.read_offset`.

When a compressed download is interrupted, Bazel retries with a nonzero `read_offset` representing the uncompressed bytes already received. Buildbarn restarts the ZSTD stream from the beginning of the blob. Bazel then appends the restarted content to its partial output, producing an oversized file and a digest mismatch.

## Affected version

Observed with:

  - bb-storage commit: `10acc76a8295d86f64baa8485c8e616bce0d53ca`
  - Bazel: `9.2.0`
  - `--remote_cache_compression` enabled
  - Buildbarn configured with `supportedCompressors: ['ZSTD']`

The affected path appears to have been introduced by [PR #258](https://github.com/buildbarn/bb-storage/pull/258).

 ## Relevant code

The uncompressed path correctly passes `in.ReadOffset` to `ToChunkReader`:

https://github.com/buildbarn/bb-storage/blob/10acc76a8295d86f64baa8485c8e616bce0d53ca/pkg/blobstore/grpcservers/byte_stream_server.go#L47-L62

The ZSTD path does not use `in.ReadOffset` and passes the complete blob directly to the encoder:

https://github.com/buildbarn/bb-storage/blob/10acc76a8295d86f64baa8485c8e616bce0d53ca/pkg/blobstore/grpcservers/byte_stream_server.go#L64-L72

This behavior was introduced as part of: https://github.com/buildbarn/bb-storage/pull/258

## Production evidence

A remote cache download expected:

```
  e31efa8db7403d522c31aa12a9be8c1803a76fe8968ce0acad5aa389476bcbb8/709569128
```

Retries for that same digest produced several different oversized outputs:

```
  811c07177cada0ca514ba52b1de17b14f8a397ce45389ebef8956fcd6f7a1778/1918315112
  b2af.../1943874152
  a9f6.../2397776488
```

The varying sizes are consistent with partial downloads followed by one or more complete copies being appended after retries.

Bazel reported:

```
  Remote Cache: Output download failed:
  Expected digest 'e31efa8d.../709569128'
  does not match received digest '811c0717.../1918315112'.
```

Bazel rejected the cache result and rebuilt the output, so the build completed successfully, but this causes excessive network transfer and lost cache hits.

## Reproduction outline

  1. Configure Buildbarn with `supportedCompressors: ['ZSTD']`.
  2. Store a sufficiently large CAS blob.
  3. Start a compressed `ByteStream.Read` with `read_offset = 0`.
  4. Consume and decompress the first `N` bytes, then interrupt the RPC.
  5. Retry the same resource with `read_offset = N`.
  6. Decompress the retry response and append it to the first `N` bytes.
  7. Observe that the retry contains the full blob from offset zero, producing `N + blob` size bytes and a digest mismatch.

Multiple interruptions can append multiple partial or complete copies.

## Expected behavior

For a compressed resource, a request with nonzero `read_offset` should return a new ZSTD stream containing the logical uncompressed blob starting at that offset.

## Actual behavior

The response contains a ZSTD stream for the entire blob starting at offset zero.

## Suggested fix

Apply `in.ReadOffset` to the **uncompressed blob** before passing data into the ZSTD encoder.

Please also add regression coverage for:

  - ZSTD reads with `read_offset = 0`
  - ZSTD reads with a nonzero offset
  - interrupted downloads resumed through multiple requests
  - verifying the concatenated decompressed output matches the original digest and size

## Workaround

  Disabling Bazel remote-cache compression avoids the affected branch: `--noremote_cache_compression`

snakethatlovesstaticlibs:
This should close #360

AI was used to create this code

### DIFF
--- a/pkg/blobstore/grpcservers/byte_stream_server.go
+++ b/pkg/blobstore/grpcservers/byte_stream_server.go
@@ -35,12 +35,18 @@ func NewByteStreamServer(blobAccess blobstore.BlobAccess, readChunkSize int, zst
 }
 
 func (s *byteStreamServer) Read(in *bytestream.ReadRequest, out bytestream.ByteStream_ReadServer) error {
-	if in.ReadLimit != 0 {
-		return status.Error(codes.Unimplemented, "This service does not support downloading partial files")
-	}
 	digest, compressor, err := digest.NewDigestFromByteStreamReadPath(in.ResourceName)
 	if err != nil {
 		return err
+	}
+	if in.ReadLimit != 0 {
+		if compressor != remoteexecution.Compressor_IDENTITY {
+			// REAPI requires non-zero read limits on compressed ByteStream
+			// reads to be rejected with INVALID_ARGUMENT.
+			// https://github.com/bazelbuild/remote-apis/blob/becdd8f9ff811df88a22d3eadd6341753d51d167/build/bazel/remote/execution/v2/remote_execution.proto#L313-L317
+			return status.Error(codes.InvalidArgument, "Read limits are not permitted for compressed blobs")
+		}
+		return status.Error(codes.Unimplemented, "This service does not support downloading partial files")
 	}
 	ctx := out.Context()
 	switch compressor {
@@ -69,7 +75,21 @@ func (s *byteStreamServer) Read(in *bytestream.ReadRequest, out bytestream.ByteS
 			return status.Errorf(codes.ResourceExhausted, "Failed to acquire ZSTD encoder: %v", err)
 		}
 		defer encoder.Close()
-		return b.IntoWriter(encoder)
+
+		r := b.ToChunkReader(in.ReadOffset, s.readChunkSize)
+		defer r.Close()
+		for {
+			chunk, err := r.Read()
+			if err == io.EOF {
+				return nil
+			}
+			if err != nil {
+				return err
+			}
+			if _, err := encoder.Write(chunk); err != nil {
+				return err
+			}
+		}
 	default:
 		return status.Errorf(codes.Unimplemented, "This service does not support downloading compression type: %s", compressor)
 	}
--- a/pkg/blobstore/grpcservers/byte_stream_server_test.go
+++ b/pkg/blobstore/grpcservers/byte_stream_server_test.go
@@ -27,6 +27,32 @@ import (
 	"go.uber.org/mock/gomock"
 )
 
+func readAndDecompressZSTD(t *testing.T, ctx context.Context, client bytestream.ByteStreamClient, resourceName string, readOffset int64) []byte {
+	t.Helper()
+	req, err := client.Read(ctx, &bytestream.ReadRequest{
+		ResourceName: resourceName,
+		ReadOffset:   readOffset,
+	})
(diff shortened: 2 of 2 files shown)


pkg/blobstore/grpcservers/byte_stream_server_test.go
```diff
@@ -202,6 +228,68 @@ func TestByteStreamServer(t *testing.T) {
...
+	})
+
+	t.Run("ReadIdentityWithReadLimit", func(t *testing.T) {
+		req, err := client.Read(ctx, &bytestream.ReadRequest{
+			ResourceName: "blobs/09f7e02f1290be211da707a266f153b3/5",
+			ReadLimit:    10,
+		})
+		require.NoError(t, err)
+		_, err = req.Recv()
+		testutil.RequireEqualStatus(t, status.Error(codes.Unimplemented, "This service does not support downloading partial files"), err)
```
snakethatlovesstaticlibs: Hmm we don't need a new test for this old behavior

meroton-benjamin:
This bug was identified and fixed in pr #347 as well but as that is a much larger pull request it is likely to need some more time before it is ready to merge. @EdSchouten Any opinion on merging the specific bug fix separately?

pkg/blobstore/grpcservers/byte_stream_server.go
```diff
@@ -35,13 +35,19 @@ func NewByteStreamServer(blobAccess blobstore.BlobAccess, readChunkSize int, zst
 }
 
 func (s *byteStreamServer) Read(in *bytestream.ReadRequest, out bytestream.ByteStream_ReadServer) error {
-	if in.ReadLimit != 0 {
```
EdSchouten, line 38: What's up with this change? I don't think we have any code for respecting this, even for the uncompressed case.

snakethatlovesstaticlibs: This change leaves existing behavior as is for the uncompressed case, but changes the return value to `InvalidArgument` for the compressed case, because the spec wants it that way

```
// When downloading compressed blobs:
// * `ReadRequest.read_offset` refers to the offset in the uncompressed form
//   of the blob.
// * Servers MUST return `INVALID_ARGUMENT` if `ReadRequest.read_limit` is
//   non-zero.
```
from https://github.com/bazelbuild/remote-apis/blob/becdd8f9ff811df88a22d3eadd6341753d51d167/build/bazel/remote/execution/v2/remote_execution.proto#L313-L317

It probably doesn't matter since clients likely won't send a read limit for the compressed case, but I figured I'd fix it anyway. Please let me know if you want it reverted and I can do that too

EdSchouten (approved):
Will merge once CI is happy.

snakethatlovesstaticlibs:
Awesome, thank you @EdSchouten! Looks like the jobs passed

PR #363, opened 2026-08-22
Add a generic message reader interface

meroton-benjamin:
Adds a generic message reader interface for reading proto messages from storage. Many Buildbarn components depend on blobstore.BlobAccess only as a source of proto.Message and this interface allows us to sever that dependency in components which are otherwise not interested in the blobstore.BlobAccess interface.


pkg/storage/message_reader.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// MessageReader can be used to read a proto.Message from storage and
+// return its contents.
+type MessageReader[T proto.Message] interface {
+	// Return a parsed message from storage.
+	//
+	// Implementations may use the provided message to store the
+	// unmarshalled message.
+	ReadMessage(ctx context.Context, d digest.Digest, msg T) (T, error)
```
EdSchouten: What's with the `msg T` argument? Normally you only provide that to `ToProto()` to indicate the type you're interested in. For MessageReader that should already be implied by the interface itself.

meroton-benjamin: msg T argument removed in favor of adding an allocator to the constructor.

meroton-benjamin:
State of latest version:

- `MessageReader` moved to `pkg/cas`
-  Non-CAS uses of the `MessageReader` interface has been removed.
- The storage message has been removed from the function signature and instead the constructor takes an allocator.

Looking at places which reads messages from the CAS in bb-storage, there only seems to be `CompletenessCheckingBlobAccess` and `NestedBlobReplicator`. In `CompletenessCheckingBlobAccess`, it actually only fetches tree objects in a streaming fashion from the CAS while `NestedBlobReplicator` fetches tree objects in a streaming fashion and Directory/Action objects in a non streaming fashion.

In the last commit here `Use message reader in bb-copy` a resolution to this has been done by changing the `NestedBlobReplicator` constructor to:

`func NewNestedBlobReplicator(replicator CASReplicator, maximumMessageSizeBytes int, actionReader cas.MessageReader[*remoteexecution.Action], directoryReader cas.MessageReader[*remoteexecution.Directory], treeReader BlobStreamReader, digestKeyFormat digest.KeyFormat)`

With:

```
type CASReplicator interface {
	ReplicateMultiple(ctx context.Context, digests digest.Set) error
}

type BlobStreamReader interface {
	ReadStream(ctx context.Context, d digest.Digest) (io.ReadCloser, error)
}
```

Which allows me to tie up the loop. Does this look like a good signature and if so I can consume the `BlobStreamReader` interface from  `CompletenessCheckingBlobAccess` as well.

pkg/blobstore/blob_access_message_reader.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"context"
+
+	"github.com/buildbarn/bb-storage/pkg/cas"
+	"github.com/buildbarn/bb-storage/pkg/digest"
+	"google.golang.org/protobuf/proto"
+)
+
+type blobAccessMessageReader[T proto.Message] struct {
```
EdSchouten, line 11: If you change this line to:
```go
type blobAccessMessageReader[T any, TPtr interface {
        *T
        proto.Message
}] struct {
```
Then you can use `T` to refer to a concrete message type, and `TPtr` whenever you need a pointer to one. That way you can get rid of the allocator function.

pkg/cas/message_reader.go
```diff
@@ -0,0 +1,16 @@
...
+	"context"
+
+	"github.com/buildbarn/bb-storage/pkg/digest"
+
+	"google.golang.org/protobuf/proto"
+)
+
+// MessageReader can be used to read a parsed proto.Message from the
+// Content Addressable Storage.
+type MessageReader[T proto.Message] interface {
```
EdSchouten, line 13: If you wanted to, you could change this `proto.Message` to `any`. That way you get rid of the dependency on the Protobuf package. Though in this case I'm ambivalent.

pkg/blobstore/replication/nested_blob_replicator.go
```diff
@@ -8,25 +8,40 @@ import (
...
 
+// CASReplicator replicates blobs to the destination Content
+// Addressable Storage.
+type CASReplicator interface {
+	ReplicateMultiple(ctx context.Context, digests digest.Set) error
+}
+
+// BlobStreamReader provides a stream to the raw bytes of a blob.
+type BlobStreamReader interface {
+	ReadStream(ctx context.Context, d digest.Digest) (io.ReadCloser, error)
```
EdSchouten, line 26: I'm just thinking: aren't we going to need an interface like this on the NaiveBuildDirectory side to download files? If so, maybe this should also live in `pkg/cas`?

pkg/blobstore/blob_access_message_reader.go
meroton-benjamin: Neato, and with type inference we can then do NewBlobAccessMessageReader[remoteexecution.Action] without having to explicitly add our extra Ptr type.

pkg/cas/message_reader.go
meroton-benjamin: For me at least the code reads cleaner with it as `T proto.Message` so I left it as that. If you feel like we definitly want to remove the uncessary dependency holler and I'll change it to Any

pkg/blobstore/replication/nested_blob_replicator.go
meroton-benjamin: I moved the interface to `pkg/cas` and also consumed it from `completeness_checking_blob_access`, and in the the bb-remote-execution side I consume it from `blob_access_directory_fetcher.go` .

NaiveBuildDirectory reads file with Buffer.IntoWriter, do we want to add a BlobWriter interface to pkg/cas?

```
type BlobWriter interface {
    WriteBlob(ctx context.Context, d digest.Digest, w io.Writer) error
}
```

Another case where we have remaining use of the blobstore.BlobAccess api is as source of `FindMissing`. Do we want to define a FindMissingProvider as well and let blobstore.BlobAccess fulfill that?

```
package findmissing

...

type Provider interface {
    FindMissing(ctx context.Context, digests digest.Set) (digest.Set, error)
}
```

I feel like they are logical continuations of the current abstractions and follows the interface segregation principle but I don't know if the abstractions carry their own weight.

meroton-benjamin:
Closed in favor of #365 as github does not allow modifications to source/destination branch.

PR #364, opened 2026-09-03
feat: respect proxy env vars when `client.Configuration.proxy_from_environment` is set

rrbutani:
This matches the [behavior of `http.DefaultTransport`][def-trans] and results in environment variables `$http_proxy`, `$https_proxy`, and `$no_proxy` (and their uppercase counterparts) being consulted to determine which proxy to use/whether to use a proxy for a given URL.

Docs for `ProxyFromEnvironment` are [here][proxy-from-env-doc], source code is [here][src1], [here][src2], and [here][src3].

[def-trans]: https://github.com/golang/go/blob/61b3e65bd2216be76f6208864f84ae184b29ae37/src/net/http/transport.go#L49
[proxy-from-env-doc]: https://pkg.go.dev/net/http#ProxyFromEnvironment

[src1]: https://github.com/golang/go/blob/ebe55c0c38935b39778c38ec8edb1b03c4e96058/src/net/http/transport.go#L501-L520
[src2]: https://github.com/golang/go/blob/ebe55c0c38935b39778c38ec8edb1b03c4e96058/src/net/http/transport.go#L1003-L1015
[src3]: https://github.com/golang/net/blob/eda109f56e86afc41ee941c810e5e4e49f901ec3/http/httpproxy/proxy.go

## motivation

I'm deploying `bb-remote-asset` in an environment where external network access is restricted and must go through proxies, like so:
```bash
export http_proxy=http://proxy.acme.com:100  # route external HTTP traffic via this proxy
export https_proxy=http://proxy.acme.com:200 # likewise, but for HTTPS traffic
export no_proxy=internal.acme.com            # however some domains are exempt and cannot
                                             # be routed through the proxies
```

As far as I can tell it isn't possible to express the above (different proxy URLs for http and https; domain patterns that are exempt from proxying) via `http.client.Configuration.proxy_url` [^workarounds].

[^workarounds]:
    If I understand correctly it is possible to workaround this by having `bb-remote-asset` download assets via RE actions (`FetcherConfiguration.remote_execution`) or by pointing `bb-remote-asset` at a "wrapper" proxy server in a sidecar that consults `$http_proxy`/`$no_proxy`/etc — however, if possible, I'd prefer to have this be handled directly in `bb-remote-asset`

## open questions

~~As written, this PR changes `http.client.NewRoundTripperFromConfiguration` to consult the proxy envivonment variables by default (i.e. when `http.client.Configuration.proxy_url` is not set).~~

~~If this is undesirable (or seems like a potentially confusing/breaking change) I can update this PR to make this behavior opt-in (and mutually exclusive w/`proxy_url`).~~

EDIT: see below

### DIFF
--- a/pkg/http/client/configuration.go
+++ b/pkg/http/client/configuration.go
@@ -25,12 +25,15 @@ func NewRoundTripperFromConfiguration(configuration *pb.Configuration) (http.Rou
 		ForceAttemptHTTP2: !configuration.GetDisableHttp2(),
 		TLSClientConfig:   tlsConfig,
 	}
-	if proxyURL := configuration.GetProxyUrl(); proxyURL != "" {
-		parsedProxyURL, err := url.Parse(proxyURL)
+	switch proxy := configuration.GetProxy().(type) {
+	case *pb.Configuration_ProxyUrl:
+		parsedProxyURL, err := url.Parse(proxy.ProxyUrl)
 		if err != nil {
 			return nil, util.StatusWrap(err, "Failed to parse proxy URL")
 		}
 		defaultTransport.Proxy = http.ProxyURL(parsedProxyURL)
+	case *pb.Configuration_ProxyFromEnvironment:
+		defaultTransport.Proxy = http.ProxyFromEnvironment
 	}
 	var roundTripper http.RoundTripper = &defaultTransport
 
--- a/pkg/proto/configuration/http/client/client.proto
+++ b/pkg/proto/configuration/http/client/client.proto
@@ -13,8 +13,17 @@ message Configuration {
   // certificate will be used when left unset.
   buildbarn.configuration.tls.ClientConfiguration tls = 1;
 
-  // If set, forward all traffic through a proxy with a given URL.
-  string proxy_url = 2;
+  oneof proxy {
+    // If set, forward all traffic through a proxy with a given URL.
+    string proxy_url = 2;
+
+    // If set, use the proxy for a given request, as indicated by the
+    // environment variables http_proxy, https_proxy and no_proxy (or
+    // the uppercase versions thereof). Requests use the proxy from the
+    // environment variable matching their scheme, unless excluded by
+    // no_proxy.
+    google.protobuf.Empty proxy_from_environment = 3;
+  }
 
   message HeaderValues {
     string header = 1;
--- a/pkg/proto/configuration/http/client/client.pb.go
+++ b/pkg/proto/configuration/http/client/client.pb.go
@@ -24,9 +24,13 @@ const (
 )
 
 type Configuration struct {
-	state         protoimpl.MessageState        `protogen:"open.v1"`
-	Tls           *tls.ClientConfiguration      `protobuf:"bytes,1,opt,name=tls,proto3" json:"tls,omitempty"`
-	ProxyUrl      string                        `protobuf:"bytes,2,opt,name=proxy_url,json=proxyUrl,proto3" json:"proxy_url,omitempty"`
+	state protoimpl.MessageState   `protogen:"open.v1"`
+	Tls   *tls.ClientConfiguration `protobuf:"bytes,1,opt,name=tls,proto3" json:"tls,omitempty"`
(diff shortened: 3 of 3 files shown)


EdSchouten:
Hey there,

> ## open questions
> As written, this PR changes `http.client.NewRoundTripperFromConfiguration` to consult the proxy envivonment variables by default (i.e. when `http.client.Configuration.proxy_url` is not set).
> 
> If this is undesirable (or seems like a potentially confusing/breaking change) I can update this PR to make this behavior opt-in (and mutually exclusive w/`proxy_url`).

I'm fine with adding this feature, but I do think we should make it opt-in. Can you please change the configuration schema as follows? Thanks!

```diff
diff --git pkg/proto/configuration/http/client/client.proto pkg/proto/configuration/http/client/client.proto
index 3c2741d..295d2be 100644
--- pkg/proto/configuration/http/client/client.proto
+++ pkg/proto/configuration/http/client/client.proto
@@ -13,8 +13,17 @@ message Configuration {
   // certificate will be used when left unset.
   buildbarn.configuration.tls.ClientConfiguration tls = 1;

-  // If set, forward all traffic through a proxy with a given URL.
-  string proxy_url = 2;
+  oneof proxy {
+    // If set, forward all traffic through a proxy with a given URL.
+    string proxy_url = 2;
+
+    // If set, use the proxy for a given request, as indicated by the
+    // environment variables HTTP_PROXY, HTTPS_PROXY and NO_PROXY (or
+    // the lowercase versions thereof). Requests use the proxy from the
+    // environment variable matching their scheme, unless excluded by
+    // NO_PROXY.
+    google.protobuf.Empty proxy_from_environment = 3;
+  }

   message HeaderValues {
     string header = 1;
```

pkg/http/client/configuration.go
```diff
@@ -25,12 +25,15 @@ func NewRoundTripperFromConfiguration(configuration *pb.Configuration) (http.Rou
 		ForceAttemptHTTP2: !configuration.GetDisableHttp2(),
 		TLSClientConfig:   tlsConfig,
 	}
-	if proxyURL := configuration.GetProxyUrl(); proxyURL != "" {
-		parsedProxyURL, err := url.Parse(proxyURL)
+	switch proxy := configuration.GetProxy().(type) {
+	case *pb.Configuration_ProxyUrl:
+		parsedProxyURL, err := url.Parse(proxy.ProxyUrl)
```
rrbutani, line 30: as written, this change (i.e. dropping the `proxy.ProxyUrl != ""` check) will cause configurations that _explicitly_ set `proxy_url` to an empty string to now result in an error

is this change acceptable?

pkg/proto/configuration/http/client/client.proto
```diff
@@ -13,8 +13,17 @@ message Configuration {
...
+    // If set, forward all traffic through a proxy with a given URL.
+    string proxy_url = 2;
+
+    // If set, use the proxy for a given request, as indicated by the
+    // environment variables http_proxy, https_proxy and no_proxy (or
+    // the uppercase versions thereof). Requests use the proxy from the
+    // environment variable matching their scheme, unless excluded by
+    // no_proxy.
+    google.protobuf.Empty proxy_from_environment = 3;
+  }
```
rrbutani, line 26: this is taken verbatim from https://github.com/buildbarn/bb-storage/pull/364#issuecomment-5523952994 but with one small tweak: lowercase variable names

(`ProxyFromEnvironment` will prefer lowercase variables names starting in go 1.28: https://github.com/golang/go/issues/79656, https://github.com/golang/net/commit/a02ddfa7eacb4cf63a5bea6b23761244a6df69f6, https://tip.golang.org/doc/next)

is this acceptable?

rrbutani (commented):
@EdSchouten thanks for the speedy response! 🙂

I've made the change; left a couple of notes below. Please let me know if you'd prefer the bits in `configuration.go` be written in a different way (I'm not very familiar with go).

EdSchouten (approved):
Looking perfect!

PR #365, opened 2026-09-05
Abstract reading messages from the CAS

meroton-benjamin:
This commit abstracts reading messages from the Content Addressable Storage (CAS) behind two primary interfaces cas.MessageReader[T] and cas.BlobStreamReader.

Many Buildbarn components depend on blobstore.BlobAccess only as a source of proto.Message and these interface allows us to sever that dependency in components which are otherwise not interested in the blobstore.BlobAccess interface.


meroton-benjamin:
This pull request is the next iteration of #363 but as github doesn't allow us to change the source branch of the pull a new pull request was created. Major changes:

* BlobStreamReader interface moved to `pkg/cas`
* BlobStreamReader is now consumed by `pkg/blobstore/completenesschecking` thereby limiting blob access api exposure further.

pkg/cas/nested_blob_replicator_test.go
```diff
@@ -18,142 +18,160 @@ func TestNestedBlobReplicator(t *testing.T) {
 	ctrl, ctx := gomock.WithContext(context.Background(), t)
 
 	replicator := mock.NewMockBlobReplicator(ctrl)
-	nestedReplicator := replication.NewNestedBlobReplicator(replicator, digest.KeyWithoutInstance, 10000)
+	actionReader := mock.NewMockMessageReader[*remoteexecution.Action](ctrl)
```
EdSchouten, line 21: Oh nice. Pretty awesome to see that go mock supports generics properly nowadays.

pkg/blobstore/blob_access_blob_stream_reader.go
```diff
@@ -0,0 +1,25 @@
+package blobstore
```
EdSchouten: Q: Any reason we can't place these implementations in pkg/cas, next to where the interfaces are defined? Would that introduce cyclic dependencies? I think it doesn't...

pkg/cas/blob_stream_reader.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"context"
+	"io"
+
+	"github.com/buildbarn/bb-storage/pkg/digest"
+)
+
+// BlobStreamReader reads a blob from the Content Addressable Storage
+// (CAS) in a streaming fashion.
+type BlobStreamReader interface {
```
EdSchouten: What are your thoughts on calling this StreamReader?

internal/mock/BUILD.bazel
```diff
@@ -320,6 +320,29 @@ gomock(
...
+    name = "cas",
+    out = "cas.go",
+    interfaces = ["BlobStreamReader"],
+    library = "//pkg/cas",
+    mockgen_model_library = "@org_uber_go_mock//mockgen/model",
+    mockgen_tool = "@org_uber_go_mock//mockgen",
+    package = "mock",
+)
+
+# Normally we would put the mocks of this interface in cas.go but as it
```
EdSchouten: Glad to see you got this working. I tried doing source based generation at some point in the past, but couldn't get it to work properly.

Instead of having multiple `gomock()`s, what are your thoughts on using source based generation for both interface types? As in, collapse these two together.

pkg/blobstore/replication/cas_replicator.go
```diff
@@ -0,0 +1,13 @@
+package replication
+
+import (
+	"context"
+
+	"github.com/buildbarn/bb-storage/pkg/digest"
+)
+
+// CASReplicator replicates blobs to the destination Content
```
EdSchouten: Hmmm... I'm just thinking about (the implications of adding) this interface.

If I understand it correctly, your reason is sort as follows: "In the future, BlobReplicator is only going to receive digests of chunks; not full files. But in case of NestedBlobReplicator we need to replicate full files. So we effectively need the same kind of interface as BlobReplicator, but then for full files." Is that correct?

EdSchouten (commented):
This is starting to look great! We should be able to merge this soon.

pkg/blobstore/blob_access_blob_stream_reader.go
meroton-benjamin: That seems to work without adding any cyclical dependencies

pkg/cas/blob_stream_reader.go
meroton-benjamin: renamed to StreamReader

internal/mock/BUILD.bazel
meroton-benjamin: The gomock rule only takes a single source file for source based generation I have not managed to set it up in such a manner that I could do this for both . To collapse them together we'd need to put both interfaces in the same source file (like an `interfaces.go`) adding the extra mock call seemed like the better choice..

pkg/blobstore/replication/cas_replicator.go
meroton-benjamin: Yes that is the reasoning I had for adding this interface. When we add cdc then BlobReplicators will still only going to replicate single chunks/messages for other BlobAccess:es but NestedBlobReplicator needs to replicate full blobs from one cas to another which means it needs a chunking aware replicator (copying both the constituent chunks and the chunk lists).

EdSchouten: Gotcha! Yeah, that makes sense. Though in that case I would recommend doing the following:

1. Move nested_blob_replicator.go and this new interface into pkg/cas, as it's completely CAS specific, and should eventually never directly operate on the digests used by pkg/blobstore.
2. In this interface, rename ReplicateMultiple() to something different. Maybe just Replicate()? It doesn't matter what, as long as it's different from ReplicateMultiple(). That prevents it from being satisfied by BlobReplicator instances.
3. Add a very simple implementation that is backed by a BlobAccess. We will need to rework that one once we actually add chunking.

internal/mock/BUILD.bazel
```diff
@@ -152,6 +152,29 @@ gomock(
...
+# is generic it needs to be supplied as a source file which we therefore
+# put in cas_message_reader.go.
+gomock(
+    name = "cas_message_reader",
+    out = "cas_message_reader.go",
+    interfaces = ["MessageReader"],
+    library = "//pkg/cas",
+    mockgen_tool = "@org_uber_go_mock//mockgen",
+    package = "mock",
+    source = "//pkg/cas:message_reader.go",
```
EdSchouten, line 178: Ah, gotcha. I overlooked this specific `source` line. Yeah, we're going to need two separate `gomock()` calls. Fairly dumb that gomock/rules_go can't just take a list of source files and extract the interface(s) from the correct file.

pkg/blobstore/completenesschecking/completeness_checking_blob_access_test.go
```diff
@@ -149,10 +153,13 @@ func TestCompletenessCheckingBlobAccess(t *testing.T) {
 				buffer.BackendProvided(dataIntegrityCallback.Call),
 			),
 		)
-		contentAddressableStorage.EXPECT().Get(
+
+		errReader := io.NopCloser(iotest.ErrReader(status.Error(codes.Internal, "Hard disk has a case of the Mondays")))
```
EdSchouten:
```go
errReader := NewMockReadCloser(ctrl)
errReader.EXPECT().Read(gomock.Any()).Return(0, status.Error(codes.Internal, "Hard disk has a case of the Mondays"))
errReader.EXPECT().Close()
```

Seems a bit more idiomatic within our code base, and has the advantage that we also test that stuff doesn't leak.

meroton-benjamin: Switched to using MockReadCloser

pkg/blobstore/replication/cas_replicator.go
meroton-benjamin: nested_blob_replicator.go and cas_replicator.go now reside in pkg/cas.

`replication.CASReplicator` with method `ReplicateMultiple(context.Context, digest.Set) error` is now `cas.Replicator` with method `Replicate(context.Context, digest.Set) error`.

The initial trivial implementation just forwards to a `replication.BlobReplicator` that exists internally. I imagine that with chunking we'll have the code paths diverge completely and bb_copy will simply have a suitable strategy which makes sense for replicating blobs in a cdc world.

EdSchouten (approved):

EdSchouten:
Looks good! Will merge once CI is happy.

PR #366, opened 2026-09-06
 Replace CAS with CS and CLS

meroton-benjamin:
This commit builds on top of our split and splice blob support to make
it a mandatory first class feature in Buildbarn. With this commit the
Content Addressable Storage (CAS) is created from two Storage
configurations that work in tandem. A Chunk Storage (CS) which is
content addressed and contains chunks of blobs, and a Chunk List Storage
(CLS) which is addressed by a blob digest and contains a manifest
describing the chunks that make up the blob.

All api calls are automatically translated to use Chunk Lists created
with RepMaxCDC. Effectively this means that large blobs no longer exists
in the storage layer, individual chunks of the large blobs are in turn
deduplicated in such a manner that the chunks are stored only once.

The automatic translation makes certain that clients that are not cdc
aware can still continue to use the storage backend without performing
any changes. Clients which support RepMaxCDC also gets a significant
reduction in the amount of blobs to transfer as they only need to
transfer modified chunks rather than the entire blob.


pkg/blobstore/cdc/bypass.go
```diff
@@ -0,0 +1,22 @@
...
+
+import "context"
+
+type chunkListValidationBypassKey struct{}
+
+// NewContextWithChunkListValidationBypass creates a derived context
+// that signals downstream storage layers that the chunk list has
+// already been validated (or was freshly generated) and does not need
+// expensive re-validation.
+func NewContextWithChunkListValidationBypass(ctx context.Context) context.Context {
```
EdSchouten, line 11: I think that we should get rid of this, and turn it into a function argument. For example, change BlobAccess.Put() to take a `validated bool` parameter.

pkg/ttlcache/ttl_cache.go
```diff
@@ -0,0 +1,97 @@
...
+		evictionSet:   evictionSet,
+		maxItems:      maxItems,
+		cacheDuration: cacheDuration,
+		items:         make(map[K]cachedItem[V]),
+	}
+}
+
+// NewTTLCacheFromConfiguration wraps NewTTLCache with the parameters
+// specified in a configuration message.
+func NewTTLCacheFromConfiguration[K comparable, V any](configuration *digest_pb.ExistenceCacheConfiguration, clock clock.Clock, name string) (*TTLCache[K, V], error) {
```
EdSchouten, line 43: On the Bonanza side I actually have a somewhat comparable abstraction:

https://pkg.go.dev/bonanza.build/pkg/ds/lossymap

It started out as a generalization to the key-location map of Bonanza's equivalent of LocalBlobAccess. But it can also be used for other types of 'lossy maps'.

What are your thoughts on moving the lossymap package into bb-storage?

pkg/proto/configuration/blobstore/blobstore.proto
```diff
@@ -705,7 +765,7 @@ message ReferenceExpandingBlobAccessConfiguration {
 
   // Optional: Storage backend to use when Reference objects refer to
   // objects stored in another Content Addressable Storage.
-  BlobAccessConfiguration content_addressable_storage = 5;
+  ContentAddressableStorageConfiguration content_addressable_storage = 5;
```
EdSchouten, line 768: Yeah, this is interesting. We'll need to figure out how to make ICAS work in the post chunking world. But are we sure that this is conceptually correct? Could a Chunk Store of type ReferenceExpanding be backed by another **CAS**? I can understand how it can be backed by another Chunk Store.

If we don't know how to solve this specific aspect, I'd be happy to phase out this specific feature. As in, keep ICAS support there, but remove the ability to reference another CAS.

pkg/proto/blobstore/chunklist/chunk_list.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
```
EdSchouten, line 1: chunklist.proto to match the directory name.

pkg/digest/function.go
```diff
@@ -113,6 +113,9 @@ func (f Function) NewDigestFromProto(digest *remoteexecution.Digest) (Digest, er
 	if digest == nil {
 		return BadDigest, status.Error(codes.InvalidArgument, "No digest provided")
 	}
+	if digest.SizeBytes < 0 {
+		return BadDigest, status.Error(codes.InvalidArgument, "Negative size digest provided")
+	}
```
EdSchouten, line 118: Why is this specific change necessary? `Function.NewDigest()` also has a similar check, right?

pkg/cas/stream_reader.go
```diff
@@ -13,3 +13,19 @@ type StreamReader interface {
 	// Read a blob from the CAS with the specific digest.
 	ReadStream(ctx context.Context, d digest.Digest) (io.ReadCloser, error)
 }
+
+type streamReader struct {
```
EdSchouten, line 17: storageBackedStreamReader?

pkg/blobstore/grpcservers/content_addressable_storage_server.go
```diff
@@ -150,10 +156,86 @@ func (contentAddressableStorageServer) GetTree(in *remoteexecution.GetTreeReques
...
+			return nil, util.StatusWrap(err, "Failed to check blob existence")
+		}
+		if !missing.Empty() {
+			return nil, status.Errorf(codes.NotFound, "Blob %s not found", blobDigest)
+		}
+		return &remoteexecution.SplitBlobResponse{
+			ChunkDigests:     []*remoteexecution.Digest{in.BlobDigest},
+			ChunkingFunction: remoteexecution.ChunkingFunction_REP_MAX_CDC,
+		}, nil
+	}
```
EdSchouten, line 225: Where is this specific aspect covered? I'm looking at remote_execution.proto, and I don't think it describes the situation where the file is already a chunk?

pkg/blobstore/grpcservers/content_addressable_storage_server.go
```diff
@@ -150,10 +156,86 @@ func (contentAddressableStorageServer) GetTree(in *remoteexecution.GetTreeReques
...
+
+	return &remoteexecution.SpliceBlobResponse{
+		BlobDigest: in.BlobDigest,
+	}, nil
 }
 
-func (contentAddressableStorageServer) SplitBlob(ctx context.Context, in *remoteexecution.SplitBlobRequest) (*remoteexecution.SplitBlobResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "This service does not support splitting blobs")
+func (s *contentAddressableStorageServer) SplitBlob(ctx context.Context, in *remoteexecution.SplitBlobRequest) (*remoteexecution.SplitBlobResponse, error) {
```
EdSchouten, line 193: Can we please implement the latest version of the spec? So call this GetChunkMapping() instead of SplitBlob().

(Maybe we should also rename the Chunk List Store to the Chunk Mapping Store? That better maps to REv2.)

pkg/blobstore/grpcclients/cls_blob_access.go
```diff
@@ -0,0 +1,122 @@
...
+		return nil, err
+	}
+	cacheCapabilities := serverCapabilities.CacheCapabilities
+	// Only return fields that pertain to Chunk List Storage.
+	return &remoteexecution.ServerCapabilities{
+		CacheCapabilities: &remoteexecution.CacheCapabilities{
+			SplitBlobSupport:  cacheCapabilities.SplitBlobSupport,
+			SpliceBlobSupport: cacheCapabilities.SpliceBlobSupport,
+		},
+	}, nil
```
EdSchouten, line 121: I thought we agreed that GetCapabilities() should always be invoked against the Chunk Store, not the Chunk List Store. In other words, this needs a `panic()`, just like we have for ICAS, ISCC, etc.

pkg/cas/operations.go
```diff
@@ -0,0 +1,212 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+// IsSingleChunk checks if a digest is represented by a single chunk.
+func IsSingleChunk(params cdc.Parameters, d digest.Digest) bool {
+	return d.GetSizeBytes() < 2*params.MinChunkSizeBytes
+}
+
+// GetBytes returns the bytes of a digest from the CAS as a byteslice.
+func GetBytes(ctx context.Context, cas ContentAddressableStorage, d digest.Digest) ([]byte, error) {
```
EdSchouten, line 24: A function like this should respect an upper size limit.

pkg/cas/content_addressable_storage.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+
+	"github.com/buildbarn/bb-storage/pkg/blobstore/cdc"
+	"github.com/buildbarn/bb-storage/pkg/blobstore/chunklist"
+	"github.com/buildbarn/bb-storage/pkg/digest"
+)
+
+// ContentAddressableStorage is an interface which describes
+// interactions with a Content Addressable Storage (CAS).
+type ContentAddressableStorage interface {
```
EdSchouten, line 13: I think we need to take a step back and ask ourselves what this interface actually abstracts. Right now it's only there to stitch a couple of things together: Chunk Store, Chunk List Store, and fetching of capabilities, and I'm not sure that's worth a lot.

Why can't we have implementations of MessageReader and StreamReader that are directly backed by:

- An instance of cdc.ParametersFetcher
- A Chunk Store BlobAccess
- A Chunk List Store BlobAccess

?

For uploading files/protos into the CAS, that's where we need to write a new abstraction. But that can potentially live in bb-remote-execution?

pkg/blobstore/grpcclients/cs_blob_access.go
```diff
@@ -395,7 +395,10 @@ func (ba *casBlobAccess) GetCapabilities(ctx context.Context, instanceName diges
 	// issue batch operations.
 	return &remoteexecution.ServerCapabilities{
 		CacheCapabilities: &remoteexecution.CacheCapabilities{
-			DigestFunctions: digest.RemoveUnsupportedDigestFunctions(cacheCapabilities.DigestFunctions),
+			DigestFunctions:   digest.RemoveUnsupportedDigestFunctions(cacheCapabilities.DigestFunctions),
+			SplitBlobSupport:  cacheCapabilities.SplitBlobSupport,
+			SpliceBlobSupport: cacheCapabilities.SpliceBlobSupport,
```
EdSchouten, line 400: So the entire idea behind these GetCapabilities() implementations is that they return/yield the kinds of properties that are **variable**. In this implementation we're going to **require** that implementations support chunking.

So with that in mind, I think that this function should only check that the CacheCapabilities returned by the server have SplitBlobSupport and SpliceBlobSupport set to true. It doesn't need to return those fields here. Then in cmd/bb_storage/main.go, we should set these fields to true (at the same place where we set SupportedCompressors).

Similarly, this code should never return a ServerCapabilities that doesn't have RepMaxCdcParams set.

pkg/blobstore/configuration/cs_blob_access_creator.go
```diff
@@ -28,55 +30,54 @@ import (
 	"cloud.google.com/go/storage"
 )
 
-var casCapabilitiesProvider = capabilities.NewStaticProvider(&remoteexecution.ServerCapabilities{
+var csCapabilitiesProvider = capabilities.NewStaticProvider(&remoteexecution.ServerCapabilities{
 	CacheCapabilities: &remoteexecution.CacheCapabilities{
 		DigestFunctions: digest.SupportedDigestFunctions,
```
EdSchouten, line 35: This should have SupportsSplitBlobs and such set.

pkg/blobstore/blob_access.go
```diff
@@ -16,7 +15,6 @@ type BlobAccess interface {
 	capabilities.Provider
 
 	Get(ctx context.Context, digest digest.Digest) buffer.Buffer
```
EdSchouten, line 17: This may now return `(buffer.Buffer, error)`

pkg/blobstore/blob_access.go
```diff
@@ -16,7 +15,6 @@ type BlobAccess interface {
...
-	GetFromComposite(ctx context.Context, parentDigest, childDigest digest.Digest, slicer slicing.BlobSlicer) buffer.Buffer
 	Put(ctx context.Context, digest digest.Digest, b buffer.Buffer) error
```
EdSchouten, line 18: Consider making Buffer a generic type now.

- CAS: `buffer.Buffer[[]byte]`
- AC: `buffer.Buffer[*remoteexecution.ActionResult]`
- CLS: `buffer.Buffer[cdc.ChunkList]`

?

```go
type BufferMarshaler[T any] {
      Marshal(digestFunction digest.Function, T any) ([]byte, error)
      Unmarshal(digestFunction digest.Function, []byte) (T, error)
}
```

pkg/blobstore/cdc/bypass.go
meroton-benjamin: validated is now a property of chunklist.ChunkList

pkg/proto/configuration/blobstore/blobstore.proto
meroton-benjamin: The other `CAS` may have other chunking parameters making it so that if we refer to chunks from a different chunk storage then we have to consider it to be a CAS in the context.

pkg/proto/blobstore/chunklist/chunklist.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
```
meroton-benjamin: ack

pkg/digest/function.go
meroton-benjamin: it was not

pkg/blobstore/grpcservers/content_addressable_storage_server.go
meroton-benjamin: The api requires us to be able to answer GetChunkMapping for all api calls even though we do not store a list for single chunk blobs. As we can synthesize the list I figured here would be a good place but do you want to see it in e.g. ChunkListValidatingBlobAccess instead?

pkg/blobstore/grpcservers/content_addressable_storage_server.go
meroton-benjamin: Implemented with latest version of the spec, I have yet to rename Chunk List Store to Chunk Mapping Store.

pkg/blobstore/grpcclients/cls_blob_access.go
meroton-benjamin: fixed

pkg/cas/content_addressable_storage.go
meroton-benjamin: I have removed the ContentAddressableStorage interface in favor of taking whatever is needed in the operations themselves

pkg/blobstore/grpcclients/cs_blob_access.go
meroton-benjamin: ack

pkg/blobstore/configuration/cs_blob_access_creator.go
meroton-benjamin: ack

pkg/blobstore/blob_access.go
meroton-benjamin: Returns (T, error)

pkg/blobstore/blob_access.go
meroton-benjamin: blobstore.BlobAccess is now generic

pkg/cas/operations.go
meroton-benjamin: fixed

meroton-benjamin:
State of latest commit:

Review comments addressed and rebased on top of the lossymap implementation. This version does not use the [lossymap](https://github.com/buildbarn/bb-storage/pull/368) but it could be an alternative to using the generic ttlcache.TTLCache. One question that remains before we do that is if we case about the eviction set eviction policy. It may be desireable to pick a different implementation for different caches, say an LRU implementation for the chunk cache, a FIFO implementation for the chunk list cache etc etc.

pkg/blobstore/buffer/chunk.go
```diff
@@ -0,0 +1,147 @@
+package buffer
```
EdSchouten, line 1: Do we think that it's fitting to put this in pkg/blobstore/buffer? Maybe the correct thing to do is to rename pkg/blobstore/chunklist to pkg/blobstore/chunk, and then move this in there as well?

pkg/blobstore/buffer/chunk.go
```diff
@@ -0,0 +1,147 @@
...
+
+// Chunk is an abstraction over a single chunk in the Chunk Storage (CS)
+// that may or may not be compressed. Chunk objects are thread safe but
+// not trivially copyable refer to them by pointer.
+type Chunk struct {
+	zstdPool zstd.Pool
+
+	mu sync.Mutex
+
+	hasData atomic.Bool
```
EdSchouten, line 23: Out of curiosity, when would a buffer not have any data in decompressed form? Given that we now always perform checksum validation, we always end up decompressing objects anyway. This is probably not necessary?

pkg/blobstore/buffer/chunk.go
```diff
@@ -0,0 +1,147 @@
...
+	"google.golang.org/grpc/codes"
+)
+
+// Chunk is an abstraction over a single chunk in the Chunk Storage (CS)
+// that may or may not be compressed. Chunk objects are thread safe but
+// not trivially copyable refer to them by pointer.
+type Chunk struct {
+	zstdPool zstd.Pool
+
+	mu sync.Mutex
```
EdSchouten, line 21: Why do we need this mutex?

pkg/blobstore/buffer/chunk.go
```diff
@@ -0,0 +1,147 @@
...
+
+	hasData atomic.Bool
+	data    []byte
+
+	hasCompressedData atomic.Bool
+	compressedData    []byte
```
EdSchouten, line 27: Maybe it's a bit more elegant to use a `atomic.Pointer[[]byte]` here?

pkg/blobstore/buffer/read_at_closer.go
```diff
@@ -0,0 +1,9 @@
+package buffer
```
EdSchouten, line 1: Given that nothing in pkg/blobstore/buffer uses this interface, could we remove this? Or maybe move it elsewhere?

pkg/blobstore/cdc/parameters.go
```diff
@@ -0,0 +1,48 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// ParametersFetcher retrieves the Content Defined Chunking (CDC)
+// parameters that govern how blobs are decomposed into chunks. Unless
+// an error is returned, the returned value is guaranteed to be non-nil
+// and to have been validated, so that callers may rely on its values
+// without performing any checks.
+type ParametersFetcher interface {
```
EdSchouten, line 20: Just to check: this was going to be moved into pkg/capabilities, right? CDCParametersFetcher or something.

pkg/blobstore/chunklist/chunk_list.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// Fetcher retrieves a ChunkList for a digest.
+type Fetcher interface {
+	FetchChunkList(ctx context.Context, digest digest.Digest) (ChunkList, error)
+}
+
+// FindChunkOffset returns the index of the chunk containing the given
+// offset and the offset within that chunk.
+func FindChunkOffset(chunkList ChunkList, off uint64) (index int, chunkOffset int64) {
```
EdSchouten, line 24: Maybe this should be a method of ChunkList?

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,304 @@
...
+			return chunklist.ChunkList{}, status.Error(codes.NotFound, "Blob not found in CAS")
+		}
+
+		chunkList := chunklist.ChunkList{
+			Offsets:   []uint64{0},
+			Digests:   []digest.Digest{d},
+			Validated: true,
+		}
+		return chunkList, nil
+	}
```
EdSchouten, line 113: Under what circumstances is this valid/do we ever run into this? As in, I can imagine that high-level code that gets a digest needs to distinguish between the large file/single chunk case, but does this need to live in BlobAccess?

Maybe it's better to reject such requests outright?

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,304 @@
...
+
+	params, err := ba.cdcParametersFetcher.FetchCDCParameters(ctx, d.GetInstanceName())
+	if err != nil {
+		return err
+	}
+
+	// Check that all referenced chunks are present in storage.
+	value, err = ba.flattenChunks(ctx, params, value)
+	if err != nil {
+		return status.Error(codes.NotFound, "At least one chunk is missing from storage.")
```
EdSchouten, line 159: General remark: please don't end error messages with a period/full stop.

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	return chunklist.ChunkList{
+		Offsets:   flattenedOffsets,
+		Digests:   flattenedDigests,
+		Validated: userChunkList.Validated,
+	}, nil
+}
+
+func (ba *chunkListValidatingBlobAccess) findMissingChunks(ctx context.Context, d digest.Digest) (digest.Set, error) {
+	storedChunkList, err := ba.BlobAccess.Get(ctx, d)
```
EdSchouten, line 275: Shouldn't this use the ChunkListFetcher?

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,304 @@
...
+	if err != nil {
+		return digest.EmptySet, util.StatusWrap(err, "Failed to decode chunk list")
+	}
+	builder := digest.NewSetBuilder(len(storedChunkList.Digests))
+	for _, digest := range storedChunkList.Digests {
+		builder.Add(digest)
+	}
+	return ba.chunkStorage.FindMissing(ctx, builder.Build())
```
EdSchouten, line 283: This approach causes us to call FindMissing O(n) times. My suspicion is that most files will only have a very small number of chunks. Wouldn't it be smarter to aggregate those and call FindMissing() only when a certain threshold has been reached?

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+func (ba *chunkListValidatingBlobAccess) Put(ctx context.Context, d digest.Digest, value chunklist.ChunkList) error {
+	if value.Validated {
+		// ChunkList has already been validated, push it directly to
+		// downstream blob store.
+		return ba.BlobAccess.Put(ctx, d, value)
+	}
+
+	if len(value.Digests) == 0 {
```
EdSchouten, line 140: Q: Under which circumstances could this occur? Wouldn't it be better/smarter if we only ever instantiated ChunkLists with two or more elements?

Something similar can be enforced for the combined size of the chunk list: never instantiate ChunkList if the combined size of the digests doesn't correspond to the size of the full object.

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	// Verify the whole blob against the advertised digest.
+	if actual := wholeGen.Sum(); actual != d {
+		return status.Errorf(codes.InvalidArgument, "Blob digest mismatch: advertised %s, actual %s", d, actual)
+	}
+
+	// Store the canonical chunk list.
+	if err := ba.BlobAccess.Put(ctx, d, canonicalChunkList); err != nil {
+		return util.StatusWrap(err, "Failed to save canonical chunk list")
```
EdSchouten, line 218: Do we actually need to perform canonicalization like this? As far as I know, the only thing that this BlobAccess needs to validate at this specific point is that `digest.Hash` is equal to that of the objects when concatenated.

pkg/blobstore/coder/chunk_coder.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	return nil
+}
+
+func (c *chunkCoder) Encode(chunk *buffer.Chunk, d digest.Digest) ([]byte, error) {
+	// TODO: Should ctx be part of this signature?
+	ctx := context.Background()
+	if err := c.checkDigest(ctx, chunk, d); err != nil {
+		return nil, err
+	}
```
EdSchouten, line 47: I don't think this is necessarily needed here. Can't we simply have an agreement that any instance of `buffer.Chunk` has contents matching the associated digest?

pkg/blobstore/coder/chunk_coder.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
+	"github.com/buildbarn/bb-storage/pkg/digest"
+	"github.com/buildbarn/bb-storage/pkg/zstd"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type chunkCoder struct {
+	zstdPool         zstd.Pool
+	encodeCompressed bool
```
EdSchouten, line 15: This flag isn't being checked anywhere.

pkg/blobstore/coder/chunk_list_coder.go
```diff
@@ -0,0 +1,122 @@
...
+// determined to compose the digest which it is being addressed by (see
+// chunk_list_validating_blob_access.go for the logic to validate a
+// chunk list).
+//
+// Binary format:
+//
+//	struct ChunkList {
+//	  uint8_t version;                  // The version, currently 0x00.
+//	  uint8_t digest_function;          // The REv2 digest function enum.
+//	  uint32_t count:                   // The number of chunks.
```
EdSchouten, line 39: To me it's not really clear why we need these fields.

- Version: If we ever need to change the encoding of chunk lists, then we likely want LocalBlobAccess to suppress/ignore any existing entries. The way this is implemented right now, we'd just get a hard error. The system would not be able to recover gracefully.
- The digest function is implied from the digest under which the chunk list is stored.
- The number of elements can be trivially inferred from the size of the payload.

I'd say, just store 36 bytes per entry without any special container format.

pkg/blobstore/coder/xxh64_suffix_coder.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/buildbarn/bb-storage/pkg/digest"
+	"github.com/cespare/xxhash/v2"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type xxh64SuffixCoder struct{}
+
+// NewXXH64SuffixCoder creates a Coder that appends an 8-byte XXH64
+// checksum. XXH64 is a modern non cryptographic hash function designed
```
EdSchouten, line 15: Why not SHA-256? It should be pretty fast on modern hardware.

pkg/blobstore/coder/chunk_list_coder.go
```diff
@@ -0,0 +1,122 @@
...
+
+type chunkListCoder struct {
+	prevalidated bool
+}
+
+// NewChunkListCoder returns a Coder that can encode and decode a
+// chunklist.ChunkList into an efficient binary format.
+//
+// The binary schema uses a Struct-of-Arrays (SoA) layout to group sizes
+// together which maximizes compressibility but does not itself compress
```
EdSchouten, line 24: Even with the sizes and hashes split, I don't think that compressing these records is ever advantageous. The hashes are 32 bytes in size and are simply not compressible. The sizes are only 4 bytes. So in the best case we only save 4/(32+4) = 1/9 the space.

pkg/blobstore/coder/zstd_coder.go
```diff
@@ -0,0 +1,83 @@
...
+	// expected.
+	zstdDecodeGrowthFactor = 1
+
+	// zstdMinimumBufferSize prevents very small initial buffers from
+	// being used. A value of 64 corresponds to the value used natively
+	// by bytes.Buffer.
+	zstdMinimumBufferSize = 64
+)
+
+type zstdCoder struct {
```
EdSchouten, line 31: So this Coder is only going to be used by non-chunkstore, right? So this would allow compressing AC entries, etc. Have we ever observed this to be necessary? AC entries are tiny.

pkg/blobstore/completenesschecking/completeness_checking_blob_access.go
```diff
@@ -191,13 +204,6 @@ func (ba *completenessCheckingBlobAccess) checkCompleteness(ctx context.Context,
 			}
 			return nil
 		}); err != nil {
-			// Any errors generated above may be caused by
-			// data corruption on the Tree object. Force
-			// reading the Tree until completion, and prefer
-			// read errors over any errors generated above.
```
EdSchouten, line 197: I believe that in the case of REv2 this still holds. It may well be the case that storage serves us a bad ChunkList, where the chunks when concatenated don't add up to the original digest. Implementations of StreamReader still need to check for this. They should reject handing out the final chunk in those cases.

pkg/blobstore/coder/xxh64_suffix_coder.go
```diff
@@ -0,0 +1,41 @@
...
+// to be extremely fast.
+func NewXXH64SuffixCoder() Coder[[]byte, []byte] {
+	return &xxh64SuffixCoder{}
+}
+
+func (xxh64SuffixCoder) Encode(data []byte, parentDigest digest.Digest) ([]byte, error) {
+	hash := xxhash.Sum64(data)
```
EdSchouten, line 22: I would consider adding `parentDigest` to the hash as well. That way we don't just validate that the object is intact. We also verify that it belongs to the digest under which it is stored.

pkg/blobstore/configuration/cs_blob_access_creator.go
```diff
@@ -20,90 +26,88 @@ import (
...
 	"cloud.google.com/go/storage"
 )
 
-var casCapabilitiesProvider = capabilities.NewStaticProvider(&remoteexecution.ServerCapabilities{
+var csCapabilitiesProvider = capabilities.NewStaticProvider(&remoteexecution.ServerCapabilities{
 	CacheCapabilities: &remoteexecution.CacheCapabilities{
-		DigestFunctions: digest.SupportedDigestFunctions,
+		DigestFunctions:   digest.SupportedDigestFunctions,
+		SplitBlobSupport:  true,
+		SpliceBlobSupport: true,
```
EdSchouten, line 40: Again, I think that setting these fields should be the responsibility of cmd/bb_storage/main.go. Inside of BlobAccess we should only set fields that are actually variable.

pkg/blobstore/buffer/chunk.go
EdSchouten: Or maybe use a `sync.Once`?

pkg/blobstore/grpcclients/cls_blob_access.go
```diff
@@ -0,0 +1,97 @@
...
+	return &clsBlobAccess{
+		contentAddressableStorageClient: remoteexecution.NewContentAddressableStorageClient(client),
+		maximumMessageSizeBytes:         maximumMessageSizeBytes,
+	}
+}
+
+func (ba *clsBlobAccess) Get(ctx context.Context, blobDigest digest.Digest) (chunklist.ChunkList, error) {
+	digestFunction := blobDigest.GetDigestFunction()
+
+	// TODO: Replace with streaming variant
```
EdSchouten, line 32: Even though this is already stated as a TODO, I'm specifically adding a comment here so we don't forget. In my opinion we should only invoke the new API internally.

pkg/blobstore/grpcclients/cls_blob_access.go
```diff
@@ -0,0 +1,97 @@
...
+	offset := uint64(0)
+	for i, proto := range splitBlobsResponse.ChunkDigests {
+		d, err := digestFunction.NewDigestFromProto(proto)
+		if err != nil {
+			return chunklist.ChunkList{}, err
+		}
+		chunkList.Offsets[i] = offset
+		chunkList.Digests[i] = d
+		offset += uint64(d.GetSizeBytes())
+	}
```
EdSchouten, line 57: This should likely be the place where we also validate that the combined size of the chunk list matches the digest. Probably want to use https://pkg.go.dev/math/bits#Add64 here to ensure stuff doesn't overflow.

pkg/blobstore/grpcclients/cs_blob_access.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(resp.GetResponses()) != 1 {
+		return nil, status.Errorf(codes.Internal, "Expected 1 response, got %d", len(resp.Responses))
+	}
+
+	r := resp.Responses[0]
+	if r.GetStatus().GetCode() != int32(codes.OK) {
```
EdSchouten, line 92: Maybe cleaner to write something like this:
```go
if err := status.ErrorProto(r.Status); err != nil {
    return nil, err
}
```

pkg/blobstore/grpcclients/cs_blob_access.go
```diff
@@ -0,0 +1,223 @@
...
+		return nil, status.ErrorProto(r.Status)
+	}
+
+	switch r.Compressor {
+	case remoteexecution.Compressor_IDENTITY:
+		return buffer.NewChunk(ba.zstdPool, r.Data), nil
```
EdSchouten, line 98: `buffer.NewChunk()` can't fail right now. Wouldn't it be better to do the checksum validation at this exact spot?

pkg/blobstore/grpcservers/content_addressable_storage_server.go
```diff
@@ -97,16 +158,24 @@ func (s *contentAddressableStorageServer) BatchReadBlobs(ctx context.Context, in
...
-		})
+	compressor := remoteexecution.Compressor_IDENTITY
+	if slices.Contains(in.AcceptableCompressors, remoteexecution.Compressor_ZSTD) {
+		compressor = remoteexecution.Compressor_ZSTD
+	}
+	params, err := s.cdcParametersFetcher.FetchCDCParameters(ctx, instanceName)
+	if err != nil {
+		return nil, err
+	}
+	for i := 0; i < len(digests); i++ {
```
EdSchouten, line 169:
```go
for i, digest := range digests {
}
```

pkg/blobstore/grpcservers/content_addressable_storage_server.go
```diff
@@ -146,14 +215,250 @@ func (s *contentAddressableStorageServer) BatchUpdateBlobs(ctx context.Context,
...
+		return cas.PutBytes(ctx, s.zstdPool, s.chunkStorage, s.chunkListStorage, params, d, data)
+	case remoteexecution.Compressor_ZSTD:
+		decoder, err := s.zstdPool.NewDecoder(ctx, bytes.NewReader(data))
+		if err != nil {
+			return util.StatusWrap(err, "Failed to acquire ZSTD decoder")
+		}
+
+		// Limit the amount of data that is read to one byte beyond the
+		// advertised size, so that corrupt or malicious streams cannot
+		// trigger unbounded decompression.
```
EdSchouten, line 230: This is not sufficient to ensure a client can't send a malicious request that causes the server to run out of memory. You could create a legitimate request that contains a 1 TB blob that compresses down to just a couple of bytes. Then this code will call `io.ReadAll(1TB + 1)`.

pkg/blobstore/mirrored/mirrored_blob_access.go
```diff
@@ -49,89 +47,64 @@ type mirroredBlobAccess struct {
...
+	backend, replicator, name, otherName := ba.backendAndReplicator()
+	var zero T
+	ret, err := backend.Get(ctx, digest)
+	if err == nil {
+		return ret, nil
+	}
+	if status.Code(err) != codes.NotFound {
+		return zero, util.StatusWrap(err, name)
+	}
+	err = replicator.ReplicateMultiple(ctx, digest.ToSingletonSet())
```
EdSchouten, line 81: Doesn't this mean that we actually load the same object twice from the second backend? The idea behind ReplicateSingle() was that this could be prevented.

pkg/blobstore/blob_access.go
```diff
@@ -3,21 +3,31 @@ package blobstore
...
-	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
-	"github.com/buildbarn/bb-storage/pkg/blobstore/slicing"
+	remoteexecution "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
 	"github.com/buildbarn/bb-storage/pkg/capabilities"
 	"github.com/buildbarn/bb-storage/pkg/digest"
 )
 
+// capablitiesProvider is a copy of the capabilities.Provider interface
+// which must be written inline for go mockgen to function in source
+// mode.
```
EdSchouten, line 13: This is awful. :-(

pkg/blobstore/empty_blob_injecting_blob_access.go
```diff
@@ -32,32 +33,42 @@ type emptyBlobInjectingBlobAccess struct {
 // blob is always present.
 //
 // More details: https://github.com/bazelbuild/bazel/issues/11063
-func NewEmptyBlobInjectingBlobAccess(base BlobAccess) BlobAccess {
+func NewEmptyBlobInjectingBlobAccess(base BlobAccess[*buffer.Chunk]) BlobAccess[*buffer.Chunk] {
```
EdSchouten, line 36: I was thinking: do we even need a BlobAccess like this now?

With chunking in place, we already have logic at the high level to distinguish between small and large objects. Maybe it's better if the empty digest is also handled in those cases?

In other words: we forbid the creation of empty Chunk objects.

pkg/blobstore/validation_caching_read_buffer_factory.go
EdSchouten, line 1: Good riddance! :)

pkg/blobstore/buffer/chunk.go
meroton-benjamin: In the get path of a blob access we don't decompress to validate the chunks. So when fetching compressed data there is decompression on the storage side to validate the integrity of the blob but otherwise all in betweens just pass the compressed data back without inspecting it.

We could reduce that to zero decompressions on the get path by letting the ChunkCoder not perform integrity checks on decoding and joining the ChunkCoder with the XXH64SufficCoder for integrity checks (effectively taking the XXH64 of the compressed bytes).

pkg/blobstore/buffer/chunk.go
meroton-benjamin: Slice assignment is not atomic so the mutex is needed for correctness. The risk is otherwise that a thread could read new slice header + old length though I'm uncertain if that could happen in practice in the codebase. In either case with the switch to atomic.Pointer[[]byte] makes it moot.

pkg/blobstore/buffer/chunk.go
meroton-benjamin: I did atomic.Pointer as the function theoretically could fail and sync.Once can not take a context or return an error so a cancelation on the GetEncoder would consume the Once and have the value be error for ever.

pkg/blobstore/buffer/read_at_closer.go
meroton-benjamin: ack

pkg/blobstore/cdc/parameters.go
meroton-benjamin: ack

pkg/blobstore/chunklist/chunk_list.go
meroton-benjamin: ack

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
meroton-benjamin: The trivial case is already handled by content_addressable_storage_server so we should never run into this in the blob access. Removed.

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
meroton-benjamin: ack

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
meroton-benjamin: ack

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
meroton-benjamin: Like the other trivial case this logic was moved into content_addressable_storage_server.go which means that no chunk list below two chunks will enter the blobaccess layer.

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
meroton-benjamin: Yes it would but when we get the response we need to keep track of which blob had a missing chunk (as it's not the missing chunk itself that should be reported missing, it's the parent blob). I'm unsure of how big of a problem it is in practice as we have to do a get on the underlying chunk list anyway and are thus already doing O(n) calls. Maybe a TODO would be sufficient here?

pkg/blobstore/chunklistvalidating/chunk_list_validating_blob_access.go
meroton-benjamin: We have to do canonicalization here because the client is allowed to send a chunk list using any arbitrary chunking methodology. After passing through the validation layers put path we can then know that the get path will yield RepMaxCDC chunked chunk lists.

EdSchouten: My concern is that this leads to a duplication of data in the worst case.

Given that we only intend on supporting RepMaxCDC and not FastCDC, what are your thoughts on:

- Requiring that `RegisterChunkMappingRequest.chunking_function == REP_MAX_CDC`,
- Simply failing with `INVALID_ARGUMENT` if the canonicalized chunk list does not match the originally provided instance?

Even though it's not 100% according to the spec, it has the advantage that we will immediately figure out which clients incorrectly implement RepMaxCDC.

pkg/blobstore/coder/xxh64_suffix_coder.go
EdSchouten: Clarification: we should likely include `parentDigest.GetKey(digest.KeyWithoutInstance)`. So not the instance name.

pkg/blobstore/coder/chunk_list_coder.go
meroton-benjamin: The purpose of the version field was in case we wanted to add things to the binary format but still be able to interpret the older format (and the error was for older versions of bb storage to not attempt to decode unknown versions). But I have removed version, digest_function and count from the binary format

pkg/blobstore/coder/xxh64_suffix_coder.go
meroton-benjamin: It might not be relevant as I have no data on if this in aggregate is an appreciable amount of the required performance, but essentially I wanted a short signature (64 bits) where accidental collision is astronomically unlikely and xxh64 is capable of reaching close to memory speeds (10+GB/s) without requiring any specific hardware support.

We could use the first 64 bits of sha256 but it may result in significant amount of overhead as software based sha-256 caps out at around 0.5GB/s per core.

pkg/blobstore/coder/zstd_coder.go
meroton-benjamin: I mostly added it for symmetry with the chunk store but it might be true that we never need to do this.

pkg/blobstore/coder/chunk_list_coder.go
meroton-benjamin: Zstd encoder removed

pkg/blobstore/coder/xxh64_suffix_coder.go
EdSchouten: For the KeyLocationMap we use very simple hash functions, because those are called in a critical path while holding locks. This doesn't apply in this specific case, as we don't hold any locks while decoding data.

I don't think we should care about performance being an issue here. Right after this we're going to Protobuf unmarshal the message, which is going to allocate a bunch of objects on the heap.

pkg/blobstore/coder/xxh64_suffix_coder.go
meroton-benjamin: done

pkg/blobstore/configuration/cs_blob_access_creator.go
meroton-benjamin: ack

pkg/blobstore/grpcclients/cs_blob_access.go
meroton-benjamin: ack

pkg/blobstore/grpcclients/cls_blob_access.go
meroton-benjamin: fixed

pkg/blobstore/grpcservers/content_addressable_storage_server.go
meroton-benjamin: ack

pkg/blobstore/grpcservers/content_addressable_storage_server.go
meroton-benjamin: Good point, I just removed the PutBytes convenience function as PutReader already handles that case perfectly well and consumers can just wrap their bytes with bytes.NewReader themselves, they don't need a helper for it.

pkg/blobstore/blob_access.go
meroton-benjamin: It is indeed, really sad about that aspect of our gomocks but could not manage to find any other solution.

pkg/blobstore/empty_blob_injecting_blob_access.go
meroton-benjamin: I am not sure, what about things like completeness checking blob access, tree/directory objects or chunk lists? Couldn't they refer to empty blobs and thus we would have to handle it at their call sites as well to be spec compliant.

pkg/blobstore/mirrored/mirrored_blob_access.go
meroton-benjamin: It does, but isn't this already the behavior internally for e.g. concurrency_limiting and queueing replicators?

pkg/blobstore/coder/chunk_coder.go
meroton-benjamin: fixed

pkg/blobstore/buffer/chunk.go
meroton-benjamin: moved to pkg/blobstore/chunk

